PR #157, opened 2026-07-01
refactor(api): extract chapter endpoints from books.py into chapters.py (#94)

frankbria:
Closes #94 (P3.1).

First slice of the `books.py` decomposition (plan 010). A **behavior-preserving move**, not a redesign: the 9 chapter-management handlers were relocated verbatim from `books.py` into a new `chapters.py` router, mounted under the same `/books` prefix so **all URLs, request/response schemas, and status codes are unchanged**.

## What moved
`create_chapter`, `get_chapters_metadata`, `get_tab_state`, `save_tab_state`, `get_chapter`, `update_chapter`, `delete_chapter`, `list_chapters`, `update_chapter_status_bulk` — a **contiguous, self-contained block** (books.py had no module-level helpers, so no shared code to duplicate).

Content/analytics, chapter-questions, and drafts/style **stay in books.py** — each is a future slice (see the map's suggested order).

## Changes
- **`books.py`**: 3,496 → 2,938 lines (−558); 43 → 34 handlers. Trimmed the 10 imports the move orphaned (7 chapter schemas + 3 chapter transactions). Pre-existing unused imports left untouched (out of scope).
- **`chapters.py`** (new): the extracted router, 9 handlers, 87% covered by the existing chapter suites.
- **`router.py`**: `include_router(chapters.router, prefix="/books", tags=["chapters"])` after the books include; handler order preserved so literal sub-paths (`/metadata`, `/tab-state`, `/bulk-status`) still resolve before `/{chapter_id}`.
- **`plans/010-books-decomposition-map.md`** (deliverable): full handler-by-cluster catalog with line ranges + suggested extraction order for the remaining slices.
- **`test_chapters_characterization.py`** (new): the invariant unique to a *move* refactor — **route-table identity**. Asserts every chapter path is still served, now owned by `chapters.router` (no duplicate mount on books.router), with literal sub-paths ordered before `/{chapter_id}`.

## Verification
- **Route table (method, path) byte-identical** before/after (69 routes; `diff` clean).
- Existing behavioral suites (`test_books_chapters_crud_coverage`, `test_books_chapter_content_coverage`) pass **before and after** the move — the characterization safety net.
- Full backend suite: **961 passed / 15 skipped, 92.38% coverage**. Ruff: net-new issues cleared. Pre-commit gates green.

## Known limitations
- Drift note: `books.py` had grown to 3,496 lines (plan written at 3,184); ranges were re-derived per the plan's drift-check instruction.
[...]

Issue #94, opened 2026-06-21
[P3.1] Decompose the 3,184-line books.py (map + first extraction)


frankbria:
# Plan 010: A safe decomposition plan for the 3,184-line books.py (design + first extraction)

> **Executor instructions**: This is a **design + characterization-test +
> single-extraction** plan, NOT a big-bang refactor. Do the characterization
> tests and the FIRST extraction only; produce the decomposition map as a
> deliverable. Follow steps in order, run every verification command, and STOP
> on any STOP condition rather than improvising a larger refactor. Update
> `plans/README.md` when done unless a reviewer maintains the index.
>
> **Drift check (run first)**: `git diff --stat e6980f3..HEAD -- backend/app/api/endpoints/books.py backend/app/api/endpoints/router.py`
> If `books.py` changed, re-derive line ranges before proceeding.

## Status

- **Priority**: P3
- **Effort**: L (scoped down to: map + characterization tests + one extraction)
- **Risk**: MED (refactor risk is why this is staged and test-gated)
- **Depends on**: 001 (test gate). Strongly benefits from 009-style coverage first.
- **Category**: tech-debt / architecture
- **Planned at**: commit `e6980f3`, 2026-06-21
- **Issue**: <filled by --issues>

## Why this matters

`backend/app/api/endpoints/books.py` is **3,184 lines** with ~50 route handlers
spanning unrelated concerns: book CRUD, chapters, TOC, questions, draft
generation, summaries, exports, analytics, and chapter content. It is ~45× the
backend's median module size and only ~46% covered. Every book-area feature
edits this one file, maximizing merge conflicts, review fatigue, and accidental
coupling. A full split is high-risk for a single pass, so this plan does it
**safely and incrementally**: (1) produce a decomposition map, (2) add
characterization tests that pin current behavior of the first slice to extract,
(3) extract exactly ONE cohesive router (chapters) into its own module behind
the same URL paths, proving the pattern. Subsequent slices become their own
follow-up plans using the map.

## Current state

- `backend/app/api/endpoints/books.py` — 3,184 lines, single `router = APIRouter()`,
  mounted in `backend/app/api/endpoints/router.py:10`:
  ```python
  router.include_router(books.router, prefix="/books", tags=["books"])
  ```
- Other endpoint modules already follow the one-domain-per-file pattern
  (`users.py`, `export.py`, `sessions.py`, `transcription.py`,
[...]

PR #165, opened 2026-07-01
fix(chapters): version-guard bulk-status + structured read-handler errors (#159)

### LINKED ISSUE #159: [P3.8] chapters.py: bulk-status TOC write lacks concurrency guard; several read handlers lack structured error handling
Two latent issues in `backend/app/api/endpoints/chapters.py`, both **pre-existing** (verbatim-moved from `books.py` in #94, confirmed byte-identical) — surfaced by CodeRabbit on #157 and deferred from that behavior-preserving extraction.

## 1. `update_chapter_status_bulk` — lost-update / TOCTOU on the TOC (Major)
The handler does `get_book_by_id` → mutate the in-memory `chapters` list → `update_book(...)` overwriting the **entire** `table_of_contents`, with **no version check** between read and write. Unlike `create/update/delete_chapter` (which use `toc_transactions` and surface `409 Concurrent modification`), a concurrent TOC edit in the window is silently clobbered.

Fix: route the persist through the optimistic-concurrency `toc_transactions` path (or add a version guard) and return 409 on conflict. NB: this same non-transactional full-TOC-overwrite pattern (`update_book(book_id, update_data, ...)`) is used across many books.py handlers (summary/toc/etc.), so consider a shared fix.

## 2. Missing structured error handling on read/aggregate handlers (Major)
`get_chapters_metadata`, `get_tab_state`, `list_chapters`, `save_tab_state` call DB/service layers with no `try/except`, so an unexpected error propagates as an unlogged 500 — inconsistent with `create/update/delete_chapter` which catch, log, and return a structured 500. Wrap the service calls in the same pattern (preserving the existing 404/403 checks).

Also worth doing while here: the `B008` (Depends in defaults) and `B904` (`raise ... from e`) ruff nits CodeRabbit flagged in this file.

Related: #158 (dead-branch status code in `create_chapter`).

frankbria:
Closes #159.

Two latent defects in `backend/app/api/endpoints/chapters.py`, both pre-existing (moved verbatim from `books.py` in #94, surfaced by CodeRabbit on #157).

## 1. bulk-status lost-update → 409
`update_chapter_status_bulk` read the book, mutated the in-memory chapters, then overwrote the **entire** `table_of_contents` with `update_book(...)` — **no version check**, so a concurrent TOC edit in the window was silently clobbered (unlike create/update/delete_chapter, which surface `409`).

Now it persists through a new `update_chapter_statuses_with_version_guard()` in `toc_transactions.py` that does a single **compare-and-swap** `update_one` filtered on `table_of_contents.version`. Because a bulk status change touches one document, a single conditional update is atomic on its own — no multi-document transaction — so the guard holds on **both replica-set and standalone** deployments (a transaction would instead surface a concurrent commit as a `WriteConflict`). A stale version → `modified_count == 0` → `Version conflict` → **409**, mirroring the sibling handlers. The helper re-emits the `book_update` audit-log entry the old `update_book()` path produced.

The handler keeps its upfront ownership pre-check (so wrong-owner stays **403** and missing book stays **404**) and its transition validation (invalid transition **400** before any write).

## 2. Structured error handling on read handlers
`get_chapters_metadata`, `get_tab_state`, `list_chapters`, `save_tab_state` had no `try/except`, so an unexpected error propagated as an **unlogged 500**. Each now wraps its body in `try / except HTTPException: raise / except Exception: logger.error(..., exc_info=True) → structured 500`, preserving the existing 404/403 checks and response shapes.

## Tests
- `test_db/test_toc_transactions.py`: helper unit tests — happy path (version increments, statuses set, **audit log emitted**), **stale version → Version conflict (no write)**, nested subchapter, no-matching, timestamp, book-not-found, wrong-owner.
- `test_api/test_routes/test_chapters_error_handling.py` (new): endpoint version-conflict → **409**, bulk-status unexpected error → structured **500**, each read handler → structured **500**, and read handlers still **404** on missing book.
- Full backend suite: **984 passed / 15 skipped, 92.16% coverage** (gate ≥85%). `codex` cross-family review clean after fixes (it drove the CAS-not-transaction and audit-log-preservation changes).

## Scope decisions
- **Applied in `chapters.py`, not `books.py`.** The CodeRabbit plan assumed `chapters.py` didn't exist; #94/#157 already extracted these handlers there.
- **Skipped the plan's ruff B008/B904 config (Phase 3).** Ruff B rules aren't enabled in this repo and the lint hook is non-blocking (`|| true`) — a repo-wide `[tool.ruff.lint]` config is new lint scope for an optional nit, out of scope for a bug fix.

## Known limitations
- The same non-transactional full-TOC-overwrite pattern exists in other `books.py` handlers (`update_chapter_content`, summary/toc, etc.). The issue explicitly deferred that as a possible **shared fix**; this PR fixes only the two named handlers. Follow-up candidate.
- The bulk-status version-conflict path is covered at the unit level (stale `expected_version`) and via the endpoint's 409 mapping; a true HTTP-level concurrent-writer race isn't simulated (no arbitrary-timeout/flaky test).

## Summary by CodeRabbit

* **New Features**
  * Added optimistic concurrency to bulk chapter status updates, including nested subchapter support and optional `last_modified` timestamping.
* **Bug Fixes**
  * Implemented structured error handling for chapter metadata, chapter listing, tab-state retrieval, and tab-state saving (standardized `500` messaging while preserving `403/404`).
  * Prevents lost updates via TOC version compare-and-swap; returns `409` on conflicts.
  * Ensures access-log failures don’t break already-committed bulk updates.
* **Tests**
  * Added API and DB tests covering conflict/no-write behavior, authorization/data errors, and structured `500/409` responses.


frankbria:
Addressed CodeRabbit's review in c34b09b:

- **Major (version-less TOC → false 409)**: fixed — the CAS filter now matches a legacy TOC with no `version` field on its absence (`$exists: false`) rather than the defaulted `1`, so a first versioned write can't produce a false conflict. New test `test_bulk_status_helper_versionless_toc_no_false_conflict`.
- **Major (post-commit `log_access` failure fails a committed update)**: fixed — the bulk-status access-logging loop is now wrapped in try/except (logs and continues), matching the existing `update_chapter_content` pattern. New test `test_bulk_status_logging_failure_does_not_fail_committed_update` asserts 200 + persisted status.
- **Minor (raw-string regex)**: applied.

Full backend suite: 986 passed / 15 skipped, 92.17% coverage.

PR #227, opened 2026-07-04
fix(books): version-safe TOC/summary writes to stop lost updates (#177)

### LINKED ISSUE #177: [P0.5] Lost update: chapter-content auto-save overwrites entire TOC with no version guard
**Problem:** `update_chapter_content` (the 3s autosave target) reads the whole book, mutates one chapter in memory, then persists the ENTIRE `table_of_contents` via `update_book`'s version-less `$set`. Two chapters auto-saving concurrently (multiple open tabs), or a concurrent TOC reorder, silently clobber each other — real data loss. `generate_table_of_contents` additionally hardcodes `version:1`, resetting the CAS counter.

**Evidence:** `backend/app/api/endpoints/books.py:1598-1612`; `backend/app/db/book.py:80-84` (`find_one_and_update` `$set`, no version filter). Contrast `update_chapter_statuses_with_version_guard` in `toc_transactions.py`. (Known deferred #159 follow-up.)

**Acceptance criteria:**
- `update_chapter_content` routes through a compare-and-swap on `table_of_contents.version` (409 on mismatch) or a targeted positional/arrayFilters `$set` for the single chapter.
- Same treatment for `update_book_summary` (:657) and `generate_table_of_contents` (increment from current version, don't hardcode 1).
- Test simulates interleaved writers and asserts no lost update.

**Depends on:** none.

frankbria:
Closes #177 (P0.5, critical / data-integrity).

## Problem
`update_chapter_content` (the 3s autosave target), `update_book_summary` / `patch_book_summary`, and `generate_table_of_contents` all persisted the **whole** `table_of_contents` (or the whole `summary_history` array) through `update_book`'s version-less `$set`. Two chapters auto-saving concurrently (multiple open tabs), or a concurrent TOC reorder, **silently clobbered each other** — real data loss. `generate_toc` also hardcoded `version: 1`, resetting the compare-and-swap counter other writers rely on.

## Fix
Two new DB helpers in `app/db/book.py`, mirroring the CAS pattern from `update_chapter_statuses_with_version_guard` (#159):

- **`apply_chapter_content_update`** — targeted **arrayFilters positional `$set`** of only the one chapter's fields + an atomic **`$inc` of `table_of_contents.version`**. Concurrent saves to *different* chapters no longer collide, so there's **zero false-conflict / no 409 / no frontend change** for the high-frequency autosave path (chosen over a whole-TOC version guard for exactly that reason — AC's "targeted positional/arrayFilters `$set`" branch). The update query also requires the target chapter (and parent, for a subchapter) to still exist, so a chapter deleted mid-edit is a clean **404** instead of a false success that only bumps the version.
- **`update_book_summary_atomic`** — an **aggregation-pipeline update** that archives the document's **current** summary at write time (not a value the caller read earlier), so two concurrent edits (A→B, A→C) **both** land in history (single-doc writes serialize; the second sees the first's committed summary). History capped at the last 20. User summaries starting with `$` are wrapped in `$literal`. Both `PUT` and `PATCH /summary` route through it.
- **`generate_table_of_contents`** — increments the version from the current TOC (`0 → 1` on first generation) instead of hardcoding `1`.

## Tests (real MongoDB)
`tests/test_db/test_lost_update_guard.py` — interleaved-writer tests proving **no lost update** for chapter content and summary history; targeted single-chapter/subchapter update; stale-chapter no-match (TOCTOU 404); legacy TOC without a version; `$`-leading summary stored literally; history cap. Plus a generate-toc version-increment endpoint test.

Backend: **1073 passed / 14 skipped, 92% coverage**. Pre-commit gates green (no `--no-verify`).

## Reviews
`codex` (cross-family, 3 rounds) drove: the stale-`previous_summary` clobber → pipeline archive; the false-success/`matched_count` gap → chapter-existence query predicate + endpoint 404; `PATCH /summary` still doing read-modify-write → shared helper; and a `$`-summary-as-field-path bug → `$literal`. All fixed with tests.

## Known limitation
`generate_table_of_contents` remains a full-TOC replace (its intended semantics for an explicit, 2-per-5-min rate-limited regenerate); only the hardcoded version is fixed, per the AC. A concurrent content-save during a regenerate can still be overwritten — but a regenerate is destructive by design. A version guard there would 409 the normal single-shot generate flow, so it's deliberately not added.

## Summary by CodeRabbit

* **Bug Fixes**
  * Chapter edits now update only the changed chapter instead of overwriting the entire table of contents, reducing the risk of lost updates during concurrent saves.
  * Book summaries now keep revision history more reliably, with older versions archived automatically and limited to the most recent 20 entries.
  * Regenerating a table of contents now advances its version correctly rather than resetting it.
  * Added regression coverage for concurrent updates and version handling.


frankbria:
Addressed both CodeRabbit findings in c212b1b:
- **Major (books.py:653)** — `update_book_summary`/`patch_book_summary` now return **404** when `update_book_summary_atomic` returns `None` (book deleted between the ownership check and the write), instead of a false 200. Added a helper test covering the None (missing/wrong-owner) path.
- **Minor (test docstring)** — corrected to describe the aggregation-pipeline update (`$concatArrays`/`$slice`), not `$push`/`$slice`.

Backend suite green (pre-commit, no bypass).

PR #228, opened 2026-07-05
fix(users): unique index + idempotent create to stop duplicate user records (#178)

### LINKED ISSUE #178: [P0.6] Duplicate user records: no unique index on users.auth_id + check-then-insert race
**Problem:** First authenticated request auto-creates the backend user via a plain `insert_one`, and no unique index exists on the users collection at startup (only question indexes are built). An SPA's parallel first-load requests each find no user and insert → multiple user docs with the same `auth_id`, split `book_ids`, and nondeterministic reads.

**Evidence:** `backend/app/core/security.py:146` (get-then-insert), `backend/app/db/user.py:33-37` (`insert_one`), `backend/app/main.py:64-67` (only `ensure_question_indexes()` at startup; `indexing_strategy.py` unique-index defs never run).

**Acceptance criteria:**
- Unique index on `users.auth_id` (and `email`) created at startup.
- Auto-create is idempotent: `find_one_and_update(..., upsert=True)` or catch `DuplicateKeyError` and re-fetch.
- Test asserts concurrent first-loads produce exactly one user record.

**Depends on:** none.

frankbria:
Closes #178 (P0.6, critical, data-integrity).

## Problem
The auth auto-create path (`security.py`) did a **check-then-insert** with **no unique index** on the users collection (startup only built question indexes). An SPA's parallel first-load requests each found no user and inserted → **multiple user docs sharing one `auth_id`**, split `book_ids`, and nondeterministic reads.

## Fix
- **`ensure_user_indexes()`** (`app/db/user.py`) — unique index on `users.auth_id` (plain) and `email` (sparse), created at startup in `main.py` alongside `ensure_question_indexes()`. Per-index `try/except` so one failing build can't brick startup.
- **`create_user()`** idempotent — catch `DuplicateKeyError`, re-fetch the `auth_id` winner (the race the issue describes). Re-raises when the collision is on a *different* key (an email already taken by another `auth_id`) so it surfaces rather than silently returning `None`.
- **`POST /users/`** maps `DuplicateKeyError` → **409** (its intended conflict response) for the rare insert that slips past the endpoint's pre-checks under a race.

Approach matches the issue's AC ("catch `DuplicateKeyError` and re-fetch") — smallest diff, `create_user`'s signature unchanged.

## Tests (real Mongo)
- `test_user_unique_index.py`: both unique indexes built; **8 concurrent `create_user` for one `auth_id` → exactly one record**; email-collision re-raise.
- `test_users_coverage.py`: endpoint `DuplicateKeyError` → 409.
- Full backend suite: **1079 passed / 14 skipped, 91.87% coverage** (gate green, no `--no-verify`).

## Demo (outcome evidence)
Ran end-to-end against real Mongo: startup builds both unique indexes; 10 concurrent auto-creates for one user → **1 record, all 10 callers see the same `_id`, no crash**; email collision on a different `auth_id` → `DuplicateKeyError` (→ 409).

## Reviews
- **codex** (cross-family, pre-PR): flagged the blanket `DuplicateKeyError` catch degrading a would-be-409 into a `None`→500 on the legacy endpoint. Fixed with the re-raise + 409 mapping above.

## Known limitations
- If a production collection **already holds** duplicate `auth_id`/`email` docs, the unique build fails at startup and is **logged** (not fatal) — an operator must dedupe once before the index (and thus the race guard) takes effect. Documented with a `ponytail:` comment at the build site.
- The issue cited `indexing_strategy.py` as holding unused users unique-index defs; that file is actually about chapter tabs — no such defs existed, so the index is created fresh here.

## Summary by CodeRabbit

* **Bug Fixes**
  * User creation now returns a clear conflict error when the same account ID or email is already in use.
  * Improved handling of simultaneous sign-in or account-setup attempts so duplicate records are avoided and the correct existing user is returned when possible.
  * Added stronger protection against duplicate user accounts during registration and session-based account creation.

### DIFF
--- a/backend/app/api/endpoints/users.py
+++ b/backend/app/api/endpoints/users.py
@@ -2,6 +2,7 @@ import logging
 
 from fastapi import APIRouter, Depends, HTTPException, status, Request, UploadFile, File
 from fastapi.security import HTTPBearer
+from pymongo.errors import DuplicateKeyError
 from typing import List, Dict
 from datetime import datetime, timezone
 
@@ -292,6 +293,13 @@ async def create_new_user(user: UserCreate):
     try:
         created_user = await create_user(user_data)
         return created_user
+    except DuplicateKeyError:
+        # Concurrent insert slipped past the pre-checks and hit the unique index
+        # (issue #178) — surface the intended conflict, not a 500.
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail="User with this auth ID or email already exists",
+        )
     except Exception:
         logger.error("Error creating user", exc_info=True)
         raise HTTPException(
--- a/backend/app/core/security.py
+++ b/backend/app/core/security.py
@@ -1,4 +1,5 @@
 from passlib.context import CryptContext
+from pymongo.errors import DuplicateKeyError
 from typing import Optional, Dict, Any, List
 from fastapi import HTTPException, status, Request
 from app.core.better_auth_session import (
@@ -197,6 +198,20 @@ async def get_current_user_from_session(request: Request) -> Dict:
 
         except HTTPException:
             raise
+        except DuplicateKeyError:
+            # A concurrent first-load request won the insert (issue #178). Re-fetch
+            # so both requests resolve to the one shared record instead of erroring.
+            logger.info(
+                f"Concurrent auto-create race for {user_id}; using the existing record"
+            )
+            from app.db.user import get_user_by_auth_id
+
+            user = await get_user_by_auth_id(user_id)
+            if user is None:
+                raise HTTPException(
+                    status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                    detail="Failed to create user account. Please try signing in again or contact support."
+                )
         except Exception as e:
             logger.error(f"Failed to auto-create user {user_id}: {str(e)}")
             raise HTTPException(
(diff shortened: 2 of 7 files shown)


frankbria:
### Cross-family review (codex) — pre-PR + post-commit

Ran `codex` (cross-family) twice. Two Medium findings, both on the legacy `POST /users/` interaction with the new idempotency; both fixed.

1. **Blanket `DuplicateKeyError` swallow → None→500 on legacy endpoint** (round 1). Fixed by re-raising when there's no `auth_id` winner + mapping `DuplicateKeyError`→409 at the endpoint.
2. **Idempotent `create_user` bypassed the legacy endpoint's intended 409** on a duplicate-`auth_id` race (returned 201-with-winner, could surface the winner's fields) (round 2). Fixed by making `create_user` a **pure-insert primitive that re-raises**, with each caller owning its policy:
   - auth auto-create path → catches `DuplicateKeyError`, re-fetches the winner (idempotent success for the concurrent-first-load race the issue targets);
   - legacy `POST /users/` → `DuplicateKeyError`→**409** (a real conflict).

Also dropped `background=True` from the unique-index builds (a correctness-critical unique index should build foreground).

**Verification:** full backend suite **1081 passed / 14 skipped, 91.87% cov** (gate enforced, no `--no-verify`). End-to-end demo against real Mongo: startup builds both unique indexes; 10 concurrent inserts for one `auth_id` → **exactly one record**, the 9 duplicates rejected by the index; the auto-create path re-fetches the winner (unit-tested in `test_security.py`).

Issue #175, opened 2026-07-04
[P0.3] Offload OpenAI calls off the FastAPI event loop


frankbria:
**Problem (hand-verified):** `_make_openai_request` defines `_sync_request()` on the **synchronous** `OpenAI` client and its `_async_wrapper` calls it inline with no `asyncio.to_thread`/`AsyncOpenAI`. The blocking multi-second-to-minute HTTP call runs on the event loop. Deployment runs a single Uvicorn worker (`ecosystem.config.template.js`), so one user's generation freezes the whole process — health checks, autosaves, and every other user's request stall until it returns.

**Evidence:** `backend/app/services/ai_service.py:50` (`self.client = OpenAI(...)`), `:196-207` (`_async_wrapper` → `_sync_request()` inline). Contrast `backend/app/services/export_service.py:213` which correctly uses `asyncio.to_thread`.

**Acceptance criteria:**
- AI calls run off-loop (`AsyncOpenAI` or `await asyncio.to_thread(_sync_request)`).
- Run ≥2 Uvicorn workers under PM2 for headroom.
- A test/benchmark shows a concurrent request is served while an AI call is in flight.

**Depends on:** none. Highest-leverage, smallest fix — do first.

Issue #173, opened 2026-07-04
[P0.1] Cap per-user AI usage / OpenAI spend (cost-control)


frankbria:
**Problem:** There is no per-user metering or quota on AI generation. Rate limits are per-time-window, not a spend budget, and are bypassable behind nginx (see P0.8). A single beta user — or a leaked session cookie — can trigger unlimited GPT-4 generations, an uncapped externally-triggerable cost on a paid product.

**Evidence:** `backend/app/services/ai_service.py` (all generation), AI handlers in `backend/app/api/endpoints/books.py`, `backend/app/api/dependencies.py` (window-based limiter only). No `quota`/`usage_limit`/`plan_tier` code anywhere in the repo.

**Acceptance criteria:**
- A per-user generation counter (daily and/or monthly) persisted in Mongo, incremented on every AI endpoint (TOC, clarifying questions, chapter questions, draft, style/content/transcription enhancement).
- At the cap the endpoint returns 429 with a clear, user-facing message; cap is configurable per environment.
- Counter keys off the user's plan/entitlement when P0.2 lands.
- Test proves the Nth+1 call in a window is rejected.

**Depends on:** none (can ship a flat cap now; refine per-plan after P0.2). Real enforcement needs **P0.8**.

PR #230, opened 2026-07-06
fix(security): per-user shared-store rate limiting; pin uvicorn proxy headers (#180)

### LINKED ISSUE #180: [P0.8] Rate limiting is tenant-wide and bypassable behind nginx
**Problem:** The limiter keys on `request.client.host`, but Uvicorn runs without `--proxy-headers`, so behind nginx every client appears as the proxy IP. Fixed-path endpoints collapse into one shared bucket (one user can 429 the endpoint for the whole tenant base), and the counter is a plain in-process dict — per-worker, resets on restart, never evicted (unbounded growth).

**Evidence:** `backend/app/api/dependencies.py:13` (in-proc dict), `:82` (`client_ip = request.client.host`); `ecosystem.config.template.js:9` (`--host 0.0.0.0 --port 8000`, no `--proxy-headers`).

**Acceptance criteria:**
- Uvicorn runs with `--proxy-headers --forwarded-allow-ips=<nginx>` (or Starlette ProxyHeadersMiddleware) so client IPs are real.
- Limiter uses a shared store consistent across workers/restarts and keys per authenticated user where appropriate.
- Test asserts two different users don't share a bucket.

**Blocks:** real enforcement of P0.1.

frankbria:
Closes #180

## Problem

`get_rate_limiter` keyed buckets on `client_ip:endpoint` in a plain in-process dict:
- **per-worker** — PM2 runs uvicorn `--workers 2` (#175), so every limit was effectively doubled and inconsistent
- **reset on restart**, never evicted (unbounded growth)
- **IP-keyed** — behind a NAT/proxy misconfiguration, one user could 429 an endpoint for everyone sharing the IP

## Fix

- **Per-authenticated-user buckets**: the limiter now takes `Depends(get_current_user_from_session)` and keys on `auth_id`. Verified every rate-limited endpoint already requires session auth, so FastAPI's per-request dependency cache makes this free (no extra DB hit). Client IP remains only a defense-in-depth fallback (`request.client` None-guarded).
- **Shared persistent store**: counts live in Mongo via the existing atomic `$inc`+TTL `increment_usage` DAO (#173) — consistent across workers, survives restarts, TTL self-eviction, zero new DAO code. Epoch-aligned fixed windows; the 429 + `X-RateLimit-*`/`Retry-After` contract is unchanged.
- **Proxy headers pinned**: ecosystem template adds `--proxy-headers --forwarded-allow-ips=127.0.0.1`. Note the issue premise was partially stale: uvicorn 0.35 defaults `proxy_headers=True` trusting 127.0.0.1, staging nginx already sends `X-Forwarded-For` from localhost, and staging logs already show real client IPs — the flags make that explicit instead of default-dependent.
- **Deleted** the deprecated `rate_limit()` + the unbounded `rate_limit_cache` dict (zero production callers — this was the exact never-evicted dict from the issue) and a dead books.py import.

## Acceptance criteria

- ✅ Uvicorn runs with `--proxy-headers --forwarded-allow-ips=<nginx>` (ecosystem template; nginx proxies via localhost)
- ✅ Limiter uses a shared store consistent across workers/restarts, keyed per authenticated user
- ✅ `test_two_users_do_not_share_a_bucket` asserts user A exhausting the limit does not 429 user B

## Tests

Rewrote `TestRateLimiting` against real Mongo (9 tests): the AC two-users test, cross-worker shared bucket (two limiter closures, 4th request 429s), 429 header contract, BYPASS_AUTH short-circuit, per-endpoint isolation, deterministic window reset (patched clock), IP fallback isolation, Mongo-failure fail-closed pin, and an HTTP-level DI wiring test (a regression dropping the `current_user` sub-dependency fails it).

Backend: **1086 passed / 13 skipped, 91.87% coverage**. Gates green, no bypass.

## Third-party review

opencode (GLM) pre-PR, 2 rounds. Round 1: 3 Major / 6 minor / 6 test gaps. Fixed: `request.client` None-guard, DI wiring test, fail-closed pin test, clock-patched window test, burst-tradeoff docs. Rebutted with evidence (round 2 verified all sound): fail-open-on-Mongo-down (auth hits the same Mongo first — no new blast radius; fail-closed is the safe direction), fallback-order divergence (matches the `get_ai_usage_quota` sibling; `auth_id` is set at every creation point), per-path bucket proliferation (pre-existing intended per-resource semantics; ~100B docs, TTL-bounded, 1:1 with requests).

## Known limitations

- **Fixed-window boundary burst**: up to 2× limit across an epoch-aligned boundary (standard tradeoff; documented in the docstring). The old limiter allowed the same burst, just at an unpredictable boundary.
- **Per-path buckets** mean parameterized routes (e.g. per-book TOC regen) give per-resource, not aggregate, protection — a user with N books gets N× the per-book limit. AI endpoints are backstopped by the per-user quota (#173); non-AI routes are not. Pre-existing semantics, unchanged here.
- One Mongo `$inc` round-trip per rate-limited request (these endpoints already hit Mongo for auth + audit; negligible).


frankbria:
## Third-party review (opencode / GLM) — post-PR pass

**Approved — ship it.**

Reviewed the full PR for #180 (per-process IP-keyed rate limiter → per-authenticated-user, Mongo-backed, epoch-aligned fixed-window via the existing atomic `$inc`+TTL counter). Verified across two adversarial passes: (1) bucket isolation is sound — distinct `auth_id`s never share, and multi-worker sharing is proven by test; (2) the subject fallback chain (`auth_id → id → clerk_id → client-ip`) is unreachable past `auth_id` since the user doc is fetched by that field under a unique index, so the order inconsistency with the audit path is moot defense-in-depth; (3) fail-closed on Mongo failure adds no new blast radius because `get_current_user_from_session` already hits Mongo per request before the limiter body runs; (4) `Retry-After`/`X-RateLimit-Reset` math is correct and race-free (single `now` capture, conservative rounding). Accepted and documented tradeoffs: the 2× boundary burst inherent to fixed windows (now globally predictable, noted in the docstring), and per-path bucketing on parameterized routes (1:1 doc-per-request, TTL-bounded, backstopped by the AI quota on generation endpoints). The `--proxy-headers --forwarded-allow-ips=127.0.0.1` pin in the ecosystem template closes the X-Forwarded-For spoofing vector. Test coverage spans per-user isolation, cross-worker counting, header correctness, DI wiring via real HTTP, fail-closed propagation, deterministic window reset, and the IP fallback. All 26 tests pass, ruff clean.

**Round history**: R1 — 3 Major / 6 minor / 6 test gaps → 5 fixed (`request.client` None-guard, HTTP-level DI wiring test, fail-closed pin test, clock-patched window test, burst docs), rest rebutted with evidence. R2 — all fixes verified correct, all rebuttals confirmed sound; 2 doc nits applied. R3 (final diff) — caught one dead `import time` (fixed in a09b52c), then CLEAN.

frankbria:
## Demo — real app, real Mongo, real better-auth sessions (no bypass)

Setup: `uvicorn app.main:app --workers 2 --proxy-headers --forwarded-allow-ips=127.0.0.1` (exact template args) on :8010, `BYPASS_AUTH=false`, dedicated Mongo DB seeded with two real better-auth users+sessions (alice, bob). All requests via curl with real session cookies through the full auth path.

### AC3 — two users don't share a bucket
`PATCH /users/me` (limit 5/60), minute-aligned window:
```
Alice requests 1-5 -> 200 200 200 200 200
Alice request 6:
  HTTP/1.1 429 Too Many Requests
  x-ratelimit-limit: 5
  x-ratelimit-remaining: 0
  x-ratelimit-reset: 1783310820
  retry-after: 59
Bob, same endpoint, same window -> 200   ✅
```

### AC2 — shared store across workers AND restarts
- The 5 counted requests round-robined across **2 uvicorn workers** and the 6th 429'd exactly at the limit — the old per-worker dict would have allowed ~2× before any 429.
- Then a **full process kill + restart** (fresh workers, empty memory), same window:
```
HTTP/1.1 429 Too Many Requests
retry-after: 32          ✅ bucket persisted in Mongo; old dict would have reset to 0
```

### AC1 — proxy headers honored (and trusted only from localhost)
Running with the template's `--proxy-headers --forwarded-allow-ips=127.0.0.1`:
```
curl -H "X-Forwarded-For: 203.0.113.99" /api/v1/health   (from localhost = trusted proxy)
  log: Request started: GET /api/v1/health from 203.0.113.99
  log: INFO: 203.0.113.99:0 - "GET /api/v1/health HTTP/1.1" 200 OK
curl /api/v1/health   (no header)
  log: Request started: GET /api/v1/health from 127.0.0.1   ✅
```
(Independently verified staging: nginx sends XFF from localhost and backend logs already show real client IPs — the template flags pin this rather than relying on uvicorn defaults.)

Demo hiccup worth recording: first run seeded users as `@demo.test` — pydantic `EmailStr` rejects the reserved `.test` TLD at **response** validation, so PATCH 500'd while the limiter counted correctly (429 still fired on request 6). Reseeded with a routable domain; not a code issue.

PR #231, opened 2026-07-06
fix(ai): size draft max_tokens to target_length; fail loudly on truncation (#181)

### LINKED ISSUE #181: [P1.1] Chapter drafts silently truncated to ~750 words regardless of requested length
**Problem:** `generate_chapter_draft` uses the default `max_tokens=1000` (~750 words) and, unlike `enhance_text`/`transform_text_style`, has no `finish_reason=='length'` guard. The endpoint accepts `target_length` up to 10000 and the prompt asks for that many words, but any request over ~750 words returns a truncated draft with `success:true` and a `word_count` reflecting the truncation. It's inserted into the editor as-is.

**Evidence:** `backend/app/services/ai_service.py:932` (calls `_make_openai_request` with default cap), `:182` (default `max_tokens=1000`), `:1019` (the guard that other methods have); `backend/app/api/endpoints/books.py:2605` (accepts up to 10000).

**Acceptance criteria:**
- `max_tokens` sized to `target_length` (~×1.6 tokens, clamped to model limit).
- `finish_reason=='length'` guard: truncated output fails loudly (or is clearly flagged) instead of saving as success.
- Test asserts a large-target request isn't silently shortened.

frankbria:
Closes #181.

## Problem
`generate_chapter_draft` called `_make_openai_request` with the default `max_tokens=1000` (~750 words) and — unlike `transform_text_style`/`enhance_text`/`enhance_transcription` — had no `finish_reason=='length'` guard. Any request over ~750 words returned a truncated draft with `success:true` and a `word_count` reflecting the truncation, inserted into the editor as-is. The endpoint's outer `except Exception` also swallowed its own `HTTPException`, so even explicit failures degraded to a generic 500.

## Fix
- **`ai_service.py`**: `max_tokens = clamp(target_length × 1.6, 500, 4000)` (module constants `DRAFT_WORDS_TO_TOKENS_FACTOR`, `DRAFT_MAX_COMPLETION_TOKENS` — the same fixed 4000 cap the sibling methods use, leaving prompt headroom in gpt-4's 8192 window; the 500 floor keeps small targets from spuriously tripping the guard). Added the sibling `finish_reason=='length'` guard: a truncated draft returns `success:False` with a clear message instead of being saved.
- **`books.py`**: AI failure now raises **503** with the service's message (sibling preview-endpoint pattern), and `except HTTPException: raise` precedes the generic handler so it propagates.

## Tests
- `test_max_tokens_scaled_from_target_length` (2000 → 3200), `test_max_tokens_clamped_to_model_cap` (10000 → 4000), `test_max_tokens_has_headroom_floor_for_small_targets` (100 → 500), `test_truncated_output_is_rejected_not_saved` (mirrors the sibling convention).
- `test_generate_draft_ai_returns_failure_503` replaces the old test that characterized the swallowing bug (asserted the generic 500).
- Full backend suite: **1089 passed / 13 skipped, 91.88% coverage**.

## Reviews
Pre-PR third-party review: **opencode (GLM)** — 1 Major + 2 Minor. Minor (floor headroom) fixed; Major and the other Minor rebutted with evidence (see Known Limitations and the review comment below).

## Known Limitations
- The UI offers up to "5,000 words (Extended)" but gpt-4's shared 8192-token context caps completions at ~4000 tokens (~2500–3000 words). GPT-4 typically undershoots long targets and finishes with `stop` inside the budget, so large requests usually succeed with a shorter-than-asked draft (a big improvement over the prior silent ~750-word cap); when the budget is genuinely exhausted the request now fails loudly per the issue AC. Raising the ceiling is a model-selection change (e.g. gpt-4o/128k) — follow-up issue to be filed.


frankbria:
## Third-party review — opencode (GLM), post-PR round**No findings.**

Verification of the non-obvious points:

- **Token math**: `int(2000 * 1.6)` = 3200 ✓; `min(max(10000*1.6=16000, 500), 4000)` = 4000 ✓; `min(max(100*1.6=160, 500), 4000)` = 500 ✓ — all three tests' asserted values match the implementation.
- **Exception ordering**: `except HTTPException: raise` precedes `except Exception`, correctly stopping the broad handler from re-wrapping the 503 into a 500 (the exact bug the old test documented). The 503 raise sits inside the `try`, so it's reachable.
- **Truncation guard**: `getattr(choice, "finish_reason", None)` is defensive; the early `return` precedes `draft_content = choice.message.content`, so a partially-populated mock can't crash; returns `draft=""` + `success:False` so nothing partial is persisted. The error string contains "cut off", satisfying the test's OR assertion.
- **Test consistency**: `test_..._503` now asserts the real service message (`Failed to generate draft: model down`) rather than the generic re-wrapped string, confirming the HTTPException pass-through works end-to-end.

The 500→503 reclassification and the documented >2500-word ceiling are intentional per the issue/AC, not defects.

---
Pre-PR round (same reviewer) found 1 Major + 2 Minor: the Minor (misleading truncation advice at tiny targets) was fixed via the 500-token floor; the Major (4000-token cap below the UI's 5,000-word option) is rebutted as non-deterministic (GPT-4 undershoots long targets; loud failure when the budget is genuinely exhausted is the issue's AC) and tracked as follow-up #232; the other Minor (mock string) was verified in-diff.

PR #233, opened 2026-07-07
[P1.2] Surface AI question-generation failures as structured errors; tag template fallbacks (#182)

### LINKED ISSUE #182: [P1.2] AI question-generation failures are swallowed into generic fallback questions (HTTP 200)
**Problem:** `generate_chapter_questions` wraps the OpenAI call in `except Exception: return []`, and the caller fills generic template questions when the list is empty. A real OpenAI outage (bad key, removed model, network error) returns HTTP 200 with plausible-looking fake questions and no error surfaces — unlike the TOC path which raises a structured error.

**Evidence:** `backend/app/services/ai_service.py:805-808`; `backend/app/services/question_generation_service.py:590-600` (fallback), `:831-870` (templates).

**Acceptance criteria:**
- AI failure either surfaces a structured error (like the TOC path) or the response is tagged `is_fallback`/`source` so the client/tests can tell real from fallback.
- Test patches `_make_openai_request` to raise and asserts the response does not masquerade as successful AI generation.

frankbria:
Closes #182.

## Problem

`ai_service.generate_chapter_questions` swallowed every exception into `return []`, and the service layer substituted template questions on any failure — so a genuine OpenAI outage (bad key, removed model, network error) returned **HTTP 200 with plausible-looking fake questions**. The TOC path, by contrast, raises a structured error.

## Fix (per the issue plan's option 3: propagate + tag)

- **`AIServiceError` now propagates** through both layers (`ai_service.py`, `question_generation_service.py`) to the API route, where the existing `handle_question_generation_error` maps it to a structured **503** (retryable) / **500** with `QUESTION_GENERATION_FAILED` — matching the TOC pattern. Covers generate, bulk regenerate, and single-question regenerate (all three route-tested).
- **Template fallbacks remain** only for empty/thin AI responses and non-AI unexpected errors, now tagged with a new **`is_fallback: bool = False`** field on `QuestionBase` (mirrors the `regeneration_count` safe-default precedent; schema-less Mongo → no migration; legacy docs read back `False`). Flows automatically into `QuestionCreate`/`Question`/`GenerateQuestionsResponse`.
- **`regenerate_single_question`** carries `is_fallback` into its in-place replacement, so a template replacement can't inherit the old question's `False`.

## Test-suite repair this forced

Four test files (`test_api/test_question_regeneration.py`, `test_api/test_questions_integration.py`, `test_debug_questions.py`, `test_debug_chapter_questions.py`) were making **real OpenAI calls** (401 with the stale/unconfigured key) and depended on the swallow-to-fallback bug for their 200 paths. Each now patches the AI boundary with a deterministic autouse fixture — which also removes live network calls from the suite.

## Acceptance criteria → evidence

1. *AI failure surfaces a structured error (like the TOC path)* — `test_generate_questions_openai_outage_503_not_200` patches `_make_openai_request` to raise and asserts 503 + `QUESTION_GENERATION_FAILED` through the full stack; same for bulk and single regenerate.
2. *Response tagged `is_fallback` so clients/tests can tell real from fallback* — route tests assert `is_fallback == true` on the empty-AI-response path and `false` on AI paths; mixed AI+fill responses tag per-question.
3. *Test patches `_make_openai_request` to raise and asserts no masquerade* — `test_generate_chapter_questions_reraises_ai_service_error` (service) + the route test in (1).

## Third-party review (opencode / GLM, pre-PR)

- **M1 rebutted**: "bulk regenerate deletes before generating, so an outage now leaves the chapter empty (503) where it previously got templates." The destructive ordering is pre-existing; the old outcome was templates masquerading as AI output with HTTP 200 — precisely this bug. User-authored content is never lost (`preserve_responses=true` keeps answered questions; `false` is an explicit discard), and the lost artifacts are reproducible machine output. Generate-then-swap needs a save/delete restructure — follow-up issue filed.
- **M2/M3/N4 fixed with tests**: single-regenerate outage 503 route test; `is_fallback` carry-through persisted assertion; mixed AI+fallback tagging test.
- **N5** (pre-existing hardcoded `chapter_title="Chapter"` in fallback fill) → folded into the follow-up issue. **N6** (pre-fix template questions read back as `is_fallback=False`) → documented accepted tradeoff, forward-looking only.

## Known limitations

- Bulk regenerate's delete-before-generate window (M1 above) — follow-up issue.
- Historical template questions persisted during pre-fix outages are indistinguishable from AI output (default `False`); no backfill.

## Tests

Backend **1101 passed / 13 skipped, 91.92% coverage** (gate ≥85%). Frontend untouched (additive response field only).


frankbria:
## Third-party review (opencode / GLM) — both rounds

**Round 1 (pre-PR)** — 1 Major, 2 Minor, 3 Nits:
- **M1** (bulk regenerate deletes before generating → outage now leaves the chapter empty with a 503 where it previously got silent templates): **rebutted** — the destructive ordering is pre-existing; the prior outcome was templates masquerading as AI output with HTTP 200, which is exactly the #182 bug; user-authored responses are never lost (`preserve_responses=true` keeps answered questions, `false` is an explicit discard); lost artifacts are reproducible machine output. Filed as **#234** (generate-then-swap, mirroring the single-question path's generate-before-mutate design).
- **M2** (no route test for single-question regenerate outage): **fixed** — `test_regenerate_single_question_openai_outage_503`, which also asserts the original question is untouched.
- **M3** (`is_fallback` carry-through on single regenerate untested): **fixed** — `test_regenerate_single_question_fallback_replacement_tagged` asserts both the response and the **persisted** Mongo doc.
- **N4** (mixed AI+fallback tagging untested): **fixed** — `test_mixed_ai_and_fallback_tagging`.
- **N5** (pre-existing hardcoded `chapter_title="Chapter"` in fallback fill): folded into **#234**.
- **N6** (pre-fix template questions read back `is_fallback=false`): documented accepted tradeoff — the tag is forward-looking, no backfill.

**Round 2 (post-PR, on the updated diff)** — **"No new blocking or high-severity findings… The diff is clean to merge."** The reviewer independently verified: `is_fallback` persistence via `model_dump()` (not just response echo), exception-handler ordering in both layers, that `_make_openai_request` propagates `AIServiceError` (validating the deep-patch tests), that the autouse AI fixtures don't mask error paths (outage tests override them with inner patches), and accepted the M1 rebuttal (its Finding 1 = #234, "No action required in this PR").

## Demo

`docs/demos/2026-07-06-issue-182-question-generation-errors.md` — real backend ×3 against real MongoDB: `main` vs fix with the **genuinely stale local OpenAI key** (true 401 outage, no simulation), plus the fix behind a wire-boundary stub for the responds-but-unusable cases. Old app: HTTP 200 with untagged fiction templates for a non-fiction book while its own log swallows the 401. New app: structured `QUESTION_GENERATION_FAILED`. Stub acts: `is_fallback: true` on all templates, `false` on real AI output.

PR #251, opened 2026-07-10
fix(ai): retry OpenAI calls at exactly one layer (#188)

### LINKED ISSUE #188: [P1.8] Nested retry wrappers can multiply OpenAI calls to max_retries^2
**Problem:** `_make_openai_request` already retries internally, but `generate_clarifying_questions` and `generate_table_of_contents` wrap it in a *second* `_retry_with_backoff`. On a persistent RateLimit/Timeout/InternalServer error the loops nest → up to `AI_MAX_RETRIES²` real OpenAI calls plus additive backoff sleeps — multiplying cost, latency, and (with P0.3) total stall time.

**Evidence:** `backend/app/services/ai_service.py:207` (inner retry), `:305` and `:546` (outer `_retry_with_backoff(self._make_openai_request, ...)`).

**Acceptance criteria:**
- Retry happens at exactly one layer. Call `_make_openai_request` directly at these two sites (or strip its internal retry) — not both.
- Test asserts a persistently-failing call makes at most `AI_MAX_RETRIES` requests.

frankbria:
Closes #188.

## Problem
`generate_clarifying_questions` and `generate_toc_from_summary_and_responses` wrapped `_make_openai_request` — which already retries internally via `_retry_with_backoff` — in a **second** `_retry_with_backoff` (`ai_service.py:314`/`:555` on main).

## Premise correction (verified before fixing)
The feared `AI_MAX_RETRIES²` call multiplication does **not** occur on main — but only by accident: the inner layer converts every OpenAI error into an `AIServiceError` subclass, and the outer wrapper's `except AIServiceError: raise` re-raises without retrying. The nesting still caused real defects today:

1. A spurious extra "Attempting API call (attempt 1/N)" log line per request (4 log records for 3 real calls).
2. The retry/backoff logs carried a **different** correlation id than the calling request (the outer layer consumed the caller's id; the inner minted its own), splitting the trace.
3. The non-squaring depended fragilely on the inner layer never leaking a raw `openai.*` exception — any future edit to the error conversion would silently reintroduce squaring.

## Fix
- Both sites call `_make_openai_request` directly (the AC's first branch — the other 6 call sites already rely on the internal retry, so stripping the internal layer would have touched 8 sites).
- `_make_openai_request` gains an optional `correlation_id` threaded to its internal `_retry_with_backoff`, so attempt/backoff logs stay correlated with the request. Other call sites unaffected (param defaults to `None` → fresh uuid, same as before).

## Tests
New `TestSingleRetryLayer` (4 tests):
- AC pins: persistent `RateLimitError` → exactly `AI_MAX_RETRIES` real requests, for both sites.
- Re-wrap guards (RED on main): exactly one "Attempting API call" log per real request; all attempt logs share one correlation id.

`test_ai_service_error_handling` re-targeted its mock from `_make_openai_request` (a seam that pinned the removed outer conversion layer) to the real OpenAI boundary `client.chat.completions.create` — same assertions, now exercising the real `except Exception → AI_UNEXPECTED_ERROR` conversion.

Backend: **1148 passed / 13 skipped, 92.21% cov**; ruff findings in touched files are all pre-existing on main (0 on added lines).

## Reviews
- Pre-PR opencode (GLM): **"clean to merge"** — independently verified the seam move is "stronger, not weaker", the correlation threading, and that no double-wrap remains; its one informational minor (comment distinguishing AC-pin tests from re-wrap guards) adopted.

## Known limitations
- The AC call-count tests also pass on the old code (squaring never actually occurred); the log-based tests are the true re-wrap guards — documented in the test class docstring.


frankbria:
## Post-PR third-party review — opencode (GLM), fresh session

**Verdict: clean to merge.** Independent verification (did not trust the PR description):

1. **No nested retry path remains** — the only production caller of `_retry_with_backoff` is `ai_service.py:226` (inside `_make_openai_request`); all 8 callers of `_make_openai_request` invoke it directly.
2. **correlation_id threading correct** — both fixed methods mint → thread → reuse; the optional param breaks none of the 9 existing test patches.
3. **Guards are genuine** — the log-count and single-correlation-id tests fail if a second retry layer is reintroduced (an outer `_retry_with_backoff` consumes `correlation_id` without forwarding, so the inner re-mints → 2 ids → fail).
4. **Seam move is strictly stronger** — `test_ai_service_error_handling` now exercises the real raw-exception→`AI_UNEXPECTED_ERROR` conversion instead of stubbing past it.
5. **No consumer-visible behavior change** — error types, real call counts, and timing unchanged; the issue's `max_retries²` was latent (could not manifest on main), the fix removes the fragile accident that prevented it.

**Minors (non-blocking, pre-existing/out of scope):** other AI methods don't pass a `correlation_id` (log-parity opportunity, not a regression); dead unreachable fallback at `ai_service.py:179-185` predates this PR.

Issue #265, opened 2026-07-10
[P1.13] PUT /users/{auth_id} skips sanitize_input — inconsistent stored-XSS posture with PATCH /users/me


frankbria:
Found by the opencode (GLM) pre-PR review of #244 — pre-existing, not introduced there.

`update_profile` (PATCH /users/me) runs string fields through `sanitize_input` before persisting; `update_user_data` (PUT /users/{auth_id}) writes the same document fields with no sanitization. Profile strings written via PUT (self or admin) are stored unsanitized — inconsistent stored-XSS/injection posture between two endpoints writing the same document.

AC:
- PUT /users/{auth_id} sanitizes string fields the same way PATCH /users/me does (shared helper preferred).
- Test: a string field containing markup is stored identically through both endpoints.

Issue #220, opened 2026-07-04
[P0.2.1] Stripe customer/subscription model + webhook (raw-body HMAC verify)


frankbria:
Parent epic: #174. Decision record: `docs/adr/2026-07-04-beta-entitlement-model.md`. Deferred from the free-invite beta; out of scope until paid launch is scheduled.

Persist a Stripe customer id + subscription state on the user, and add a webhook endpoint (extend the placeholder `backend/app/api/endpoints/webhooks.py`) that verifies the Stripe signature against the **raw request body** (HMAC; no raw-body/signature verification exists yet). On `customer.subscription.*` events, map the Stripe price → a `plan` in `app/core/entitlements.py` and update the user's `plan`.

AC: signed webhook updates `user.plan`; invalid signature → 400; replay/idempotency handled.

PR #266, opened 2026-07-10
fix(security): PATCH /users/me can no longer self-elevate role — privileged fields unwritable via API (#244)

### LINKED ISSUE #244: [P0.9] Privilege escalation: PATCH /users/me accepts role — any user can self-elevate to admin
Found during the #220 post-PR review (opencode/GLM, PR #243), pre-existing and unrelated to that change.

`UserUpdate` (`backend/app/schemas/user.py`) exposes `role: Optional[str]`, and `PATCH /api/v1/users/me` (`backend/app/api/endpoints/users.py`, `update_profile`) dumps the schema with `exclude_unset=True` straight into the generic `update_user()` `$set` — so any authenticated user can send `{"role": "admin"}` and persist it. Anything gated on `role` (e.g. `SessionRoleChecker`) is then escalatable by self-service.

AC:
- `role` removed from `UserUpdate` (or ignored/stripped by the /me handler); admin role changes require an admin-only path.
- Regression test: authenticated `PATCH /users/me` with `{"role": "admin"}` does not change the stored role.
- Audit any other privileged fields reachable via the same schema.

frankbria:
Closes #244 ([P0.9] Privilege escalation: PATCH /users/me accepts role).

## Root cause
`UserUpdate` exposed `role: Optional[str]`, and `PATCH /api/v1/users/me` dumped `model_dump(exclude_unset=True)` straight into the generic `update_user()` Mongo `$set` — any authenticated user could persist `{"role": "admin"}` and everything gated on `SessionRoleChecker` was self-escalatable.

## Fix (root cause, not symptom)
- **`role` removed from `UserUpdate`** (and its OpenAPI example) — privileged fields follow the #220 policy: not API-writable.
- **`extra="forbid"` on `UserUpdate`** — the declared fields are now the single writable allowlist, enforced by pydantic before any handler runs. An unknown/privileged key (`role`, `plan`, `is_active`, `stripe_*`, `book_ids`, `auth_id`) is a **loud 422**, never a silent `$set` — closing the mass-assignment *class*, not just the `role` instance, and preserving the audit signal (escalation attempts are visible as 4xx).
- **Dead non-admin role guard deleted** from `PUT /users/{auth_id}` (it referenced the removed field). No API path writes `role` for anyone — admins included; no admin role-change UI/script exists (verified), so roles are managed directly in the database.

## Privileged-field audit (AC 3)
`plan`/`stripe_customer_id`/`stripe_subscription_id`/`is_active`/`book_ids`/`auth_id` were already deliberately absent from `UserUpdate` (#220) and are now also forbid-guarded. `metadata` is stored under its own key and nothing in `app/` ever reads `user["metadata"]` (authz or otherwise). `email` stays self-service (duplicate-key guarded, pre-existing behavior).

## Tests
- Parameterized regression: `PATCH /users/me` with each privileged field → **422 + stored document byte-identical** (before/after full-doc compare, real Mongo).
- `PUT /users/{auth_id}` with `role` → 422 + nothing persisted; **admin-caller variant** pins that no API path writes role.
- `test_put_user_non_admin_role_change_returns_403` was superseded: the 403 guard is gone with the field; replaced by the strictly stronger unwritable-pin above (the 403 only proved rejection, not unwritability).
- Backend **1157 passed / 13 skipped, 92.21% coverage**; ruff clean; pre-commit gates green.

## Review
opencode (GLM) pre-PR review ×2: round 1 — 1 Major (mass-assignment class) **adopted** via `extra="forbid"`; its duplicate handler-level allowlist **rebutted** (schema is the single source of truth, pydantic already enforces at the boundary); sibling-field + admin-context test minors adopted; metadata minor rebutted with grep evidence. Round 2: **clean to merge** after a one-word doc fix (applied). Pre-existing `PUT` sanitize_input gap filed as #265.

## Known limitations
- `extra="forbid"` is a strict contract change: clients sending undeclared keys now get 422 instead of silent-ignore. The only production caller (`useProfileApi.ProfileUpdateData`) sends declared keys exclusively (verified); full suite green.
- Admins can no longer change roles via the API (no consumer existed); role changes are a direct-DB operation until a real admin surface is needed.

### DIFF
--- a/backend/app/api/endpoints/users.py
+++ b/backend/app/api/endpoints/users.py
@@ -310,13 +310,6 @@ async def update_user_data(
             status_code=status.HTTP_404_NOT_FOUND, detail="User not found"
         )
 
-    # Prevent non-admin users from changing their role
-    if user_update.role and current_user["role"] != "admin":
-        raise HTTPException(
-            status_code=status.HTTP_403_FORBIDDEN,
-            detail="Only admins can change user roles",
-        )
-
     # Update with current timestamp
     update_data = {
         k: v
--- a/backend/app/schemas/user.py
+++ b/backend/app/schemas/user.py
@@ -91,9 +91,17 @@ class UserUpdate(BaseModel):
     bio: Optional[str] = Field(None, max_length=1000)
     preferences: Optional[UserPreferences] = None
     metadata: Optional[Dict[str, Any]] = None
-    role: Optional[str] = None
+    # role is deliberately absent (#244): it flows into a generic $set in the
+    # /users handlers, so exposing it here lets any user self-elevate to admin.
+    # Same policy as plan/stripe_* (#220) — privileged fields are not
+    # API-writable; roles are managed directly in the database.
 
     model_config = ConfigDict(
+        # extra="forbid" (#244): these fields ARE the writable allowlist for the
+        # generic $set in the /users handlers. Unknown keys (role, plan, …) are
+        # a loud 422, not a silent ignore — adding a privileged field here is
+        # the only way to make it writable, and that takes a review.
+        extra="forbid",
         json_schema_extra={
             "example": {
                 "email": "updated@example.com",
@@ -108,7 +116,6 @@ class UserUpdate(BaseModel):
                     "marketing_emails": False,
                 },
                 "metadata": {"preferences": {"theme": "dark"}},
-                "role": "admin",
             }
         }
     )
--- a/backend/tests/test_api/test_routes/test_users_coverage.py
+++ b/backend/tests/test_api/test_routes/test_users_coverage.py
@@ -162,13 +162,68 @@ async def test_put_user_other_non_admin_returns_403(auth_client_factory):
     assert "permissions" in resp.json()["detail"]
 
 
-async def test_put_user_non_admin_role_change_returns_403(auth_client_factory):
+# ---------------------------------------------------------------------------
+# Issue #244 — privileged fields are not writable via the API
+# UserUpdate is extra="forbid", so an unknown/privileged key is a 422 and the
+# handler (and its generic $set) never runs.
+# ---------------------------------------------------------------------------
+PRIVILEGED_FIELDS = [
+    ("role", "admin"),
+    ("plan", "pro"),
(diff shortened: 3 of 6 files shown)


frankbria:
## Post-PR review — opencode (GLM), fresh session

**Verdict: approve; 1 Major to verify (resolved) + 3 minors.**

- **Core fix correct**: `role` removed from `UserUpdate` + `extra="forbid"` closes the escalation at the schema layer (root fix, not a handler strip). Tests assert **stored DB state**, not just HTTP status. No remaining escalation path through `UserUpdate`.
- **M1 (breaking contract of `extra="forbid"`)** — **verified clean**: the only production caller of PATCH `/users/me` is `useProfileApi.updateUserProfile`, whose `ProfileUpdateData` sends exclusively declared keys (`first_name`/`last_name`/`display_name`/`bio`/`preferences`); there is no frontend PUT `/users/{auth_id}` caller; avatar is a separate multipart endpoint. Added a positive happy-path test (`test_patch_me_declared_fields_still_persist`) so a valid multi-field update is pinned at 200 + persisted.
- **m2 (no positive test)** — **adopted** (test above).
- **m3 (`before` null-check)** — **adopted**: `assert before is not None` fail-fasts if the factory doesn't seed.
- **m1/metadata** — **rebutted with grep evidence**: nothing in `app/` reads `user["metadata"]` for authz (only initialized to `{}` at creation). Out of scope to retype here.

Round-1 Major (mass-assignment *class*) drove the `extra="forbid"` change; its suggested duplicate handler-level allowlist was rebutted (the schema is the single source of truth pydantic already enforces at the boundary). Pre-existing PUT `sanitize_input` gap filed as #265 (now [P1.13]).

**Demo** (`docs/demos/2026-07-10-issue-244-role-self-elevation.md`, `showboat verify` green): two real uvicorn servers + real Mongo + a genuine seeded better-auth session — `main` self-elevates to `admin` (200, stored role `admin`); the branch rejects with 422 and the stored role stays `user`; a normal profile edit still round-trips.

Backend **1157 passed / 13 skipped, 92.21% cov**.

Issue #176, opened 2026-07-04
[P0.4] Make production security fail-safes fire (NODE_ENV vs ENVIRONMENT mismatch)


frankbria:
**Problem (hand-verified):** All backend production guards check `os.getenv("NODE_ENV")`, but PM2 sets only `ENVIRONMENT` on the backend process (not `NODE_ENV`). So in the real deployment: `BYPASS_AUTH=true` would NOT block startup (only a warning), the built-in CI test secret would be accepted as valid, and the <64-char secret warning never fires.

**Evidence:** `backend/app/core/config.py:102,126,143` and `backend/app/main.py:33` all gate on `NODE_ENV`; `ecosystem.config.template.js:12-13` backend env sets `ENVIRONMENT: '__ENVIRONMENT__'` with no `NODE_ENV` (the frontend block at :29-30 sets `NODE_ENV: 'production'`).

**Acceptance criteria:**
- Guards read the variable the deployment actually sets (`ENVIRONMENT`/`settings`), or PM2 sets `NODE_ENV=production` on the backend.
- A test asserts `BYPASS_AUTH=true` raises at startup when the deploy marker is production, and the CI test-secret is rejected in production.

**Depends on:** none.

PR #271, opened 2026-07-10
fix(security): production auth-bypass guard no longer exempted by generic CI env var (#192)

### LINKED ISSUE #192: [P1.12] Production auth-bypass guard is keyed on the generic CI env var
**Problem:** The middleware FATAL guard only fires when `bypassAuth && NODE_ENV==='production' && !isCI` where `isCI = process.env.CI === 'true'`. `CI=true` is set by most CI/PaaS/container runtimes, so a production artifact that happens to run with `CI=true` and `BYPASS_AUTH=true` silently disables auth — exactly what the guard was meant to prevent. The E2E infra deliberately drives a prod build through this hole.

**Evidence:** `frontend/src/middleware.ts:19-26`; `frontend/playwright.config.ts:85,91,102-103`.

**Acceptance criteria:**
- Gate bypass on a purpose-built flag (e.g. `E2E_ALLOW_BYPASS=1`) never present in real deploys, and/or require `NEXT_PUBLIC_ENVIRONMENT==='test'`.
- The guard stays fatal for `NODE_ENV==='production'` regardless of `CI`.

frankbria:
Closes #192.

## Problem

`frontend/src/middleware.ts` had a FATAL guard meant to prevent `BYPASS_AUTH=true` from ever running in production — but it was exempted whenever `CI=true`. Since `CI=true` is set by most CI/PaaS/container runtimes, a production artifact that happened to run with both vars **silently disabled all authentication**. The E2E infrastructure deliberately drove a production build (`npm run build && npm start`) through this hole by forwarding `CI`.

## Fix

- **`middleware.ts`**: the only exemption is now the purpose-built `E2E_ALLOW_BYPASS=1` flag. The generic `CI` env var is no longer consulted anywhere in the guard — it stays fatal for `NODE_ENV==='production'` regardless of `CI`.
- **`playwright.config.ts`**: the webServer env sets `E2E_ALLOW_BYPASS: '1'` explicitly, so the CI prod-build E2E run opts in with a flag that no real deploy sets.
- Dev-mode behavior unchanged: `BYPASS_AUTH=true` alone still bypasses when `NODE_ENV !== 'production'` (documented local `npm run dev` workflows).

## Acceptance criteria

- ✅ Bypass gated on a purpose-built flag (`E2E_ALLOW_BYPASS=1`) never present in real deploys — the AC's "and/or" second option (`NEXT_PUBLIC_ENVIRONMENT==='test'`) was deliberately not taken: keying a security guard on a general-purpose env var re-creates the exact defect class this issue removes.
- ✅ Guard stays fatal for `NODE_ENV==='production'` regardless of `CI` — pinned by test.

## Tests

`middleware.test.ts` gains a production-guard describe (6 tests):
- `CI=true` no longer exempts production (**the regression pin for this bug**)
- production + `BYPASS_AUTH=true` with no flag → throws
- only `E2E_ALLOW_BYPASS=1` exempts; the loose value `'true'` still throws
- the flag alone is **not** a bypass — `BYPASS_AUTH` is still required (defense-in-depth pin)
- dev/test-mode `BYPASS_AUTH` behavior unchanged

Frontend: **114 suites, 2071 passed / 8 skipped**, coverage gates green; lint 0 errors; typecheck clean.

## Reviews

- opencode (GLM) pre-PR: **"clean to merge"** — no Critical/Major; its 2 Minor test-gap suggestions (flag-alone-is-not-a-bypass, strict-value pin) adopted; its optional `NEXT_PUBLIC_ENVIRONMENT` belt-and-suspenders rejected as re-introducing a generic-env-var dependency.

## Known limitations

- Client-side `NEXT_PUBLIC_BYPASS_AUTH` (`ProtectedRoute`, dashboard E2E mode) is a separate build-time mechanism, out of scope per the issue.
- The backend `BYPASS_AUTH` production guard (hardened in #176) is a separate system, already independent of `CI`.


frankbria:
## Post-PR third-party review — opencode (GLM), fresh session

**Verdict: clean to merge.** No Critical/Major. Full verification trace: real-deploy shape fails closed (FATAL → 500); CI E2E prod-build path keeps working (flag present at build+runtime via the Playwright webServer env); both acceptance criteria confirmed; all test assertions verified correct incl. env-restore isolation.

**Minors (all non-blocking, triaged):**
- **M1** — dev-mode (`NODE_ENV !== 'production'`) bypass still keys on `BYPASS_AUTH` alone. Deliberate (documented local `npm run dev` workflows), but filed as a P2 hardening follow-up per the reviewer's recommendation: require the purpose-built flag in all environments.
- **M2** — flag absence in deploys is conventional, not enforced; the deploy workflow's independent `BYPASS_AUTH` gate is the structural backstop. Informational.
- **M3** — Edge build-time env inlining would only *strengthen* the fix for real deploys (flag absent at build ⇒ not injectable at runtime). Informational.

PR #274, opened 2026-07-11
fix(security): PUT /users/{auth_id} sanitizes string fields like PATCH /users/me (#265)

### LINKED ISSUE #265: [P1.13] PUT /users/{auth_id} skips sanitize_input — inconsistent stored-XSS posture with PATCH /users/me

frankbria:
Closes #265 (P1.13, security — found by the opencode pre-PR review of #244).

## Problem
`update_profile` (PATCH /users/me) runs string fields through `sanitize_input` before persisting; `update_user_data` (PUT /users/{auth_id}) wrote the same document fields with no sanitization — inconsistent stored-XSS/injection posture between two endpoints writing the same document.

## Fix
- New shared `sanitize_string_fields()` helper in `users.py` — byte-identical to the comprehension PATCH already used inline.
- PATCH refactored onto the helper (behavior unchanged); PUT applies it after its existing None-filter (semantics preserved: non-strings pass through untouched, exactly as in PATCH).

## Tests (TDD)
- `test_put_sanitizes_strings_same_as_patch` (real Mongo): a markup-bearing bio is stored identically through both endpoints. Failed RED before the fix (PUT stored raw `<script>`), passes after. An intermediate PUT with a different value guards against a tautological pass; a `"<" not in via_patch` assert pins that PATCH actually sanitizes.
- Backend: **1159 passed / 13 skipped, 92.21% coverage**; ruff clean.

## Review
opencode (GLM) pre-PR: **"clean to merge"** — verified helper equivalence, None-filter preservation, and test soundness. Its one non-blocking note (nested `preferences` strings unsanitized on both endpoints) was verified to have **no real surface**: every string field in the `UserPreferences` schema is a `Literal` enum, so markup cannot pass validation.

## Known Limitations
- `sanitize_input` only touches top-level string fields — parity with PATCH by design (and nested prefs are Literal-validated, see above).


frankbria:
## Third-party post-PR review

**opencode (GLM)** pre-PR session: **"clean to merge"** — verified `sanitize_string_fields` is byte-identical to PATCH's old inline comprehension (no PATCH behavior change), PUT's None-filter semantics are preserved (non-strings pass through untouched, same as PATCH), and the regression test is sound, not tautological (the intermediate PUT with a different value forces a genuine overwrite; `"<" not in via_patch` pins that PATCH actually sanitizes). Its one non-blocking note — nested `preferences` strings unsanitized on both endpoints — was verified to have **no surface**: every string field in the `UserPreferences` schema is a `Literal` enum, so markup cannot pass validation.

**codex** post-PR fresh session (opencode's fresh session hung with zero output — known concurrent-session contention, fallback per review invariants): **"I did not find a discrete regression introduced by this diff"** — confirmed the sanitization parity and preserved None-filtering.

Demo with stored-document evidence: `docs/demos/2026-07-10-issue-265-put-sanitize.md` (`showboat verify` green).

PR #288, opened 2026-07-13
[P2.12] fix(profile): profile save no longer wipes Settings-page preferences (#204)

### LINKED ISSUE #204: [P2.12] Profile-page save silently wipes all Settings-page preferences
**Problem:** `ProfilePage.onSubmit` PATCHes `/users/me` with `preferences: {theme, email_notifications, marketing_emails}` — only 3 fields, without merging the full preferences object. The backend does `model_dump(exclude_unset=True)` then `$set: {preferences: ...}`, replacing the entire Mongo `preferences` subdocument. So a beta user who edits their bio/name on the Profile page after configuring Writing/Export/Notification settings silently loses `default_writing_style`, `auto_save_interval`, `default_export_format`, `default_page_size`, `include_empty_chapters`, and the notification flags — with a cheerful "Profile updated" toast.

**Evidence:** `frontend/src/app/profile/page.tsx:112-124`; `backend/app/api/endpoints/users.py:109`; `backend/app/db/user.py:48-50`. (The Settings page already implements the documented merge-before-save contract; Profile doesn't.)

**Acceptance criteria:**
- ProfilePage loads the full `UserPreferences` and merges before every PATCH, sending the complete object (matching the Settings-page contract).
- A test asserts a profile save preserves unrelated preference fields.

frankbria:
Closes #204.

## Problem

`ProfilePage.onSubmit` PATCHed `/users/me` with a 3-field `preferences` object (`theme`, `email_notifications`, `marketing_emails`). The backend `$set`-replaces the entire `preferences` subdocument, so a profile save silently wiped every Settings-page preference (`default_writing_style`, `auto_save_interval`, `default_export_format`, `default_page_size`, `include_empty_chapters`, notification flags) — with a cheerful "Profile updated" toast.

## Fix (frontend-only, per the CodeRabbit plan on the issue — design choice 2: merge + load guard)

- The page retains the full loaded `UserPreferences` and merges the three editable fields into it on every save, matching the documented Settings-page contract.
- `invalidateUserPreferencesCache(updated?.preferences ?? null)` after a successful save (mirrors Settings), and the retained object refreshes from the PATCH response.
- Save is disabled until preferences load (`'loading' | 'loaded' | 'error'`); a failed load shows an alert and keeps saving disabled, closing the failure-path variant of the same wipe.
- **Pre-existing bug fixed along the way**: the hydration dirty-guard read `form.formState.isDirty` in an async callback, but RHF's formState is a lazy Proxy — `isDirty` was never computed (nothing read it during render), so the guard was always `false` and a slow fetch clobbered in-progress edits. Dirty tracking is now subscribed during render, and hydration uses `reset(values, { keepDirtyValues: true })` so untouched fields hydrate while user edits survive (the pre-PR codex P2: an all-or-nothing skip would have left `bio: ''` defaults that a save writes over stored data).

## Tests

New `ProfilePageSave.test.tsx` (real react-hook-form + real `useProfileApi` over a mocked `useAuthFetch`, mirroring `SettingsPageSave.test.tsx`):
1. Save sends the full merged preferences object — unexposed Settings fields (incl. a `future_flag`) survive.
2. `invalidateUserPreferencesCache` called after a successful save.
3. Save disabled while loading and after a failed load (alert shown, no PATCH possible).
4. Dirty path: an edit made before hydration resolves survives (`first_name: 'Zed'`), untouched fields still hydrate (`bio`, `display_name`), and the full preferences still merge.

All behavior pins **mutation-verified RED**: removing the merge spread fails tests 1+4; removing the dirty-tracking subscription fails test 4; removing `keepDirtyValues` fails test 4.

Legacy `ProfilePage.test.tsx` (10 tests, mocks `useProfileApi` without `getUserProfile`) stays green — the page treats an absent hook as loaded (test-only path; the hook is always present in production). E2E `profile-editing.spec.ts` 3/3 passed locally (chromium).

Frontend: 117 suites, 2122 passed / 5 skipped; lint 0 errors; typecheck clean; coverage gates green.

## Reviews

- opencode (GLM) was occupied by a foreign-cwd delegation (the documented hang condition) → **codex fallback pre-PR**: 1 real P2 found and fixed (`keepDirtyValues`, above). Post-PR fresh-session review to follow as a PR comment.

## Known limitations

- If the user edits a preference control before hydration resolves, the three editable preference fields save with what the form shows (WYSIWYG); unexposed fields are still preserved.
- Backend `PATCH /users/me` still replaces the whole subdocument by design — every client must merge before save (Settings and Profile now both do).


frankbria:
## Third-party review record

**opencode (GLM)** was occupied by a foreign-cwd delegation both rounds (the documented concurrent-session hang condition) → **codex fallback**, pre-PR + post-PR fresh sessions.

### Pre-PR (codex) — 1 finding, fixed
- **[P2] Dirty hydration left backend-only fields unhydrated** (`profile/page.tsx`): once the dirty-guard actually worked, a user typing before the fetch resolved skipped hydration entirely, leaving `bio`/`display_name` at session defaults (`''`) that a save would write over stored data — the same wipe class on top-level fields. **Fixed** in cff8d2c with `form.reset(values, { keepDirtyValues: true })` (hydrate untouched fields, retain edits); pinned by the dirty-path test asserting `bio`/`display_name` survive (mutation-verified RED).

### Post-PR fresh session (codex) — 1 finding, fixed
- **[P2] Save guard not re-armed on user switch** (`profile/page.tsx:75`): if `user?.id` changed after a successful load, `loadState` stayed `'loaded'` with the prior account's retained preferences — a save during the new load could PATCH them into the new account. **Fixed** in 0f68e5d: the hydration effect resets `loadState`/`preferences` at the top; pinned by a user-switch test (mutation-verified RED).

Both fixes landed with tests; suite now 15 profile tests across the two files, all green (117 suites / 2123 passed overall).

Issue #200, opened 2026-07-04
[P2.8] Remove/repair orphaned 'gold standard' tests that run nowhere or import missing modules


frankbria:
**Problem:** Several flagship tests inflate perceived coverage but validate nothing: `SystemIntegration.test.tsx` imports nonexistent `@/lib/api/aiClient` and is jest-excluded; `SystemE2E.test.tsx` sits outside Playwright's `testDir` and runs in no suite; `TocGenerationWizard.test.tsx` asserts `expect(true).toBe(true)` with the import commented out; `test_draft_generation_api.py` has 0 asserts and prints "READY!".

**Evidence:** `frontend/src/__tests__/SystemIntegration.test.tsx:19`; `frontend/src/__tests__/e2e/SystemE2E.test.tsx:18`; `frontend/src/__tests__/TocGenerationWizard.test.tsx:30`; `backend/tests/test_draft_generation_api.py:14`.

**Acceptance criteria:**
- Each is either rewritten against the real API surface and actually collected/run, or deleted. No test named "gold standard" that executes nowhere.

PR #294, opened 2026-07-15
[P2.17] fix(summary): gate on the real readiness threshold, drop contradictory word/char copy (#218)

### LINKED ISSUE #218: [P2.17] Summary readiness gate: contradictory word/char copy + threshold mismatch causes a TOC dead-end
**Problem:** The summary page enforces `MIN_SUMMARY_LENGTH = 30` **characters** and the counter says "Minimum: 30 characters", but the guideline text two lines below says "Minimum 30 **words** recommended", and the AI readiness check (surfaced in `NotReadyMessage`) actually wants "at least 500-1000 words". A user passes the weak client gate with a 30-char summary, clicks "Continue to TOC Generation", then gets bounced to the NOT_READY screen — a confusing dead-end that reads as a bug.

**Evidence:** `frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:9,219,225`; `frontend/src/components/toc/NotReadyMessage.tsx:128`.

**Acceptance criteria:**
- Fix the words/characters copy mismatch and raise the client gate (or add a soft word-count warning) so it's consistent with the real AI readiness threshold before the user invests time.

frankbria:
Closes #218.

## The problem

The summary page enforced `MIN_SUMMARY_LENGTH = 30` **characters**, the counter said "Minimum: 30 characters", the guideline text two lines below said "Minimum 30 **words** recommended", and the NOT_READY screen claimed "at least 500-1000 words". A user cleared the weak client gate with a 30-character summary, clicked "Continue to TOC Generation", and got bounced to NOT_READY — a dead-end that reads as a bug.

## The real gate (verified, not assumed)

Only one threshold is enforced anywhere, and it is identical on both backend paths:

```python
word_count >= 30 and char_count >= 150
```
- `backend/app/api/endpoints/books.py` — deterministic fallback (no AI analysis)
- `backend/app/services/ai_service.py` — AI-analysis path

The AI does **not** decide readiness. The wizard gates on `meets_minimum_requirements` alone (`TocGenerationWizard.tsx`), ignoring `is_ready_for_toc`. So four numbers described "long enough" to the user and only one was real:

| Location | Claim | Enforced? |
|---|---|---|
| backend (both paths) | 30 words + 150 chars | **yes** |
| `summary/page.tsx` `MIN_SUMMARY_LENGTH` | 30 **characters** | client-only, far too weak |
| `summary/page.tsx` guideline copy | "Minimum 30 words recommended" | no |
| `NotReadyMessage.tsx` | "at least 500-1000 words" | no — fiction |
| `bookClient.ts` doc comment | "Need: 100+ words." | no — fiction |

The page's `30` and the backend's `30` were the same digits in **different units** — a coincidence that invites a wrong fix.

## The fix

New `frontend/src/lib/constants/summary-readiness.ts` mirrors the backend gate as a documented SSOT (the `auto-save.ts` / `writing-styles.ts` pattern). The page delegates to it, the counter shows words **and** characters against both real minimums, and every fictional number is replaced by the enforced one.

Chose a **hard client gate at the real threshold** over the AC's alternative soft warning: a warning still lets the user walk into the dead-end.

## Counting must match Python exactly, or the bug returns

The gate is only as good as its arithmetic, and two JS/Python counting differences each recreate the dead-end. Both were verified against the real runtimes, not reasoned about:

**Whitespace** — JS `/\s/` is not Python's whitespace set:
```
len("word1\x85word2".split()) == 2   but JS /\s+/ -> 1   (falsely rejects a valid summary)
len("word1﻿word2".split()) == 1 but JS /\s+/ -> 2   (accepts one the backend rejects)
```
A BOM is common in pasted text. The module now spells out Python's exact `str.isspace()` set.

**Characters** — JS counts UTF-16 code units, Python counts code points (**codex pre-PR P2**):
```
len('👍' * 100) == 100          (python)
'👍'.repeat(100).length === 200 (js)
```
A mostly-emoji summary read as 200 chars, cleared the 150 gate, then failed the backend's 100. Now `Array.from(text).length`.

## No work is lost

The debounced auto-save persists independently of validation, so a user under the stricter gate keeps their draft and simply cannot proceed. Pinned by test.

## Tests

**33 pins across 3 suites** — the first tests this page has ever had. Every behavior pin RED-verified at birth, plus mutation checks:

| Mutation | Result |
|---|---|
| restore the old char-only 30 gate | 3 page pins fail |
| restore "500-1000 words" | 2 NotReadyMessage pins fail |
| `split(/\s+/)` instead of Python's set | 6 parity pins fail |
| `text.length` instead of code points | 2 emoji pins fail |

Frontend: **121 suites, 2181 passed / 5 skipped**. Coverage gates green; committed without `--no-verify`.

## Reviews

opencode (GLM) hung 10 min with zero output (the documented concurrent-session condition) → **`codex review` fallback**: round 1 found the emoji P2 above (verified, adopted, pinned); round 2 on the updated diff: *"No actionable regressions were identified. The client-side readiness gate and related copy now align with the backend TOC minimums."*

## Known limitations

- `frontend/src/components/SummaryInput.tsx` carries the identical "Minimum 30 words recommended" copy but has **zero importers** — orphaned code whose only consumer is its own test file (the #200 class). Out of scope here since it renders to no user; filed as a follow-up.
- Backend and frontend now encode the same thresholds in two places. There is no shared-constant mechanism across the stack in this repo; the module documents the backend files it mirrors, and the parity tests fail loudly if either drifts.


frankbria:
## Review triage

**Pre-PR (local gate)** — opencode (GLM) hung 10 min with zero output (the documented concurrent-session condition) → fell back to `codex review`:

| Round | Finding | Disposition |
|---|---|---|
| 1 | **[P2]** `text.length` counts UTF-16 code units; the backend's `len()` counts code points. A mostly-emoji summary reads as 200 chars client-side, clears the 150 gate, then fails the backend's 100 — recreating the #218 dead-end. | **Adopted.** Verified in both runtimes first (`len('👍'*100) == 100` vs `'👍'.repeat(100).length === 200`), fixed with `Array.from().length` across the gate, the max check, and the counter. 2 pins, both RED under the old code. Demonstrated live: backend counts 106, main's counter reads 152 and enables submit, branch reads 106 and blocks. |
| 2 | *"No actionable regressions were identified. The client-side readiness gate and related copy now align with the backend TOC minimums."* | — |

**Post-PR (fresh session)** — `codex review`: *"The changes align the summary page gate and NOT_READY copy with the backend readiness thresholds, and the added counting helpers appear to preserve backend parity for the covered boundary cases. I did not identify any introduced, actionable correctness issues in the modified runtime code."*

**GLM CI review ×2** — "no defects found", having independently verified the load-bearing claim against the checked-out source: `books.py:1148-1150` and `ai_service.py:473-483` both compute `word_count = len(summary.split())`, `char_count = len(summary)` and gate on `>= 30 and >= 150`.

**Self-found, fixed before review**: JS `/\s/` is not Python's `str.isspace()` set, and the divergence cuts both ways — `\x85`/`\x1c-\x1f` (Python splits, JS doesn't → falsely rejects a valid summary) and `﻿` BOM (JS splits, Python doesn't → accepts one the backend rejects, i.e. the dead-end returns). A BOM is common in pasted text. The module now spells out Python's exact set; 6 pins, all RED under `/\s+/`.

Both review rounds converged on the same class of defect — client/server counting drift — which is exactly what this issue is about. The gate is now only as trustworthy as its arithmetic, so the arithmetic is pinned to Python's semantics in both dimensions.

**Follow-up filed**: #295 [P3.15] — `SummaryInput.tsx` carries the identical stale "Minimum 30 words recommended" copy but has zero importers (the #200/#209 dead-code class). Out of scope here since it renders to no user.

**Demo**: `docs/demos/2026-07-14-issue-218-summary-readiness-gate.md` — real backend + Mongo + genuine better-auth signup, branch :3000 vs pristine main worktree :3001, same session/book. All ACs mapped to outcome evidence; DEMO_PASSED = true.

Issue #187, opened 2026-07-04
[P1.7] Batch question-response save skips the ownership/existence check


frankbria:
**Problem:** `save_question_responses_batch` only checks fields are non-empty then upserts by `(question_id, user_id)` — it never verifies the `question_id` exists or belongs to the URL's book/chapter, unlike the single-response endpoint. A stale/foreign/mistyped `question_id` is reported `success:true` but creates an orphaned/misattributed response the normal read path never surfaces.

**Evidence:** `backend/app/db/questions.py:664-780`; contrast `backend/app/services/question_generation_service.py:271-288` (single-response validates book/chapter membership).

**Acceptance criteria:**
- Batch-fetch referenced `question_id`s scoped to `user_id`+`book_id`+`chapter_id`; reject/flag any not found.
- Test sends a foreign question_id and asserts rejection (not a silent orphan).

PR #296, opened 2026-07-15
[P2.18] fix(questions): collapse repeated question_id within one batch response save (#242)

### LINKED ISSUE #242: [P2.18] Batch question-response save writes duplicate documents for a repeated question_id within one batch
**Found by**: opencode (GLM) post-PR review of #241 (issue #187) — pre-existing, out of scope there.

**Problem**: `save_question_responses_batch` (`backend/app/db/questions.py`) processes items independently. If one batch contains the same `question_id` twice with no existing response, both items pass the per-item `find_one` existence check (neither sees the other's pending insert) and each gets a fresh `_id` — two response documents for one question. Subsequent reads/updates keyed on `(question_id, user_id)` then behave unpredictably (which doc `find_one` returns is unspecified).

**Fix sketch**: dedup the payload before processing (last-write-wins within a batch, mirroring what two sequential saves would produce), or upsert on `(question_id, user_id)` instead of insert.

**Evidence**: `backend/app/db/questions.py` — per-item `find_one` + `insert_one` loop; no intra-batch dedup.

frankbria:
Closes #242.

## The issue's premise was half the story

`save_question_responses_batch` prepared **every** item's write before executing **any** of them, so two items with the same `question_id` each ran a `find_one` that saw no existing response, and each minted a fresh `ObjectId`.

What that actually does depends on the **unique** index `question_user_idx` on `(question_id, user_id)`, which production builds at startup (`main.py:71` → `ensure_question_indexes`). Verified empirically against real Mongo before writing any code:

| unique index | actual outcome |
|---|---|
| **absent** | 2 documents stored — the issue's report, verbatim |
| **present** (any correctly-started deployment) | 1 document, but the **first** answer wins: the user's **newer answer is silently discarded**, and a raw `E11000 duplicate key error ...` from the driver is handed back to the client with `success: false` |

So the bug production actually had is worse than the one reported — not a duplicate row, but **losing the user's newest writing and calling it a database error**. The index can still be absent, too: `ensure_question_indexes` wraps every index in one `try/except` and only logs, so a collection that already holds duplicates fails to build that index (and every one after it). The fix corrects both states because it collapses in the DAO rather than relying on the index to catch it.

## The fix

Repeated `question_id`s retarget the write already prepared for that question — last-write-wins, exactly as two sequential saves would leave it.

- **Dedup happens *after* per-item validation**, so each item keeps its own attributable error: `[{q1,"text"}, {q1,""}]` still fails index 1 with "Missing response_text" and still saves `"text"`, matching sequential behavior. Deduping the payload up front would misattribute the winner's error to the earlier item.
- **Both collapsed items report the single write's outcome** and its shared `response_id`. A client that debounces a double-send must not receive `success: false` for a benign payload. The existing `total == saved + failed` per-item invariant holds.

## Demo

`docs/demos/2026-07-14-issue-242-batch-duplicate-question-id.md` — **`showboat verify` exit 0**, every block reproducing byte-for-byte from a foreign cwd. Two real `uvicorn` servers (pristine main worktree :8801 vs branch :8802) on a local mongod, `BYPASS_AUTH=false`, genuine better-auth session cookie, book created through the real API. Same batch to both, then read back through the real read path the editor calls on reload:

- **main**: `success:false`, raw `E11000` leaked, reload serves the stale `draft` — *the user's completed answer is gone*.
- **branch**: `success:true` 2/2, reload serves the full newer answer with `status: completed`.
- **index dropped**: main stores **2** documents; branch stores **1** with the newer answer.

## Tests

5 new-behavior pins in `backend/tests/test_batch_question_responses.py`, all RED before the fix (the suite's fixture creates no indexes, so it had only ever exercised the no-index path — the production shape was untested; one new test builds the real index and pins it). **Mutation-verified**: stripping the dedup fails exactly 4; keeping the collapse but letting the *first* write win fails 3 — which is what pins the *winner*, not just the document count.

Backend **1121 passed / 12 skipped, 92% coverage**. Pre-commit gates green, no `--no-verify`.

## Reviews

opencode (GLM) pre-PR: **no Critical, no Major**. Four Minors — three adopted:
- **M2**: results came back out of order (validation failures append during prep, writes during execution); the dedup extended that to the all-success duplicate case (`[Q,R,Q]` → `0,2,1`). Now sorted by index before returning. Pre-existing and visible in the #187 demo's own captured output (`1,2,0`).
- **M1**: docstring overstated "matches two sequential saves" — true for the stored answer, not `edit_history`. Tightened.
- **M3**: pinned the untested repeat edges (3× repeat, invalid-then-valid).
- **M4** (test can't distinguish fixed from unfixed on the update path): correct for the removal mutation and disclosed as such — but the first-write-wins mutation proves it *does* pin which write survives. Kept, characterized honestly as a baseline.

## Known limitations

- **`edit_history` deliberately diverges**: two sequential saves leave one entry; a collapsed pair leaves none. The intermediate answer was never persisted, so recording it as an edit would be fiction. Documented on the function.
- **The residual `E11000` leak is untouched**: two *concurrent* requests can still race the index and surface the same raw driver text. The generic `f"Database error: {e}"` path is a pre-existing class (`plans/003-stop-leaking-exception-detail.md`), out of scope. This fix makes it unreachable from a duplicate batch, which is all #242 covers.


frankbria:
## Third-party review triage (opencode / GLM)

Per the repo's pre-PR + post-PR convention. opencode was available both rounds (no contention), so no codex fallback was needed.

### Pre-PR — no Critical, no Major; 4 Minors, 3 adopted

| # | Finding | Disposition |
|---|---|---|
| M2 | `results` came back out of order — validation failures append during prep, writes during execution. The dedup *extended* this to the all-success duplicate case (`[Q,R,Q]` → indexes `0,2,1`). | **Adopted** — sorted by index before returning. Pre-existing and visible in the #187 demo's own captured output (`1,2,0`), so this makes a latent contract wart deterministic rather than inheriting it. |
| M1 | Docstring claimed collapsing "matches two sequential saves" — true for the stored answer, false for `edit_history` (sequential leaves 1 entry, collapsed leaves 0). | **Adopted** — wording tightened to say the divergence is deliberate: the intermediate answer was never a persisted state. |
| M3 | No test for 3+ repeats, or for invalid-then-valid ordering. | **Adopted** — both pinned. The 3-repeat test also discriminates the M2 sort (pre-sort order is `[0,2,3,1]`). |
| M4 | `test_..._over_existing_response_updates_once` can't distinguish fixed from unfixed on the update path. | **Kept, characterized honestly.** Correct as stated for the *removal* mutation (two `update_one`s on one `_id` last-write-win anyway). But the *first-write-wins* mutation fails it — so it does pin **which** write survives, which the insert-path tests don't cover alone. Disclosed as a baseline rather than dropped. |

### Post-PR (fresh session, no memory of round 1) — clean

> "Verdict: Clean — no Critical or Major regressions. No manufactured findings."

It independently verified the things most likely to bite:
- **`results.sort()` KeyError**: checked all six `results.append` sites — including the easily-missed outer `except Exception` in the prep loop — every one carries `"index"`. Safe.
- **`op["index"]` → `op["indexes"]`**: grepped for leftover readers; none. `successful_ops` never escapes the function, and the endpoint reads only the top-level result dict.
- **Payload aliasing**: confirmed `payload = prior_op["update"]["$set"] | prior_op["insert"]` are reference assignments to the very dicts the execute loop writes — no aliasing gap.
- **Accounting**: each index lands in exactly one bucket (validation `continue`s before the dedup block; no index can appear in two ops' `indexes`), so `total == saved + failed` holds.
- **Non-duplicate path**: unchanged — `ops_by_question.get()` returns `None`, dedup skipped entirely; hoisting `word_count` is observationally identical.

It re-derived M4 independently and reached the same conclusion I'd disclosed, which is a good sign the characterization is honest rather than convenient.

PR #297, opened 2026-07-16
[P2.19] fix(questions): generate-then-swap bulk regeneration to avoid empty-chapter window (#234)

### LINKED ISSUE #234: [P2.19] Bulk question regeneration: generate-then-swap to avoid empty-chapter window on AI outage
Follow-up from #182 / PR #233 (raised by the opencode pre-PR review).

**Problem:** `regenerate_chapter_questions` (`backend/app/services/question_generation_service.py`) deletes existing questions **before** generating replacements. Since #182, a genuine OpenAI outage during bulk regenerate correctly surfaces as a structured 503 — but the chapter is left with zero questions until the user retries after AI recovery. (Pre-#182 the outcome was worse in the dimension that issue targeted — template questions silently returned as HTTP 200 — but the delete-first ordering itself predates #182.)

**Impact bounds:** user-authored responses are never lost (`preserve_responses=true` keeps answered questions; `false` is an explicit user discard). What's lost until retry is reproducible machine-generated questions.

**Proposed fix:** generate-then-swap — generate and persist the new batch first, then delete the old questions, excluding the new ids (needs an `exclude_ids` param on `delete_questions_for_chapter` or an id-scoped delete). With `preserve_responses=true` the would-be-deleted count must be computed up front (count of questions without responses) instead of derived from the delete's return value.

**Also fold in (N5 from the same review):** `_process_generated_questions` calls `_generate_fallback_questions` with hardcoded `chapter_title="Chapter"` (two call sites), so the PLOT template renders "What is the main event or conflict in **Chapter**?" instead of the real title. Thread the actual chapter title through.

frankbria:
## Summary

Fixes #234. Bulk question regeneration (`regenerate_chapter_questions`) deleted a chapter's existing questions **before** generating the replacements. Since #182 a genuine OpenAI outage during bulk regenerate correctly surfaces as a structured **503** — but the delete had already run, leaving the chapter with **zero questions** until the user retries after AI recovery.

The fix reorders to **generate-then-swap**: generate and persist the new batch **first**, then delete the old questions **excluding the just-created ids**. On an AI outage the exception propagates before any delete is reached, so the chapter keeps its questions.

Impact bound (unchanged): user-authored answers were never at risk — answered questions are preserved (`preserve_responses=true`) or explicitly discarded (`false`). What the old ordering lost on an outage was reproducible machine-generated questions.

## Changes

- **`regenerate_chapter_questions`** (`question_generation_service.py`): generate + persist first, then `delete_questions_for_chapter(..., exclude_ids=new_ids)`. The number of replacements to generate is now computed **up front** — via a new `count_questions_without_responses` helper on the preserve path, or `count` on the discard path — instead of being derived from a delete's return value (which no longer runs first).
- **`delete_questions_for_chapter`** (`db/questions.py`): new `exclude_ids` param — the freshly created questions (which have no responses yet) are skipped so a preserve-delete doesn't sweep them up.
- **`count_questions_without_responses`** (`db/questions.py`): new helper, scoped by `book_id`/`chapter_id`/`user_id`, counting questions with no saved response.
- **N5 from the #233 review**: threaded the real chapter title through `_process_generated_questions` so the PLOT fallback template renders the actual title (e.g. *"…conflict in The Midnight Heist?"*) instead of the hardcoded `"Chapter"`.

## Tests

- 5 new service pins (`TestRegenerate` + fallback-title) and 4 new real-Mongo DAO pins (`test_question_regeneration_swap.py`) — all **RED-verified at birth**.
- Mutation-verified: reintroducing delete-before-generate fails the outage/ordering pins; dropping `exclude_ids` fails the swap pins; reverting the title threading fails the fallback pin; dropping the DAO `exclude` skip fails the DAO pin.
- Full backend suite green.

## Review

- Pre-PR cross-family review (opencode / GLM): **"no Critical/Major"** — verified all four risk areas (empty-chapter window, `exclude_ids` on both preserve paths, `count_questions_without_responses` scoping, data-loss/double-delete).

## Demo

`docs/demos/2026-07-16-issue-234-generate-then-swap.md` (`showboat verify` clean) — pristine `main` worktree vs this branch, real Mongo, deterministic AI outage:
- **Outage:** main `questions_after=0` (empty-chapter window) vs branch `questions_after=4` (intact). Both raise the same `AI_UNAVAILABLE` — the only variable is the delete ordering.
- **Happy path (branch):** answered questions and their answers preserved, unanswered swapped.
- **N5 title:** main renders *"conflict in Chapter?"* vs branch *"conflict in The Midnight Heist?"*.

## Known limitations

- `preserved_count = count - deleted_count` can report a negative value when the requested `count` is smaller than the chapter's unanswered total. This formula and the metadata field are **unchanged from the previous code** (not introduced here) — flagged by the opencode review as pre-existing; left out of scope.
- The regenerate batch is still floored at 3 questions by `generate_questions_for_chapter` (pre-existing clamp), so replacing 1–2 unanswered questions yields 3 — visible in the happy-path demo. Not a regression.


frankbria:
### Cross-family pre-PR review (opencode / GLM) — no Critical/Major

Verified against the four flagged risks:

- **Empty-chapter window (AI outage)**: Fixed. `generate_questions_for_chapter` persists the new batch first; the delete is only reached on success. `AIServiceError` propagates before any deletion — covered by `test_ai_outage_leaves_existing_questions_intact`.
- **`exclude_ids` (preserve true/false)**: Correct. `delete_questions_for_chapter` checks `if question_id in exclude: continue` *before* both the response lookup and the `not preserve_with_responses` branch, so the fresh batch survives on both paths.
- **`count_questions_without_responses` scoping**: Correct. Filtered by `book_id`/`chapter_id`/`user_id`, matching `delete_questions_for_chapter`'s query; `str(_id)` used consistently with the response-lookup convention.
- **Data-loss / double-delete**: None. `new_ids` are the real persisted ids returned by `create_questions_batch`; new batch inserted once, old deleted once.

**Minor (pre-existing, not blocking, not introduced here):** `preserved_count = count - deleted_count` can go negative if the requested `count` is smaller than the unanswered total — unchanged from the previous code. Documented under Known Limitations.

`codex review` run as corroboration. Demo (`docs/demos/2026-07-16-issue-234-generate-then-swap.md`) reproduces the main-vs-branch outage differential; `showboat verify` exit 0.

PR #269, opened 2026-07-10
fix(ci): block fork PRs from the label-gated staging E2E job (#191)

frankbria:
Closes #191.

## Problem
`e2e-staging-tests.yml` runs on labeled `pull_request` (incl. `synchronize`) and executes PR-controlled code (`npm ci` lifecycle scripts, Playwright specs) inside the `staging` environment, which holds `TEST_USER_EMAIL`/`TEST_USER_PASSWORD` and Slack/Discord webhook secrets. A maintainer labeling a benign fork PR `e2e-staging`, followed by an attacker push (`synchronize` re-triggers with the label still attached), runs attacker code against live staging — the classic pwn-request shape.

## Fix (AC branch 2: same-repo head guard)
The `e2e-staging` job's `if:` now also requires `github.event.pull_request.head.repo.full_name == github.repository`. Fork PRs never run, labeled or not; same-repo labeled PRs, the 6-hourly schedule, and `workflow_dispatch` are unchanged. The guard fails closed (`head.repo` null → skip). `load-smoke` was already `workflow_dispatch`-only.

**Why not the other AC branches**: required reviewers on the `staging` environment (branch 1) would gate *every* run referencing the environment, including the 6-hourly cron — breaking scheduled coverage. `workflow_dispatch`-only (branch 3) removes the documented, in-use labeled-PR capability.

## Verification (act dry-run truth table + actionlint)
| Event | Before | After |
|---|---|---|
| fork PR + `e2e-staging` label (synchronize) | **runs all steps** | **skipped (0 steps)** |
| same-repo PR + label | runs | runs |
| fork PR, no label | skipped | skipped |
| schedule | runs | runs |

`actionlint` clean. Demo document to follow.

## Known Limitations
- This repo is **public**, so GitHub already withholds all secrets (including environment secrets) from fork `pull_request` runs — the headline credential exfiltration was not exploitable via this exact trigger today. The guard is defense-in-depth: it stops fork code from *executing* against live staging at all, and removes the one-drift-step-away hole (a `pull_request_target` swap, or the repo going private with fork-secrets enabled).
- Trust boundary after the fix: anyone with push access to this repo can still trigger staging runs via a same-repo branch + label — intended.
- Follow-up: `glm-review.yml` has the same latent class (`pull_request` + `ZHIPU_API_KEY`); filed separately.

## Reviews
[...]

Issue #191, opened 2026-07-04
[P1.11] e2e-staging workflow runs fork PR code with live staging secrets (pwn-request)


frankbria:
**Problem:** `e2e-staging-tests.yml` triggers on labeled `pull_request` (incl. `synchronize`) and runs attacker-controllable fork code (`npm ci` lifecycle scripts, Playwright specs) in the `staging` environment with `STAGING_TEST_EMAIL`/`PASSWORD` and Slack/Discord webhooks injected. A maintainer labeling a benign PR, then the attacker pushing malicious code (synchronize re-triggers), can exfiltrate staging creds.

**Evidence:** `.github/workflows/e2e-staging-tests.yml:31` (trigger/label gate), env injection at ~:77-78,:126-127.

**Acceptance criteria:**
- Require manual approval on the `staging` environment for PR-triggered runs, or restrict the label-gated job to same-repo head (`head.repo.full_name == github.repository`), or move to `workflow_dispatch` only for external contributions.

Issue #189, opened 2026-07-04
[P1.9] Backend binds 0.0.0.0 on a shared VPS — API reachable without TLS


frankbria:
**Problem:** PM2 starts uvicorn with `--host 0.0.0.0 --port 8000` (and every deploy script repeats it), on a shared VPS with other apps. The FastAPI backend (and Next server on 3002) are reachable directly at `http://<public-ip>:8000`, bypassing the nginx reverse proxy that terminates TLS and applies CORS/security headers — unless a host firewall explicitly drops those ports (none is in the repo to confirm).

**Evidence:** `ecosystem.config.template.js:9`; `scripts/deploy.sh:110`, `scripts/deploy-fixed.sh:191`.

**Acceptance criteria:**
- Backend binds `127.0.0.1` behind nginx (update the ecosystem template so all deploy paths inherit), OR a host firewall dropping external 8000/3002 is documented and verified.

PR #306, opened 2026-07-17
security(ci): block fork PRs from the ZHIPU_API_KEY review job (#270)

### LINKED ISSUE #270: [P2.21] glm-review.yml: pull_request trigger + ZHIPU_API_KEY is the same latent pwn-request class as #191
**Problem:** `.github/workflows/glm-review.yml` triggers on `pull_request` (`opened`, `synchronize`) and injects `ZHIPU_API_KEY` into a job that processes PR content. Found while fixing #191 (PR #269).

**Why it's latent, not live:** the repo is public, so GitHub withholds repo secrets from fork `pull_request` runs; and the job runs the pinned `glm-cr` action over the diff rather than executing fork npm scripts. But the config sits one drift step (trigger swapped to `pull_request_target`, repo going private with fork-secrets enabled) from leaking the key — the exact drift class #191 closed for the staging workflow.

**Acceptance criteria:**
- Add the same same-repo head guard used in #191 (`github.event.pull_request.head.repo.full_name == github.repository`) to the review job's `if:`, or otherwise ensure fork PRs can never reach a run with `ZHIPU_API_KEY` in env.
- Bonus hardening while touching the file: pin the `glm-cr` action to a commit SHA instead of the mutable `@v1` tag.

frankbria:
Closes #270.

## Problem
`.github/workflows/glm-review.yml` triggers on `pull_request` (`opened`, `synchronize`) and binds `ZHIPU_API_KEY` into the `review` job — the same latent pwn-request class #191 (PR #269) closed for the staging E2E workflow. The repo is public today so GitHub already withholds secrets from fork `pull_request` runs, but the config sits one drift step (repo going private with fork-secrets, or trigger swapped to `pull_request_target`) from leaking the key.

## Fix (config-only, one file)
Add the same same-repo head guard used in #191 to the job's `if:`:
```yaml
if: |
  github.event.pull_request.head.repo.full_name == github.repository &&
  (github.event.pull_request.changed_files >= 5 ||
   github.event.pull_request.additions >= 20 ||
   github.event.pull_request.deletions >= 20)
```
- The `&&` is AND'd around the **whole** size OR-group in explicit parens. In Actions expressions `&&` binds tighter than `||`, so without the parens a *large* fork PR (`additions >= 20`) would satisfy the OR and run — the parens are load-bearing.
- Fails closed for a deleted fork: `head.repo` is `null`, and `null == github.repository` is `false`.
- The job calls a reusable workflow via `uses:`; a job-level `if:` is evaluated **before** `secrets:` are bound, so a false guard means `ZHIPU_API_KEY` is never injected.

**Issue premise partially stale (verified):** the bonus AC (SHA-pin the action) was **already done** — the reusable workflow is pinned to `@b877d15a…` with an explanatory comment. No change needed there.

## Evidence (no workflow test framework — the #189/#191 precedent)
Because the job calls a *remote* reusable workflow, `act` can't run it directly, so the boolean guard is exercised in an isolated harness carrying only the guard.

**`act -n` truth table:**
| Case | Verdict |
|------|---------|
| fork PR + large diff | **SKIPPED — secret never binds** |
| deleted-fork (`head.repo` null) | **SKIPPED — fails closed** |
| same-repo + large diff | RUNS — legit review preserved |
| same-repo + tiny diff | SKIPPED — pre-existing size threshold |

**RED/GREEN mutation:** fork PR, large diff — old (size-only) guard **RUNS** (the vulnerability), new guard **SKIPPED** (fixed).

`actionlint` clean; YAML valid. Full evidence: `docs/demos/2026-07-17-issue-270-glm-review-fork-guard.md`.

## Review
opencode (GLM) pre-PR review: **clean to merge, no findings** — independently verified guard placement (secrets bind after job `if:`), null-fork fail-closed, operator precedence (parens load-bearing), and no bypass vector (trigger is `pull_request` only).

## Known limitations
- Defense-in-depth: on a public repo GitHub already withholds secrets from fork runs; the guard matters for the drift cases the comment calls out.
- This PR's own CI will show `GLM Review: skipping` (same-repo but no path-ignored/size change to the workflow itself is small) — expected, not a failure.


frankbria:
### Pre/post-PR cross-family review — clean

Two independent reviewers, both **clean, no findings**:

- **opencode (GLM), pre-PR** — verified guard placement (job `if:` gates before `secrets:` bind on a reusable-workflow call), null-fork fail-closed, operator precedence (parens load-bearing since `&&` binds tighter than `||`), no bypass vector (`pull_request`-only trigger).
- **codex, post-PR (independent)** — re-verified all three (precedence OK via parenthesized size predicate; `head.repo.full_name == github.repository` is false when `head.repo`/`full_name` is null → job skipped before binding `ZHIPU_API_KEY`; fork PRs can't satisfy the guard, same-repo PRs require write access; action SHA-pinned). Cited GitHub Actions expression operator/null-equality docs.

Issue #210, opened 2026-07-04
[P3.4] Delete debug scratch tests committed under backend/tests


frankbria:
**Problem:** Scratch scripts pollute the suite: `test_debug_loop.py` (681 lines, 176 `print()`s, 15 collected test fns run in CI), `test_services_summary.py` (0 tests, mutates `os.environ` on import — can poison collection order), and `run_test_*.py`/`simple_endpoint_test.py` (hardcoded tokens, a `d:\Projects\...` Windows path).

**Evidence:** `backend/tests/test_api/test_routes/test_debug_loop.py:11`; `backend/tests/test_services_summary.py:13`; `backend/tests/test_services/run_test_question_responses.py:20`, `run_test_api.py:12`.

**Acceptance criteria:** delete these files (or move genuine cases into properly-named asserting tests).

Issue #196, opened 2026-07-04
[P2.4] Advertised 'Session Management' security feature is dead code


frankbria:
**Problem:** `SessionMiddleware` only creates a session when it can read `request.state.user`, but nothing in the codebase ever sets `request.state.user` (auth is via a FastAPI dependency). So `request.state.session` is never populated: `/api/v1/sessions/current|refresh|logout` always 404, `/sessions/list` is always empty, and the advertised "30-min idle timeout / 12-hr absolute timeout / expiry warnings / suspicious-activity detection" (listed as Production Ready) never fire. The frontend `useSession` hook polls the dead endpoint and degrades to null.

**Evidence:** `backend/app/api/middleware/session_middleware.py:59-95` (no `request.state.user` setter anywhere), `backend/app/api/endpoints/sessions.py:37-41`, `frontend/src/hooks/useSession.ts:67,107`.

**Acceptance criteria:**
- Either wire session creation into the auth dependency so timeouts/detection actually work (+ an integration test asserting an authenticated request persists a session and populates `/sessions/current`), OR remove the `/sessions/*` router + middleware + frontend polling and stop advertising the feature.

Issue #186, opened 2026-07-04
[P1.6] Unauthenticated user-creation endpoint enables enumeration + junk records


frankbria:
**Problem:** `POST /api/v1/users/` has no auth dependency and returns distinct 409s for existing auth_id vs email vs 201 on success — an enumeration oracle. It also lets anyone insert arbitrary user records, which can collide on the email index and 500 a victim's real first-login. It is redundant with the session auto-create flow.

**Evidence:** `backend/app/api/endpoints/users.py:255-298` (no auth dep, distinct 409 messages), `backend/app/db/user.py:33-37` (`insert_one`).

**Acceptance criteria:**
- Endpoint requires auth/authorization or is removed; if kept, uniform response that doesn't distinguish exists/created, and rate-limited.

PR #320, opened 2026-07-20
[P3.8] Remove dead transcription + AI cache stacks; fix N+1 loops + list projection (#214)

### LINKED ISSUE #214: [P3.8] Remove or wire dead code: transcription stack + AI cache; fold in N+1 perf cleanups
**Problem:** The `/transcribe` router + AWS service (~680 LOC) is unmounted and falls back to hardcoded mock transcripts; `ai_cache_service` is permanently inert because `redis` is neither declared nor installed (so the "reduce API costs" cache AND the "return cached content on AI failure" resilience never work). Also N+1 query loops in question read/progress/rating/delete paths, and the dashboard list returns full inline chapter content.

**Evidence:** `backend/app/api/endpoints/transcription.py` (unmounted), `backend/app/services/ai_cache_service.py:17-22` (`redis` import guard always False); `backend/app/db/questions.py:243,427,464,551,664`; `backend/app/db/book.py:63-67` (no projection).

**Acceptance criteria:** remove the dead transcription + cache modules (or genuinely wire them: add/provision `redis`, mount the router with a real 503 instead of mock text); replace per-question loops with `$in`/bulk queries; project out `chapters.content` on the list endpoint.

frankbria:
Closes #214.

## Decision: remove, not wire
The AC offered "remove the dead modules **or** genuinely wire them (add/provision redis...)". Both dead subsystems are **removed** — the AC's primary branch and the project's consistent dead-code precedent (#120, #186, #196, #200, #210):

- **Transcription `/transcribe` stack** — the router was never mounted (`router.py` has no transcription include) and the service returned hardcoded **mock transcripts**. Zero production importers. Wiring would resurrect an AWS-Transcribe path nothing uses.
- **`ai_cache_service`** — `redis` is absent from `pyproject.toml`/`uv.lock`, so the import guard is permanently `False` and every cache op is a no-op. Wiring means provisioning redis on the shared VPS for a cache with **zero real dependents**. Removal is **behavior-preserving** (the cache always missed → always generated fresh; the error-fallback lookup always returned `None`), and git preserves history if redis is ever provisioned.

## Changes
1. **Delete transcription stack**: `endpoints/transcription.py`, `services/transcription_service.py`, `services/transcription_service_aws.py`, `schemas/transcription.py` + 4 test files. The **live** `enhance-transcription` voice-cleanup feature (`transcription_enhancement.py`, wired into `books.py`) is untouched.
2. **Delete `ai_cache_service.py`**; strip its 6 call sites from `ai_service.py` (`generate_clarifying_questions`, `generate_toc_from_summary_and_responses`) down to generate-only, preserving the `cached_content_available = False` + re-raise error contract. Remove the now-unreachable `AICacheError` and the dead `REDIS_URL` / `AI_CACHE_TTL` / `AI_CACHE_ENABLED` config + `.env.example` block (`AI_MAX_RETRIES` kept — used by retry logic).
3. **N+1 fix** (`db/questions.py`): `get_questions_for_chapter`, `get_ratings_for_chapter`, `get_chapter_question_progress` replace their per-question `find_one` loops with one `$in` batched lookup + dict map. (The issue also cited lines 551/664 — verified already batched via `$in`/`distinct`, so untouched.)
4. **List projection** (`db/book.py::get_books_by_user`): pure-exclusion projection dropping `table_of_contents.chapters.content` + `.subchapters.content`. The dashboard list returns `BookResponse` (exposes `toc_items`, never `table_of_contents`), so those chapter-draft HTML blobs were pure overfetch; titles/structure/`toc_items` untouched.

## Tests
- New `tests/test_db/test_n1_and_projection.py`: batched-join correctness for all 3 functions, a **`find_one`-must-not-be-called batching guard**, and the content-projection pin. Both AC-critical pins **mutation-verified RED** (revert projection → projection test fails; revert to per-question `find_one` → batching guard fails).
- AI error/retry tests (`test_ai_service_errors.py`) retargeted off the removed mock-cache fixture to `AIService()`; obsolete cache-behavior tests removed; the #188 single-retry-layer tests preserved. `AICacheError` tests removed with the class.
- Local suite (excl. the two network-bound AI files CI keys with a real OpenAI key): **1064 passed / 11 skipped**. Net **−2503/+626** lines.

## Reviews
- **Pre-PR cross-family**: opencode (GLM) hung with no output (the documented offline condition) → **`codex` fallback: clean** — confirmed cache-strip behavioral equivalence on the no-cache path, batched-join string-`question_id` matching + tallies, legal pure-exclusion projection compatible with `BookResponse`, and no dangling imports to deleted symbols.

## Known limitations
- A historical doc snapshot (`tests/TEST_COVERAGE_SUMMARY.md`) still lists a now-deleted transcription test as a "run this" example — left as-is per the #120 historical-snapshot precedent (docs-sync can refresh separately).
- Pre-existing lint (unused `hashlib`/`AIResponseParsingError` imports, a couple `E722` bare-excepts in `ai_service.py`/`questions.py`) is left untouched (predates this PR, out of scope; new file is ruff-clean).


frankbria:
**Pre-PR cross-family review** — opencode (GLM) hung with no output (the documented offline condition), so **`codex review` was the fallback** (posted here on-record):

> No newly introduced backend correctness, security, data-integrity, or maintainability defects were found. Cache removal is behaviorally equivalent for the no-cache path, batched Mongo joins preserve string `question_id` matching and tallies, the projection is pure exclusion and compatible with `BookResponse`, and source imports no longer reference deleted modules/symbols.

frankbria:
**Demo (hard gate)** — `docs/demos/2026-07-19-issue-214-remove-dead-code-perf.md`: a main-vs-branch outcome differential over real code + local Mongo. Each AC has before/after outcome evidence:

| Criterion | Outcome evidence (main → branch) |
|---|---|
| Transcription stack removed (no mock fallback) | `transcribe_audio()` returns mock `'Short audio sample.'` → module ABSENT (path gone) |
| AI cache removed | `AICacheService` present + `AIService.cache_service=True` → module ABSENT + attr `False` (AI flows still run) |
| N+1 → `$in` batched | `get_chapter_question_progress` over 5 questions: **5** `find_one` calls → **0** (one batched `find`), same result |
| List projection | list row **10,340 bytes** with chapter content → **296 bytes**, content dropped, `toc_items`/titles intact |

Removal is behavior-preserving (results unchanged).

frankbria:
Addressed the GLM precision review's single minor (`run_unit_tests.py:24-25` referencing the two deleted transcription tests): **removed the script** rather than patching its hardcoded list — it's not CI-invoked and duplicates `uv run pytest tests/` (the documented runner), so deleting it is the root-cause fix (no stale list to drift again; the #210 scratch-runner precedent). Thanks — that was a real dev-script regression from the deletions. All 'verified clean' items (N+1 string-key consistency, pure-exclusion projection vs `BookResponse`, cache-strip equivalence, zero live importers) match my analysis.

PR #323, opened 2026-07-21
fix(auth): preserve ?redirect deep-link through 2FA verification (#237)

frankbria:
Closes #237.

## Problem
A 2FA-enabled user who signs in with `?redirect=<path>` was always dropped on `/dashboard` after verification. The sign-in page returns early on `twoFactorRedirect` (handing off to the `twoFactorClient` navigation), and `verify-2fa/page.tsx` hardcoded `router.push('/dashboard')` — so the deep-link target was lost.

## Fix
- **sign-in page** stashes the already-sanitized redirect to `sessionStorage` (`auth:postVerifyRedirect`) before the 2FA hop. It **always** writes the current sanitized target (defaulting to `/dashboard`), so a fresh attempt overwrites any stale value left by an abandoned earlier 2FA sign-in in the same tab.
- **verify-2fa page** reads and clears the stash after successful verification, re-running it through `sanitizeRedirectPath` at the point of use (defense in depth), falling back to `/dashboard`. The stash survives a *failed* attempt so a retry still honors it.

Reuses the existing `sanitizeRedirectPath` (#184) at the consumption point, as the issue required.

## Review
Cross-family **codex** review (opencode occupied by a concurrent session — the documented hang condition) flagged one **Major**: the original fix skipped writing on the `/dashboard` target, leaving a stale stash consumable by a later no-redirect sign-in. Fixed by always writing the current sanitized redirect; the prior "skip-default" test was replaced with a stale-stash-overwrite test (mutation-verified RED against the skip-guard).

## Tests
- `SignInPage.test.tsx`: stash-on-2FA for a deep-link; stale-stash overwrite on a no-redirect fresh sign-in.
- `verify-2fa/__tests__/page.test.tsx` (new): honors stashed deep-link + clears it; falls back to `/dashboard`; re-sanitizes a tampered value; keeps the stash on a failed attempt.
- 15/15 pass; both behavior pins mutation-verified RED against reverted code.

## Known Limitations
None — frontend-only, self-contained; no backend or API surface touched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #237, opened 2026-07-07
[P3.11] 2FA sign-in drops the ?redirect deep-link (always lands on /dashboard)


frankbria:
Found by the post-PR review on #236 (pre-existing behavior, unchanged by that PR).

When a 2FA-enabled user signs in with `?redirect=<path>`, the sign-in page returns early on `twoFactorRedirect` (`frontend/src/app/auth/sign-in/page.tsx`) and `verify-2fa/page.tsx` hardcodes `/dashboard` after verification — so the deep-link target is lost.

**Possible fix**: thread the (sanitized via `sanitizeRedirectPath`) redirect through the verify-2fa page (query param or sessionStorage) and push it after successful verification. Must reuse `sanitizeRedirectPath` at the consumption point.

Low priority — product decision whether post-2FA deep-links matter.

Issue #184, opened 2026-07-04
[P1.4] Open redirect on sign-in via unvalidated ?redirect param


frankbria:
**Problem:** The sign-in page reads `?redirect` and `router.push`es it verbatim after auth. `router.push` navigates to absolute/protocol-relative external URLs, so `/auth/sign-in?redirect=https://evil.com` (or `//evil.com`) sends the freshly-authenticated user to an attacker site — classic phishing open redirect. The existing `sanitizeUrl` helper isn't used and permits external targets anyway.

**Evidence:** `frontend/src/app/auth/sign-in/page.tsx:83` (`searchParams.get("redirect")`), `:114` (`router.push(redirect)`).

**Acceptance criteria:**
- Only same-origin relative paths accepted: reject values not starting with a single `/` (and not `//` or containing `://`), fall back to `/dashboard`.
- Unit test covers `//evil.com`, `https://evil.com`, `javascript:` inputs.

Issue #215, opened 2026-07-04
[P3.10] Beta UX polish bundle (theme/validation/error-copy/support)


frankbria:
**Problem:** A cluster of small, one-session polish gaps found in the UX pass, none individually blocking but collectively rough for a paid beta:
- Profile page has a **duplicate Theme select that has no effect** (only local form state; `onSubmit` never calls `setTheme`) — remove it (Settings owns theme) or wire to next-themes. `frontend/src/app/profile/page.tsx:250-267`.
- **Save Settings button silently disabled by a validation error on a tab the user isn't viewing**, with no hint on the current tab. `frontend/src/app/dashboard/settings/page.tsx:69,179-184`.
- **Active Sessions list captures `ipAddress` but never renders it** — omitting the one datum most useful for spotting a rogue session. `frontend/src/components/settings/ActiveSessionsList.tsx:14,130-156`.
- **Avatar upload surfaces `[object Object]`** for FastAPI 422 array-`detail` responses. `frontend/src/components/profile/ProfilePictureUpload.tsx:50-61`.
- **Sign-up/sign-in show raw backend `error.message`** for unmatched errors to a brand-new user. `frontend/src/app/auth/sign-up/page.tsx:79,144`.
- **Book-overview load error → `window.location.reload()`** instead of an in-place refetch. `frontend/src/app/dashboard/books/[bookId]/page.tsx:144-146,373-378`.
- **Help-page support email is plain text, not a `mailto:` link.** `frontend/src/app/dashboard/help/page.tsx:41`.
- **No cancel affordance on long (10-60s) TOC/draft AI ops.** `frontend/src/components/toc/TocGenerationWizard.tsx:63-159`.
- **Genre lists offer fiction** despite the landing page's "nonfiction books" positioning. `frontend/src/app/page.tsx:56` vs `BookCreationWizard.tsx:40-50`.

**Acceptance criteria:** address the above (a single cleanup pass); each is a small isolated fix. Split out any that grows beyond a quick change.

PR #324, opened 2026-07-22
[P3.12] fix(auth): ProtectedRoute preserves ?redirect deep-link on client session expiry (#239)

### LINKED ISSUE #239: [P3.12] ProtectedRoute client guard drops the return deep link on session expiry
Found during #185 / PR #238 post-PR review (opencode N4). Pre-existing, affects dashboard and profile equally.

`frontend/src/components/auth/ProtectedRoute.tsx` pushes `/auth/sign-in` with no `?redirect=` param, so a client-side session expiry (middleware not involved — no full navigation) drops the user at sign-in without a deep link back to where they were. The middleware layer sets `?redirect=` only on full navigations.

**Suggested fix:** `router.push(`/auth/sign-in?redirect=${encodeURIComponent(pathname)}`)` using `usePathname()`; the sign-in page already validates it via `sanitizeRedirectPath` (#184). Note #237 tracks the related 2FA-path redirect drop.

frankbria:
Closes #239.

## Problem
The client-side `ProtectedRoute` guard pushed a **bare** `/auth/sign-in` on session expiry. Unlike a full navigation (where `middleware.ts` sets `?redirect=<pathname>`), a client-side expiry involves no navigation, so the middleware layer never runs — the user was dropped at sign-in with no deep-link back to where they were, and after re-auth always landed on `/dashboard`.

## Fix (single line, mirrors the middleware)
`ProtectedRoute.tsx` now reads `usePathname()` and pushes `/auth/sign-in?redirect=${encodeURIComponent(pathname)}` — exactly the deep-link the middleware already sets via `searchParams.set('redirect', pathname)`. The sign-in page already validates the param through `sanitizeRedirectPath` (#184, same-origin-only, decode-checked, `/dashboard` fallback) before honoring it, so no new trust surface is introduced.

## Outcome evidence (differential)
The behavior difference is the **navigation target**, pinned at the outcome level:
- **main**: `router.push('/auth/sign-in')` → RED against the new pins
- **branch**: `router.push('/auth/sign-in?redirect=%2Fdashboard%2Fbooks%2Fabc123')` → GREEN

Both new `ProtectedRoute` pins were **mutation-verified RED** against the bare-push code, then GREEN after the fix. `usePathname()` is a hook called unconditionally, so the four sibling suites that render the guard (`AuthPersistence`, `ProfilePage`, `ProfilePageDelete`, `ProfilePageSave`) were updated to mock it and assert the deep-link contract. Full frontend suite: **125 suites, 2209 passed / 5 skipped**.

## Review
Cross-family **codex** review: "clean to merge", no Critical/Major issues. (opencode hung — the documented concurrent-session condition — so codex was the fallback per the review invariants.)

## Known limitations
- `usePathname()` returns the path only, not the query string — a user at `/dashboard/books?tab=export` returns to `/dashboard/books` without the tab. This matches the existing middleware behavior (it also uses `pathname` only), so the two layers stay consistent; preserving search params would be a separate cross-layer change.
- The related 2FA-path redirect drop is tracked separately by #237 (already shipped in #237/PR #323) — out of scope here.

### DIFF
--- a/frontend/src/__tests__/AuthPersistence.test.tsx
+++ b/frontend/src/__tests__/AuthPersistence.test.tsx
@@ -6,6 +6,7 @@ import { useRouter } from 'next/navigation';
 // Mock Next.js navigation
 jest.mock('next/navigation', () => ({
   useRouter: jest.fn(),
+  usePathname: jest.fn(() => '/dashboard'),
 }));
 
 // Mock better-auth's useSession hook
@@ -44,9 +45,9 @@ describe('Authentication State Persistence', () => {
       </ProtectedRoute>
     );
 
-    // Expect router.push to be called with the sign-in route (better-auth uses /auth/sign-in)
+    // Expect router.push to the sign-in route with a ?redirect deep-link (#239)
     await waitFor(() => {
-      expect(mockRouter.push).toHaveBeenCalledWith('/auth/sign-in');
+      expect(mockRouter.push).toHaveBeenCalledWith('/auth/sign-in?redirect=%2Fdashboard');
     });
 
     // Should not render protected content
--- a/frontend/src/__tests__/ProfilePage.test.tsx
+++ b/frontend/src/__tests__/ProfilePage.test.tsx
@@ -10,6 +10,7 @@ import { toast } from '../lib/toast';
 // Mock Next.js hooks
 jest.mock('next/navigation', () => ({
   useRouter: jest.fn(),
+  usePathname: jest.fn(() => '/profile'),
 }));
 
 // Mock better-auth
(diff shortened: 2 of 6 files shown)


frankbria:
**Pre-PR cross-family review** (per review invariants): `codex` reviewed HEAD vs main → **"clean to merge"**, no Critical/Major correctness or security issues. opencode was attempted first but hung (the documented concurrent-session condition), so codex was the fallback.

frankbria:
## CI + review summary

**Pass:** Backend Tests, GLM review (**✅ no defects found**), GitGuardian, Snyk. E2E Staging correctly skipped (unlabeled). Frontend Tests: **125 suites / 2209 tests pass**.

**The one red check** is the global **functions** coverage gate: `84.56% vs 85%`. This is the **pre-existing repo-wide baseline** — identical to `main` (documented as red on main and merged past via admin override by both #215 on 2026-07-20 and #237 on 2026-07-21). This PR's src diff adds **no new function definitions** (only a `usePathname()` hook call + one dep-array entry inside the already-covered `ProtectedRoute`), so it cannot move the functions percentage — and the run landed on the exact `84.56%` main baseline. **Not a regression from this PR.** The coverage baseline is a separate, tracked repo-wide concern.

**Reviews:** pre-PR `codex` "clean to merge" + post-PR GLM CI review "no defects found" — both independently confirmed no open-redirect surface (the value is re-validated by `sanitizeRedirectPath`), middleware parity, and safe null-path fallback.

Merging via admin override past the pre-existing coverage baseline, consistent with the #215/#237 precedent.

Issue #203, opened 2026-07-04
[P2.11] TOC clarifying-questions shows 'Auto-saved' but never persists — answers lost on navigation


frankbria:
**Problem:** The debounced auto-save effect in `ClarifyingQuestions` sets `lastSaved(new Date())` — flipping the UI to a green "Auto-saved" checkmark — but never calls any persistence API (the code even comments "Skipping save call due to type mismatch"). Responses are only submitted at the final "Generate Table of Contents" step. `bookClient` exposes only a GET for question responses, no save/PUT. So a beta user who refreshes, navigates away, or crashes before submitting loses all typed answers — while the UI explicitly claimed they were saved.

**Evidence:** `frontend/src/components/toc/ClarifyingQuestions.tsx:51-82` (effect sets `lastSaved` at :68, no save call at :65-66), `:137-149` (renders "Auto-saved"); `frontend/src/lib/api/bookClient.ts:943` (only `getQuestionResponses`).

**Acceptance criteria:**
- Either remove the fabricated "Auto-saved" indicator and clearly state answers persist only on submit, OR wire the debounced effect to a real save endpoint (mirror the summary page's `saveBookSummary`).
- A test asserts the indicator only shows after a real successful persist.

PR #325, opened 2026-07-22
[P3.13] fix(api): honest question-response completion status + drop dead toc-readiness composite (#276)

### LINKED ISSUE #276: [P3.13] PUT /books/{id}/question-responses stores partial answer sets as status 'completed'
**Context**: Filed from the #203 pre-PR review (codex flagged it twice; verified out of scope there).

**Problem**: `PUT /books/{book_id}/question-responses` (`books.py:994`) unconditionally stamps whatever list it receives with `status: "completed"` and returns `ready_for_toc_generation: true` — even for a partial answer set. Now that #203 wires the TOC wizard's debounced auto-save to this endpoint, partial sets are routinely persisted mid-typing, so the stored status no longer means "the user finished answering".

**Why it's currently harmless (verified in #203)**:
- The `toc-readiness` endpoint computes a composite `is_ready_for_toc` from `responses_status == "completed"` but **discards it** — both return branches return only summary-analysis data.
- The TOC wizard always regenerates questions on mount and sends responses in the `generate-toc` request body (#105), which wins over the persisted fallback.

So no live consumer misbehaves today — but the stored semantics are a trap for any future consumer of `question_responses.status` or the `generate-toc` persisted-responses fallback.

**Acceptance criteria**:
- The endpoint distinguishes partial from complete sets (e.g. accept a `status` of `draft`/`completed` from the client, or derive it server-side by comparing against `clarifying_questions.questions`), OR the dead composite readiness computation and the misleading `status` field are removed outright.
- `toc-readiness`'s dead composite computation is either wired or deleted (no computed-then-discarded code).
- Tests pin whichever semantics are chosen.

frankbria:
Closes #276.

## Problem
Two defects in the pre-TOC endpoints of `books.py`, both filed from the #203 review:

1. **`PUT /books/{id}/question-responses` hardcoded `status: "completed"`** (and returned `ready_for_toc_generation: true`) for *any* list it received. Now that #203 wires the TOC wizard's debounced auto-save to this endpoint, partial answer sets are routinely persisted mid-typing — so the stored status no longer meant "the user finished answering." A trap for any future consumer of `question_responses.status` or the `generate-toc` persisted-responses fallback.
2. **`GET /books/{id}/toc-readiness` computed a composite `is_ready_for_toc` + `next_steps`** from questions/responses that **both return branches discarded** (they return only summary-analysis fields). Computed-then-discarded dead code.

Verified harmless today (no live consumer reads `status`, `ready_for_toc_generation`, the composite, or `next_steps` — checked frontend + backend), but the stored semantics were a latent trap.

## Fix
- **Derive completion status server-side** (the AC's suggested approach): a set is `completed` only when **every** clarifying question is covered by a response — matched **by question text, not a count**, so duplicate/stale responses can't reach the count while a current question stays unanswered. Answers are already validated non-empty. With no question set to compare against (unknowable completeness), we fail toward `draft` — **never a false `completed`**. `ready_for_toc_generation` mirrors it; `clarifying_questions.status` (`answered`/`pending`) tracks it too.
- **Delete the discarded toc-readiness composite** and its now-dead locals (net −49 lines in the handler). Readiness now reads purely from summary + AI analysis — the behavior both return branches already had.

## Tests (`test_books_pretoc_coverage.py`)
New `TestQuestionResponseCompletionStatus`: partial set → `draft`/`False`, complete set → `completed`/`True`, **duplicate responses don't fake completion** → `draft`, no-questions → `draft`. Plus a toc-readiness pin that an analysis-ready book with zero questions/responses still reads ready (proving the composite is gone) and that `next_steps` is absent from the response. Updated the one existing test that pinned the buggy `completed`.

- The two bug pins (`test_partial_set_is_draft`, `test_no_clarifying_questions_stays_draft`) are **RED against the old hardcode**.
- The duplicate-coverage pin is **RED against a count-based (`>=`) check** and GREEN with coverage (mutation-verified).
- Full pretoc suite **45 passed**; related toc/batch suites **46 passed**.

## Reviews
Pre-PR cross-family review via **codex** (opencode was occupied by a concurrent foreign-cwd session — the documented [[opencode-concurrent-hang]] condition). Codex raised one **Major**: a `len(responses) >= question_count` check (my first draft) could be fooled by duplicate/stale responses. Adopted and hardened to coverage-by-question-text (codex's own `==` suggestion had the same flaw), with a dedicated mutation-verified pin.

## Known limitations
- Completion matching is by exact question text — the established `{question, answer}` contract the wizard and `generate-toc` already use. Pre-existing import-block `ruff F401` warnings in `books.py` predate this PR and are left untouched (out of scope; this diff removes one dead local, net-reducing lint debt).


frankbria:
### Cross-family review (codex — opencode occupied by a concurrent foreign-cwd session, the documented hang condition)

**Pre-PR (RED→fixed):** codex raised one **Major** — my first-draft `len(responses) >= question_count` could be fooled by duplicate/stale responses (2 responses both answering Q1, omitting Q2 → count met, Q2 unanswered → false `completed`). Adopted; hardened to **coverage-by-question-text** (codex's own `==` suggestion had the same flaw). Added a dedicated pin (`test_duplicate_responses_do_not_fake_completion`), **mutation-verified RED against the count-based check**.

**Post-PR (final diff):** codex → *"Sound for Critical/Major correctness/security. No blocking defects found. The completion status now fails closed when there is no generated question set, uses coverage by current question text rather than response count, and the removed `toc-readiness` composite was not part of the returned API contract."*

Issue #56, opened 2025-12-18
Enhance voice input with AI analysis and formatting


frankbria:
## Problem
Voice input works for basic speech-to-text but the AI analysis is basic. The user manual promises intelligent voice analysis that should improve transcription quality.

## Current Behavior
- Voice-to-text transcription works
- Raw transcription appears in editor
- No AI enhancement of spoken content
- No automatic formatting or cleanup

## Expected Behavior
- Voice input transcribes speech accurately
- AI analyzes transcription for:
  - Paragraph breaks at natural pauses
  - Punctuation correction
  - Grammar improvements
  - Filler word removal ("um", "uh", "like")
- Users can choose between raw and enhanced transcription

## Implementation Guidance
1. Current voice input implementation in `frontend/src/components/chapters/ChapterEditor.tsx`
2. Add post-processing step after transcription
3. Send transcription to AI service for enhancement
4. Implement enhancement options (remove fillers, add paragraphs, fix grammar)
5. Show side-by-side comparison of raw vs. enhanced
6. Let user choose which version to keep

## Files Likely Involved
- `frontend/src/components/chapters/VoiceInput.tsx` - Voice input component (if separate)
- `frontend/src/components/chapters/ChapterEditor.tsx` - Editor integration
- `backend/app/services/ai_service.py` - Add transcription enhancement
- `backend/app/api/routes/transcription.py` - Create endpoint

## Acceptance Criteria
- [ ] Voice transcription works reliably
- [ ] AI enhancement removes filler words
- [ ] Paragraphs break at natural pauses
- [ ] Grammar and punctuation improved
- [ ] Users can toggle raw vs. enhanced
- [ ] Enhancement processing shows loading indicator
- [ ] Side-by-side preview before accepting
- [ ] E2E test verifies voice-to-enhanced workflow

## Blocked by
Issue #41 (Implement robust error handling for AI service failures)

PR #354, opened 2026-07-23
[#330] fix(voice): continuous dictation appends every segment (stale-closure data loss)

### LINKED ISSUE #330: [P0.1] Continuous voice dictation silently loses all but the last phrase (stale closure)
**Phase:** P0.1 — launch blocker (data loss in a headline feature). No dependencies.

**Problem:** `VoiceTextInput.initializeSpeechRecognition` builds the `onresult` handler once when recording starts, capturing the `value` prop in the closure. With `continuous = true`, each finalized segment runs `onChange(value + finalTranscript + ' ')` against the SAME captured `value`, so every new segment overwrites the previous one instead of appending.

**Evidence:**
- `frontend/src/components/chapters/VoiceTextInput.tsx:84` (useCallback deps `[isSupported, value, onChange]`), `:124`/`:134` (`onChange(value + ...)`)
- The summary page does it correctly with a functional updater: `frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:100` (`setSummary(prev => ...)`)
- The test mock fires only ONE final result (`frontend/src/__tests__/mocks/speechRecognition.ts:39`), so the accumulation path is never exercised — the bug ships green.

**Impact:** Dictating a multi-sentence interview answer keeps only the final sentence; user-authored content is silently destroyed in an advertised feature.

**Acceptance criteria:**
- [ ] Append via a functional updater / ref instead of the captured `value`.
- [ ] Add a frontend test that fires two separate final results and asserts BOTH survive.
- [ ] Extend the mock to emit interim + multiple final results so the path is covered.

frankbria:
Closes #330

## Problem
`VoiceTextInput.initializeSpeechRecognition` builds the `onresult` handler once when recording starts, capturing the `value` prop in its closure. With `continuous = true`, every finalized segment computed `onChange(value + segment + ' ')` against the **same captured `value`**, so each new segment **overwrote** the previous one. Dictating a multi-sentence answer kept only the last sentence — silent loss of user-authored content in an advertised feature.

The existing test/mock fired exactly one final result, so the accumulation path was never exercised and the bug shipped green.

## Fix
Mirror the latest `value` into a `valueRef` synced by an effect; the long-lived `onresult` handler reads `valueRef.current` (always fresh) instead of the frozen closure variable, so successive segments append. Removed the now-unused `value` from the `useCallback` deps. This mirrors how the summary page already accumulates correctly (`setSummary(prev => …)`) — but adapted to a `(value: string) => void` callback prop that has no functional form.

Both `onresult` write paths were fixed (cursor-insert branch + append branch); dictation actually hits the append branch because the textarea isn't mounted in voice mode.

## Tests (mutation-verified RED)
- Extended `speechRecognition.ts` mock: a `sequence` config emitting **interim + multiple final** results on successive timers, plus a deterministic `emitResult` for driving segments inside `act()`.
- New `Continuous dictation (accumulation)` suite:
  - fires **two separate final results** via a controlled wrapper (state fed back through `onChange`) and asserts **BOTH survive**;
  - a `sequence`-driven test covering interim + two finals.
- Reverting the fix (`valueRef.current` → captured `value`) turns both accumulation tests **RED** (only the last segment survives); restoring turns them GREEN.

## Acceptance criteria
- [x] Append via a functional updater / ref instead of the captured `value`
- [x] Frontend test fires two separate final results and asserts BOTH survive
- [x] Mock extended to emit interim + multiple final results so the path is covered

## Verification
- `VoiceTextInput.test.tsx`: 8/8 pass (2 new accumulation tests mutation-verified)
- `VoiceTextInputIntegration.test.tsx` (QuestionDisplay consumer): 5/5 pass
- typecheck clean; eslint 0 errors (1 pre-existing `any` warning on untouched error-event code)
- Cross-family **codex review: no introduced bugs**

## Known limitations
None. Browser `SpeechRecognition` is native/non-deterministic and unavailable headless, so real-voice E2E isn't feasible (project precedent, #56/#57); the mutation-verified accumulation test is the outcome evidence.


frankbria:
## Demo — outcome evidence (bug vs fix)

Browser `SpeechRecognition` is unavailable headless, so the honest outcome evidence is a bug-vs-fix differential on the exact acceptance scenario: dictate **two separate finalized segments** through the real `VoiceTextInput` (controlled wrapper feeding `onChange` back as `value`, the way `QuestionDisplay` does).

**Fixed branch** — both survive:
```
✓ appends every finalized segment instead of overwriting the previous one (#330)
  accumulated value = "First sentence. Second sentence. "
```

**Same scenario, stale closure reintroduced (`valueRef.current` → captured `value`)** — first sentence lost:
```
✕ appends every finalized segment...
  Expected text content: "First sentence. Second sentence."
  Received (buggy):      "Second sentence. "   ← the bug: only the last segment survives
```

Acceptance criteria mapped to evidence:
- **Append via ref, not captured value** → fixed run yields both sentences; mutant loses the first.
- **Test fires two separate finals, BOTH survive** → the passing accumulation test above.
- **Mock emits interim + multiple finals** → extended `speechRecognition.ts` `sequence` config + the `interim + two finals` test (green).

PR #358, opened 2026-07-24
[P1.2] Error tracking (Sentry) + 429 logging + real perf sink (#334)

### LINKED ISSUE #334: [P1.2] No error tracking in either app — 500s/429s die in PM2 stdout with no alerting
**Phase:** P1.2 — pre-launch. No dependencies.

**Problem:** No Sentry/Datadog/OTel/Rollbar dependency in `backend/pyproject.toml` or `frontend/package.json`. The global backend handler and frontend `error.tsx` only `logger.error`/`console.error` (`error.tsx` literally comments `// Log the error to an error reporting service` then just console.errors). Rate-limit/quota 429s emit no log line; the frontend perf-telemetry production branch only writes localStorage.

**Evidence:** `backend/app/main.py:161,15`; `frontend/src/app/error.tsx:17`; `backend/app/api/dependencies.py:117,180` (429 no log); `frontend/src/lib/performance/metrics.ts:120`.

**Impact:** A systemic 500 wave, a 429/abuse storm, or a real-user perf regression is invisible until a support email arrives; diagnosis means SSHing to grep raw PM2 logs.

**Acceptance criteria:**
- [ ] Add Sentry SDK to both apps (`sentry-sdk[fastapi]` + `@sentry/nextjs`).
- [ ] `logger.warning` before each 429 raise with subject/path/limit.
- [ ] Point the perf-metrics production branch at a real sink or drop the localStorage-only path.

frankbria:
Closes #334.

## Problem
500s, per-user 429s (rate limit / AI quota), and real-user perf regressions all died in PM2 stdout with no alerting — diagnosis meant SSHing to grep raw logs. No error-tracking dependency existed in either app.

## Changes
**AC1 — Sentry SDK in both apps (DSN-gated: inert unless a DSN is set, so CI/local/tests are no-ops)**
- **Backend**: `sentry-sdk[fastapi]==2.66.1`; `SENTRY_DSN` config (whitespace-stripped); `sentry_sdk.init` in `main.py` gated on the DSN. The global `@app.exception_handler(Exception)` now calls `capture_exception` — our handler catches every 500, so Starlette never sees it propagate and the ASGI integration alone would stay silent.
- **Frontend**: `@sentry/nextjs`; `instrumentation.ts` (server/edge init + `onRequestError`) and `instrumentation-client.ts` (client init), both DSN-gated; `error.tsx` + new `global-error.tsx` `captureException` (App Router boundary errors don't reach `window.onerror`, so the boundary must capture explicitly); `csp.ts` adds the Sentry ingest origin to `connect-src` only when a DSN is configured (else the browser blocks event delivery in prod). **Lean integration — no `withSentryConfig`** (see Known Limitations).

**AC2 — `logger.warning` before each per-user 429** (rate limiter #180 + AI quota #173) with `subject/path/limit`. (The `books.py` `AIRateLimitError` re-raises are already logged at the AI-service layer with a correlation id, so they aren't re-logged.)

**AC3 — perf-metrics production branch → real sink.** `sendToAnalytics` reports `poor`-rated metrics to Sentry (`captureMessage`, warning level, low-noise) and the write-only localStorage path is dropped. `getCachedMetrics`/`clearCachedMetrics` deleted — they only served the dropped path (no product consumers).

**Deploy**: `SENTRY_DSN_BACKEND` / `SENTRY_DSN_FRONTEND` wired into `deploy-staging.yml` and both `.env.example` files (empty secret = Sentry off).

## Tests
- **Backend**: Sentry config default/strip; global-handler `capture_exception` (with/without DSN); 429-warning caplog pins for both limiter and quota (**mutation-verified RED**); a **real-dep HTTP wiring pin** on the quota dependency.
- **Frontend**: metrics Sentry-sink (`poor`→Sentry, `good`→not, dev→not; localStorage tests removed), csp Sentry-origin-in-`connect-src`, `error`/`global-error` `captureException`.

## Verification
- Backend: 1123 passed (excl. the 2 network-bound files that hang offline; CI runs them keyed). Affected files + mutation checks green.
- Frontend: 2229 passed / 5 skipped; coverage 88.9 stmt / 79.0 br / 85.1 fn / 90.1 lines — all clear the 85/85/75/85 gate. (1 full-suite CPU-contention flake in `ChapterQuestionsPerformance`, green in isolation, unrelated.)
- `tsc --noEmit` 0 errors; eslint 0 errors; `next build` exit 0 with the Sentry integration.

## Reviews
Cross-family **codex** (pre-PR) caught a real **Major**: `request: Optional[Request] = None` on the quota dep is **not** recognized by FastAPI as the special Request injection — it raises `FastAPIError` at route registration and would **block backend startup**. The suite-wide conftest fake quota dep masked it (no route ever registered the real signature). Fixed to bare `request: Request = None` (verified: FastAPI injects it, no schema pollution, direct-call tests still default to None) and pinned with a real-dep HTTP wiring test that's RED against the old annotation.

## Known limitations
- The sink is live only once the DSN secret is set (an ops step) — same as the existing email/Stripe env plumbing.
- No source-map upload (no `withSentryConfig`); server stack traces are readable, client ones are minified until ops adds the Sentry build plugin + auth token.
- Secret presence, not liveness — an invalid DSN surfaces as Sentry's own init warning, not a startup failure (matches the DSN-gated, fail-open-to-console design).

### DIFF
--- a/.github/workflows/deploy-staging.yml
+++ b/.github/workflows/deploy-staging.yml
@@ -100,6 +100,8 @@ jobs:
           NEXT_PUBLIC_API_URL: ${{ secrets.API_URL }}
           NEXT_PUBLIC_BETTER_AUTH_URL: ${{ secrets.FRONTEND_URL }}
           NEXT_PUBLIC_ENVIRONMENT: staging
+          # Error tracking (#334) — inlined at build; empty secret = Sentry off.
+          NEXT_PUBLIC_SENTRY_DSN: ${{ secrets.SENTRY_DSN_FRONTEND }}
         run: |
           npm ci
           npm run build
@@ -246,6 +248,9 @@ jobs:
             echo "OPENAI_API_KEY=${{ secrets.OPENAI_API_KEY }}"
             echo "OPENAI_AUTOAUTHOR_API_KEY=${{ secrets.OPENAI_API_KEY }}"
             echo ""
+            echo "# Error tracking (#334) — empty secret => Sentry off"
+            echo "SENTRY_DSN=${{ secrets.SENTRY_DSN_BACKEND }}"
+            echo ""
             echo "# Better Auth Authentication"
             echo "BETTER_AUTH_SECRET=${{ secrets.BETTER_AUTH_SECRET }}"
             echo "BETTER_AUTH_URL=$FRONTEND_URL"
@@ -299,6 +304,8 @@ jobs:
             echo "EMAIL_SERVICE_PROVIDER=${{ secrets.EMAIL_SERVICE_PROVIDER }}"
             echo "EMAIL_SERVICE_API_KEY=${{ secrets.EMAIL_SERVICE_API_KEY }}"
             echo "EMAIL_FROM_ADDRESS=${{ secrets.EMAIL_FROM_ADDRESS }}"
+            # Error tracking (#334) — inlined at build below; empty = off.
+            echo "NEXT_PUBLIC_SENTRY_DSN=${{ secrets.SENTRY_DSN_FRONTEND }}"
           } > .env.production
 
           # Export environment variables for build process
@@ -309,6 +316,9 @@ jobs:
           export DATABASE_URL=${{ secrets.MONGODB_URI }}
           export DATABASE_NAME=${{ secrets.DATABASE_NAME }}
           export NEXT_PUBLIC_ENVIRONMENT=$ENVIRONMENT
+          # NEXT_PUBLIC_SENTRY_DSN must be present for the server-side build to
+          # inline it into the client bundle (#334); empty secret = Sentry off.
+          export NEXT_PUBLIC_SENTRY_DSN=${{ secrets.SENTRY_DSN_FRONTEND }}
 
           # Install dependencies and build
           # Remove old .next cache to prevent stale builds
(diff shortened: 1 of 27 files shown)


frankbria:
## Cross-family review (codex, pre-PR)

**One Major, fixed and pinned:**

> `check_quota(..., request: Optional[Request] = None)` is **not** recognized by FastAPI as the special `Request` injection — FastAPI only special-cases the bare `Request` type. A `Union`/`Optional[Request]` is treated as a normal field and raises `fastapi.exceptions.FastAPIError` at route registration, which would **block backend startup** on the 10 AI endpoints that wire the quota dep.

This was **untested** because the suite-wide conftest fake quota dep (no `request` param) is what every route actually registered during tests — the real signature never hit route analysis.

**Resolution** (verified empirically against FastAPI 0.139.0):
- `request: Optional[Request] = None` → `request: Request = None`. The bare `Request` annotation *is* injected by FastAPI at runtime; the `None` default only serves the direct-call unit tests. Confirmed: route registers cleanly, the real `Request` is injected (not None), no bogus `request` param in the OpenAPI schema.
- Added a **real-dep HTTP wiring regression test** (`test_quota_dep_registers_and_injects_request_via_http`) that mounts the genuine quota dependency on a route — **mutation-verified RED** against the old `Optional[Request]` annotation, GREEN with the fix. This closes the conftest-fake test-trust gap.

opencode (GLM) was attempted first as the cross-family primary but hung with no output (the documented condition); codex was the fallback. The GLM CI `review / review` job runs as a second independent reviewer on this PR.

PR #367, opened 2026-07-25
[#336] fix(questions): resolve chapter-question status filter before pagination

### LINKED ISSUE #336: [P1.4] Chapter-question status filter applied AFTER DB pagination corrupts page contents and counts
**Phase:** P1.4 — pre-launch. No dependencies.

**Problem:** `get_questions_for_chapter` builds the Mongo query without `status`, does `.sort().skip().limit()`, THEN applies the `status` filter in Python on the already-sliced page. `total` counts documents ignoring `status`, and `pages`/`has_more` derive from that unfiltered count while the response `total` is the post-filter count of one page.

**Evidence:** `backend/app/db/questions.py:236` (skip/limit before status), `:240` (count ignores status), `:268-278` (Python filter + page math), `:282`; endpoint `backend/app/api/endpoints/books.py:1849`. The only test uses `limit=50` with 5 questions, so the interaction is never exercised.

**Impact:** On a chapter with more questions than one page, `?status=completed` returns an inconsistent page, a wrong `total`, and paging skips raw (not status-matching) docs — a client can permanently miss matching questions.

**Acceptance criteria:**
- [ ] Resolve the status-matching id set (or `$lookup` aggregation) BEFORE skip/limit/count.
- [ ] Add a test with >1 page and a status filter asserting complete, non-overlapping results and an accurate total.

frankbria:
Closes #336.

## The bug

`get_questions_for_chapter` paged the `questions` collection with `status` absent from the Mongo query, then dropped non-matching rows from the already-sliced page in Python. A question's response status lives in a *different* collection (`question_responses`), so it could not be expressed in the initial query — the original code worked around that by filtering after `.sort().skip().limit()` had already chosen the page.

Everything a client decides on came back wrong:

- **`total` was the post-filter length of one page.** With four matching questions at `limit=2`, every page reported `total: 1`.
- **Pages came back short, or entirely empty.** `?status=draft&limit=2` returned `{"questions": [], "total": 0, "pages": 4, "has_more": true}` on a chapter that *has* a draft question — three fields describing three different sets in one body. A client reasonably reads that as "no drafts" and never finds the one that exists.
- **`pages`/`has_more` derived from a count that ignored `status`**, so they disagreed with `total` in the same response.

## The fix

Resolve the status-matching `_id` set **before** pagination and fold it into the query as `_id: {"$in": [...]}`, so `sort`/`skip`/`limit`/`count` all operate on the filtered set. Return the real `count_documents(query)` as `total`. The post-pagination Python filter is deleted.

This is the acceptance criteria's first-listed option (id set) rather than `$lookup`: `question_responses.question_id` is a *string* of `questions._id` (ObjectId), so an aggregation would need `$toString` (Mongo 4.0+) plus `$facet` for count-and-page. The id-set form is smaller and version-agnostic. The two extra reads run **only** when `status` is passed, so the unfiltered path is unchanged; the id list is bounded by the chapter's question count.

The page-enrichment loop now reuses the status map the filter already built instead of re-querying, keeping the #214 batched-lookup (no N+1) property.

## Acceptance criteria

| Criterion | Evidence |
|---|---|
| Resolve the status-matching id set before skip/limit/count | Demo: `main` returns an empty page with `total: 0` for a filter matching one question; the branch returns the question with a coherent `total`/`pages`/`has_more`. |
| Test with >1 page and a status filter asserting complete, non-overlapping results and an accurate total | `test_completed_filter_across_multiple_pages_is_complete_and_disjoint` — 2 pages at `limit=2`, union equals the completed set, no duplicates, `total == 4` on both pages — plus the HTTP-level `test_list_questions_status_filter_pages_completely`. |

## Tests

9 real-Mongo pins in `tests/test_db/test_question_status_pagination.py` (multi-page completeness/disjointness, no-empty-page, per-status, user scoping, composition with `category`, unfiltered `total` regression, unknown-status characterization) plus one endpoint wire-contract test.

**Mutation-verified three ways:**

- 5 of the 9 pins go RED against the **real pre-fix code** in a pristine `main` worktree (not a synthetic mutation).
- Disabling only the id-set resolution → the 5 status pins fail, the unfiltered `total` pin still passes.
- Reverting only `total=total` → `total=len(processed_questions)` → the 3 total-accuracy pins fail.

The 4 pins that pass on both sides are documented in the file as regression guards against defects the rewrite could have introduced, not as demonstrations of the bug.

Full backend suite: **1105 passed / 11 skipped, 92% coverage**. `questions.py` lint violations drop 14 → 11 (the deleted post-pagination filter held the long lines).

## Review

Cross-family pre-PR review by **`codex`** (opencode was occupied by a concurrent foreign-cwd session — the documented hang condition): *"The change applies response-status filtering before pagination/counting and updates totals consistently with the filtered result set. I did not find any discrete introduced bug that would break existing behavior or tests."*

## Demo

`docs/demos/2026-07-24-issue-336-question-status-pagination.md` — two real uvicorn servers (pristine `main` on :8801, branch on :8802) against the same Mongo, so the only variable is the application code.

## Known limitations

- **The `total` field changes meaning** from "page length" to "size of the filtered result set". This is what the acceptance criteria require ("an accurate total"). Verified no production consumer reads it: `bookClient.ts` consumes `pages` for its pagination loop (which becomes correct rather than changing meaning), and `question_generation_service.py:206` compares `total >= expected` for a persistence check, which only gets more lenient.
- **An unrecognised `status` value still returns unfiltered results.** The endpoint has never validated `status`, and a bare `if status:` with three `if/elif` branches has always let an unknown value fall through. Tightening that to a 422 is a separate breaking change; a characterization test pins the current behavior so the rewrite didn't silently turn it into "matches nothing".
- **The bug is latent in the shipped UI.** No production frontend code passes `status` — `QuestionContainer.tsx:83` calls `getChapterQuestions(bookId, chapterId)` with no options. This is a real public-API correctness bug, which is why the demo evidence is at the API layer.
- Not addressed (out of scope, pre-existing): that same call site defaults to `limit=10`, so the live questions UI only ever loads the first ten questions of a chapter.


frankbria:
## Cross-family review (post-PR) — `codex`

Independent review of the full PR diff by an OpenAI-family reviewer, run as a fresh session against the branch.

> The change applies response-status filtering before pagination/counting, preserves the documented lenient behavior for unknown status values, and adds regression coverage for multi-page filtered results. I did not identify a discrete introduced bug that would break existing behavior or tests.

**No Critical or Major findings.** A separate `codex` session reviewed the same code pre-PR and also came back clean.

**Reviewer selection note:** opencode (GLM, the normal primary) was occupied by a concurrent foreign-cwd session at pre-PR time and then hung silently for 12+ minutes on the post-PR attempt — the documented failure mode — so `codex` served as the cross-family reviewer for both passes. The GLM family did still review this PR independently via the in-CI `review / review` job, which reported **"✅ no defects found"** after verifying against the checked-out code and its callers rather than the diff alone.

PR #368, opened 2026-07-25
[#337] fix(toc): guard whole-TOC overwrites with optimistic lock (lost update off replica set)

### LINKED ISSUE #337: [P1.5] TOC add/update/delete/reorder do an unguarded whole-TOC overwrite (lost update off replica set)
**Phase:** P1.5 — pre-launch. No dependencies.

**Problem:** `_add_chapter_internal`/`_update_chapter_internal`/`_delete_chapter_internal`/`_reorder_chapters_internal` read the whole `table_of_contents`, mutate in memory, and `$set` it back filtered only on `{_id}` — no version guard, unlike the deliberately hardened `_update_toc_internal` (filters on `table_of_contents.version`) and the #177 autosave/summary paths. On standalone (non-replica-set) Mongo `session=None`, so it is plain last-write-wins.

**Evidence:** `backend/app/db/toc_transactions.py:230,317,398,579`. The live autosave path (`apply_chapter_content_update`) is a separate targeted write, so the two paths have divergent concurrency guarantees.

**Impact:** On standalone Mongo, a concurrent autosave and a chapter mutation silently clobber each other. Prod Atlas mitigates via transaction abort (500), but the two write paths are inconsistent and untested for concurrency.

**Acceptance criteria:**
- [ ] Add the same optimistic-lock filter (`table_of_contents.version` with `$exists:false` legacy fallback) and `modified_count` check used by `update_chapter_statuses_with_version_guard`.
- [ ] Add an interleaved-writer test proving no lost update on standalone.

frankbria:
Closes #337

## Problem

`_add_chapter_internal`, `_update_chapter_internal`, `_delete_chapter_internal` and
`_reorder_chapters_internal` (`backend/app/db/toc_transactions.py`) each read the whole
`table_of_contents`, mutate it in memory, and `$set` it back filtered **only on `{_id}`** — no
`owner_id`, no version guard, no `modified_count` check. On standalone (non-replica-set) Mongo
`session` is `None`, so this is plain last-write-wins: a concurrent autosave and a chapter
mutation silently clobber each other.

## Fix

Extracted one compare-and-swap helper, `_set_toc_guarded` — `update_one` filtered on
`{_id, owner_id, table_of_contents.version}` with an `$exists:false` fallback for legacy
version-less TOCs, plus the `modified_count == 0` check — and routed all four through it. Also
**deduped** the pre-existing `update_chapter_statuses_with_version_guard` (#159) onto the same
helper, so the guard now has a single implementation.

Two subtleties worth review attention:

- **`_version_guard()` must be called before mutation.** The callers mutate the dict they read
  *in place*, so reading `toc["version"]` at the write site would guard on the value about to be
  written and never detect a conflict. The snapshot is taken immediately after the read.
- **`modified_count` (not `matched_count`) is sufficient only by construction.** Every caller
  increments `version`, so a matched-but-unmodified write is impossible. This is documented in
  the helper as an invariant any future write through it must preserve — an idempotent write
  added later would need `matched_count`.

On a filter miss the helper re-reads to distinguish a genuine conflict from a book deleted or
reassigned mid-operation, so the API answers 404/403 rather than a misleading "modified by
another user" 409.

**API layer:** `Version conflict` → **409** added to create/update/delete chapter in
`chapters.py`. Without this the new `ValueError` fell through to a 400 carrying the raw internal
message — a regression the fix would otherwise have introduced.

## Demo — outcome evidence

Reproduction against the real local mongod (standalone, `session=None`), interleaving an autosave
into the read→write window of `add_chapter_with_transaction`. The evidence is **whether the
autosaved content still exists afterwards**, not merely whether an error was raised:

| | User A's write | User B's autosaved content |
|---|---|---|
| **Pre-fix** (`HEAD~1`) | completed with **no error** | **GONE — lost update** |
| **Post-fix** | rejected: `ValueError(Version conflict: TOC was updated by another process)` | **survived intact** |

```
=== WITHOUT FIX (pre-#337 code) ===
   [User A] add_chapter_with_transaction(...) starting -- reads TOC v1
   [User B] autosave committed: chapter c1 content saved, version -> 2
   [User A] completed with NO error
Chapters           : ['Chapter One', 'Chapter Two']
User B's autosaved content survived? NO -- LOST UPDATE

=== WITH FIX ===
   [User A] add_chapter_with_transaction(...) starting -- reads TOC v1
   [User B] autosave committed: chapter c1 content saved, version -> 2
   [User A] rejected -> ValueError(Version conflict: TOC was updated by another process)
Chapters           : ['Chapter One']
User B's autosaved content survived? YES
```

## Acceptance criteria

- [x] Same optimistic-lock filter (`table_of_contents.version` with `$exists:false` legacy
      fallback) and `modified_count` check as `update_chapter_statuses_with_version_guard`
- [x] Interleaved-writer test proving no lost update on standalone

## Tests

16 new tests. Interleaved-writer tests parametrized across all four ops fire a competing TOC
write in the read→write window and assert **both** that the loser raises **and** that the
winner's data survives byte-intact — plus legacy version-less, owner-reassigned, and
book-deleted variants, and three API-level tests asserting 409.

Verified red-before/green-after: 8 lost-update assertions failed pre-fix; the 409 tests returned
400 with the endpoint mapping reverted.

- Backend suite: **1124 passed, 11 skipped, 0 failures**
- `toc_transactions.py` coverage: **94%** (remaining misses are the replica-set
  `start_transaction()` branches, unreachable on a standalone test server — already documented in
  the test module header)

## Review

Cross-family review by opencode (GLM). Three findings; dispositions:

| Finding | Severity | Disposition |
|---|---|---|
| Replica-set transaction path surfaces a conflict as 500, not 409 | Medium | **Documented, not fixed** — see Known Limitations |
| `modified_count == 0` conflated conflict with book deleted / ownership transferred | Low | **Fixed** — helper now re-reads and raises `Book not found` / `Not authorized` / `Version conflict` distinctly |
| `version: null` TOC raises `TypeError` | Low | **Not fixed** — pre-existing and unreachable; nothing writes null, and the `+ 1` increment fails before the guard is consulted |

## Known limitations

- **Replica sets still return 500, not 409, on a concurrent edit.** These four functions run
  inside `session.start_transaction()` when a replica set is detected; there the `update_one`
  reads at the transaction snapshot, so the version filter matches and the conflict surfaces at
  commit as a WriteConflict → generic `OperationFailure` → 500. This is pre-existing behavior the
  issue explicitly acknowledges ("Prod Atlas mitigates via transaction abort (500)") and is
  **not** a regression — but the 409 contract this PR adds holds only on standalone. Left
  unchanged because it cannot be verified without a replica-set test fixture.
  **Follow-up option:** drop the transaction wrapper for these single-document writes, which is
  what `update_chapter_statuses_with_version_guard` already does deliberately (its docstring
  spells out the reasoning) — the CAS is topology-independent and sufficient for a single-doc
  write.
- `reorder_chapters_with_transaction` is imported in `books.py` but has **no live calling
  endpoint**, so it needs no 409 mapping today. If a reorder route is reintroduced it will need
  the same handling added here to `chapters.py`.
- `_update_toc_internal` was deliberately left alone: it already carries the guard plus a bespoke
  conflict-vs-generic-write-failure distinction that existing tests depend on.


frankbria:
## Cross-family review record (opencode / GLM)

Two passes, per the issue-lifecycle review invariants.

**Pass 1 — pre-PR, on the initial diff.** Three findings:

| Finding | Sev | Disposition |
|---|---|---|
| Replica-set transaction path surfaces a conflict as 500, not 409 | Medium | **Documented, not fixed.** Pre-existing behavior the issue explicitly acknowledges ("Prod Atlas mitigates via transaction abort"); fixing it means dropping the transaction wrappers, which can't be verified without a replica-set fixture. See Known Limitations in the PR body. |
| `modified_count == 0` conflated a version conflict with the book being deleted / reassigned | Low | **Fixed** in `fa382dd` — the helper now re-reads on a filter miss and raises `Book not found` / `Not authorized to modify this book` / `Version conflict` distinctly, so the API answers 404 / 403 / 409 instead of a blanket misleading 409. |
| `version: null` TOC raises `TypeError` | Low | **Not fixed (YAGNI).** Pre-existing and unreachable — nothing writes null, and `toc.get("version", 1) + 1` fails before the guard is ever consulted. |

**Pass 2 — post-PR, on the final diff** (which includes the re-read the first pass never saw). Targeted at the new code: re-read misclassification, endpoint substring shadowing, transaction behavior, and whether any test asserts less than it claims.

Verdict: **no defects.** The reviewer traced all three re-read messages through every `except ValueError` chain in `chapters.py` and confirmed no shadowing — the disambiguated messages are deliberately free of overlapping substrings, and the re-read notably *avoids* the legacy `"Book not found or not authorized"` ambiguity that would otherwise hit the `"not found"` branch first. It also independently confirmed `_version_guard` is captured before in-place mutation on all four paths, and that the bulk-status refactor avoids the aliasing trap (`updated_toc` is a fresh dict).

One actionable note: the new 403/404 re-read paths were covered only at the db layer, never driven through the endpoints' substring dispatch — precisely where a shadowing bug would live. Closed in `e1ded36` with parametrized API tests across create/update/delete asserting 403 on an ownership change and 404 on a book deletion in the read→write window.

One non-blocking note declined: the final `else` concludes "version conflict" by elimination without re-checking that the re-read version actually differs. The reviewer classified this as effectively theoretical — on standalone Mongo a `$set` of changed fields can't return `modified_count == 0` with a matching filter, and on replica sets uncertain commits raise rather than return 0. Not worth the extra branch.

Backend suite after both commits: **1130 passed, 11 skipped, 0 failures.**

Issue #338, opened 2026-07-23
[P1.6] ensure_question_indexes single try/except can silently skip the unique duplicate-response index


frankbria:
**Phase:** P1.6 — pre-launch. No dependencies.

**Problem:** All index builds sit in one try/except that only logs. If the unique `question_user_idx` on `question_responses` fails (e.g. the collection already holds duplicates), every subsequent `create_index` is skipped and the app serves writes without the uniqueness guarantee the #242 dedup fix relies on. `ensure_user_indexes` already wraps each index individually.

**Evidence:** `backend/app/db/questions.py:51-118` (single try/except), `:86-92` (unique index).

**Impact:** `save_question_response`'s non-atomic find-then-insert can persist two response docs for one `(question_id,user_id)`; the ratings unique index is also skipped.

**Acceptance criteria:**
- [ ] Wrap each `create_index` in its own try/except and log the specific failing index at ERROR so operators know the uniqueness guard is missing.
- [ ] One test asserting a mid-list failure doesn't skip later indexes.

Issue #328, opened 2026-07-23
CI: global functions coverage gate (85%) chronically red on main (~84.6%) — every frontend PR needs admin override


frankbria:
## Problem

The frontend CI `Check coverage threshold` step enforces `global.functions >= 85%`, but `main` itself sits at **~84.56–84.64%** and has since before #215 (2026-07-20). As a result the `Frontend Tests` required check is **red on essentially every frontend PR**, even when 100% of tests pass and the diff doesn't touch coverage.

This forces an **`--admin` branch-protection override on every frontend merge**, which quietly defeats the #118 enforcement goal. Documented occurrences: #215, #237, #239, #276, #280, and now #327 (#295).

## Evidence (PR #327, a pure −257/+0 dead-code deletion)
```
Test Suites: 124 passed, 124 total
Tests:       5 skipped, 2203 passed, 2208 total
Jest: "global" coverage threshold for functions (85%) not met: 84.64%
```
Lines/statements/branches gates pass; only **functions** is under, by <0.5pp, repo-wide.

## Why it recurs
The gap is a repo-wide denominator fact, not a per-PR regression. Small/deletion PRs can't move it. Nothing has closed the ~0.4pp functions gap, so the gate stays red indefinitely and admin override became the norm.

## Options (pick one)
1. **Close the gap honestly** — add function-level tests to the lowest-covered modules until global functions ≥85% (durable, restores real enforcement). Requires identifying the under-covered files (`jest --coverage` per-file report).
2. **Lower the functions threshold to the true baseline** (e.g. 84% or 80%) so the gate reflects reality and stops false-blocking, then ratchet up as coverage improves.
3. **Per-directory thresholds** instead of one global functions number, so new code is held to 85% while legacy under-coverage doesn't block unrelated PRs.

Recommend (1) or (3) so branch protection means something again; (2) is the quick stopgap.

## Acceptance
- [ ] Frontend PRs with all tests passing no longer require `--admin` to merge
- [ ] The chosen approach is documented in CLAUDE.md / CI config

_Filed from the #295/PR #327 merge — the admin override there is the trigger._

Issue #342, opened 2026-07-23
[P2.1] Supply-chain automation mis-aimed: Dependabot watches an empty root, no audit gate, dead crypto deps


frankbria:
**Phase:** P2.1 — post-launch fast-follow. No dependencies.

**Problem:** Dependabot's npm ecosystem targets `directory:"/"` (root `package.json` is a PM2 shell with zero deps and no lockfile), so the real 977-package `frontend/` tree is unmonitored. No workflow runs `npm audit`/`pip-audit`/`osv-scanner`. Dead backend deps still ship: `python-jose` (zero imports, drags in `ecdsa` Minerva timing CVE) and `passlib` (unused, unmaintained since 2020). A stray untracked `frontend/scripts/package-lock.json` pulls in the `colors` protestware package.

**Evidence:** `.github/dependabot.yml:6`; `backend/pyproject.toml:23-24`; `frontend/scripts/package-lock.json`; no scanner step in `.github/workflows/`.

**Impact:** Frontend dependency advisories/updates never fire; unused crypto libs are needless attack surface; no reproducible in-repo audit gate.

**Acceptance criteria:**
- [ ] Point Dependabot at `/frontend`.
- [ ] Add a committed `pip-audit` + `npm audit --audit-level=high` (or osv-scanner) CI step.
- [ ] Remove `python-jose`/`passlib` and re-lock.
- [ ] Delete `frontend/scripts/` (or track it and fold into Dependabot).

frankbria:
## Premise check before anyone starts

Verified each claim against the current tree (`main` @ `ee1da24`). **The two headline problems are real and worth fixing. Two sub-claims are wrong as written and would mislead an implementer** — flagging now rather than after someone hits them.

### ✅ Confirmed

**Dependabot is mis-aimed.** `.github/dependabot.yml:7` targets `directory: "/"` for npm. The root `package.json` has `dependencies: {}` and `devDependencies: {}` with **no `package-lock.json`**, so that ecosystem watches nothing. Meanwhile `frontend/package-lock.json` exists with **1122 package entries** (issue says 977 — either way, none of it is monitored). This is the core of the issue and it holds exactly as described.

**No audit gate runs.** `npm audit` appears only in `deploy-production.yml.disabled`, `deploy-production.tbd`, and `test-suite.tbd` — none of which execute. No `pip-audit`/`osv-scanner` anywhere. Confirmed.

[...]

PR #371, opened 2026-07-26
chore: commit pending worktree files (launch backlog + scripts lockfile)

frankbria:
Housekeeping only — no source or config changes.

Two files had been sitting uncommitted in the working tree. Beyond the drift itself, this made every `gh pr merge --delete-branch` fail its local-sync step with `cannot pull with rebase: You have unstaged changes` — the remote merge succeeded each time, but the local checkout was left behind, which reads like a failed merge.

## `tasks/todo.md`

The committed copy was stale notes from #328 (frontend coverage gate). The working tree holds the `/saas-launch-review` backlog generated 2026-07-22 — the 23-issue P0→P3 plan that #329 through #338 have been worked from. #338, merged in #370, is the "index guard" row in its severity table.

Commits the version actually in use.

## `frontend/scripts/package-lock.json`

`frontend/scripts/package.json` is tracked and `frontend/package-lock.json` is tracked, so repo convention is to commit lockfiles — this one was simply never added. Pins `axios@1.10.0` + `colors@1.4.0` for `scripts/system-test.js`.

Verified before committing: resolves against `registry.npmjs.org` only, contains no `_auth`/token/credential entries, and the pinned versions satisfy the `^1.10.0` / `^1.4.0` ranges in `package.json`.

## Note

`tasks/todo.md` is a tracked file that gets rewritten per work item, so it will go dirty again the next time it is regenerated. Committing it fixes the current drift but not the recurrence — worth deciding separately whether it should stay tracked or move to a gitignored scratch path.


frankbria:
Rebuilt on current main and split the two halves, since `tasks/todo.md` had gone conflicting exactly as this PR predicted.

**`frontend/scripts/package-lock.json` — committed.** Still untracked on main, still wanted. Re-verified independently before committing: lockfileVersion 3, 25 entries, every `resolved` points at `registry.npmjs.org`, no `_auth`/token/password fields, and `axios@1.10.0` + `colors@1.4.0` satisfy the `^1.10.0` / `^1.4.0` ranges in the tracked `package.json`.

**`tasks/todo.md` — untracked and gitignored instead of committed.** The original version here (the 2026-07-22 launch backlog) now conflicts with main, which holds the newer #342 supply-chain notes. Committing either copy only resolves this one instance; the file is rewritten in full per work item, so it would conflict again on the next regeneration. Untracking it addresses the recurrence this PR flagged in its own closing note — including the dirty-checkout side effect that made `gh pr merge --delete-branch` fail its local sync step and read like a failed merge. The file stays on disk locally, it is just no longer tracked.

Note the previous head failed CI on a flake, not a real problem: the suite passed clean in `Run unit tests` (130 suites, 2241 passed), then `Check coverage threshold` re-ran it instrumented and exactly one test fell over — `VoiceTextInput Component › Continuous dictation (accumulation) › covers interim + multiple final results via the mock sequence`. Coverage itself passed comfortably (88.91 stmts / 79.02 branch / 85.17 funcs / 90.14 lines vs the 85/85/75/85 floor). Worth a separate look at that test — it is asserting on `SpeechRecognition` mock timing it does not control, so Istanbul overhead reorders it.

Issue #199, opened 2026-07-04
[P2.7] Restore real rate-limit test coverage (limiter is neutered suite-wide)


frankbria:
**Problem:** `conftest.py` monkeypatches the limiter to a no-op for the whole suite, and the one endpoint-level rate-limit test is `@pytest.mark.skip`. So removing `Depends(get_rate_limiter(...))` from any AI-generation or export endpoint (the cost/abuse/DoS surface — 27 wired call sites) passes the entire suite green.

**Evidence:** `backend/tests/conftest.py:37-55` (no-op limiter), `backend/tests/test_api/test_export_endpoints.py:463` (skipped).

**Acceptance criteria:**
- ≥1 un-skipped integration test per protected endpoint class (AI generate, TOC, export, avatar upload) using the real limiter with a small cap, asserting the Nth+1 request → 429.

Issue #174, opened 2026-07-04
[P0.2] Billing / entitlement gate for paid access (epic — decompose)


frankbria:
**Problem:** No payment/subscription/entitlement code exists. For a paid SaaS this is a launch blocker. For a *closed* beta the pragmatic path is invite-gating + a free tier now with Stripe later — but an **entitlement check** (is this user allowed, on what plan) must exist so per-user AI caps (P0.1) can key off plan and so paid launch isn't a rebuild.

**Evidence:** `grep -riE 'stripe|billing|subscription|payment|entitlement|plan_tier' backend/app frontend/src` returns nothing but RxJS `form.watch` false-positives.

**Acceptance criteria:**
- A documented decision: beta = free-invite vs paid-at-launch.
- User records carry a `plan`/`entitlement` field (even if all beta users are one free plan).
- AI feature access checks entitlement; unauthorized/over-limit users get a clear path.
- (If paid-at-launch) a decomposition into atomic sub-issues for the payment provider integration.

**Blocks:** P0.1 keying caps by plan. **This is an epic — split before implementing.**

PR #373, opened 2026-07-27
[P1.8] test(ai): route-wiring guard + per-route 402/429 for AI quota & entitlement (#340)

### LINKED ISSUE #340: [P1.8] AI quota + entitlement deps have no route-wiring guard — droppable from 9/10 endpoints while CI stays green
**Phase:** P1.8 — pre-launch. No dependencies.

**Problem:** All 10 AI endpoints declare `Depends(get_ai_usage_quota())` + `Depends(get_entitlement_checker(...))`, both globally no-op'd in conftest. The rate limiter has a route-completeness pin (#199) that fails naming any route missing it; quota and entitlement have none. Entitlement is route-tested on exactly 1 of 10 endpoints; quota has zero HTTP-level tests (only isolated invocation).

**Evidence:** `tests/test_api/test_entitlement_gate.py:63` (only `/generate-toc`); `tests/test_api/test_rate_limit_routes.py:381` (the guard that exists for the limiter); conftest no-ops.

**Impact:** Removing `Depends(get_ai_usage_quota())` (cost control) or the entitlement dep from most AI routes leaves the whole suite green — cost/paywall enforcement can silently regress.

**Acceptance criteria:**
- [ ] Extend the `_walk_effective_routes` completeness pin to assert every AI endpoint declares BOTH deps.
- [ ] Add restricted-user→402 and real-quota→429 route-level tests parameterized across all 10 AI endpoints.

frankbria:
Closes #340.

**Test-only change — no production code is modified.**

## The hole

All 10 AI endpoints declare `Depends(get_ai_usage_quota())` (cost control, #173) and `Depends(get_entitlement_checker(...))` (paywall hook, #174). But the suite globally no-ops the quota and the entitlement gate self-disables under `BYPASS_AUTH`, so either dependency could be deleted from 9 of the 10 endpoints and CI would stay green.

**Verified, not assumed.** With `Depends(get_entitlement_checker("enhance_text"))` deleted from `enhance-text`, the pre-existing `test_entitlement_gate.py` + `test_ai_usage_quota.py` + `test_rate_limit_routes.py` reported **22 passed**.

## What was added

`backend/tests/test_api/test_ai_gate_routes.py` — 34 tests:

| Class | Coverage |
|---|---|
| `TestAIGateWiringCompleteness` | Walks the live FastAPI route tree; names any endpoint that lost the quota dep, the entitlement dep, or gates on the **wrong feature key**. Plus a drift guard so a new AI route can't appear without denial coverage. |
| `TestEntitlementDenialPerRoute` | restricted plan → **402** across all 10; entitled plan → **404** (reached the handler) across all 10. |
| `TestQuotaExhaustionPerRoute` | Real quota armed against real Mongo counting → **429** with production headers, across all 10. |

### The enabler (why no HTTP-level quota test existed)

`conftest.fake_get_ai_usage_quota` returned a **fresh `_always_allow` closure per route**, so there was no single `dependency_overrides` key to re-arm the real quota. It now returns **one shared module-level `noop_ai_quota`** (mirroring `noop_rate_limiter`), and a new `arm_real_ai_quota` fixture arms the genuine quota on every AI route at once — the same shape `arm_real_rate_limiter` gave #199.

`_walk_effective_routes` / `_route_dependency_calls` moved out of `test_rate_limit_routes.py` into `backend/tests/route_introspection.py` so both pins share one walker. Imported as `tests.route_introspection` (same shape as the existing `tests.db_guard`) — deliberately **not** `tests.conftest`, which double-executes conftest and rebinds the shared no-ops to objects no route was built against.

## Mutation evidence (each guard fails RED and names the route)

| Mutation | Result |
|---|---|
| Drop quota dep from `generate-draft` | 2 failed — quota pin **and** that route's 429 test |
| Drop entitlement dep from `enhance-text` | 3 failed — entitlement pin, feature pin, **and** that route's 402 test |
| Gate `analyze-summary` on `"generate_toc"` (wrong key) | 1 failed — the feature pin, **and nothing else** |

`books.py` restored clean after each; `git diff --quiet` verified.

## Review

Third-party review by **opencode (GLM)** ran pre-PR. Six findings triaged, five fixed in `85cd800`:

- **Fixed** — `!= 402` companion assertion was near-tautological (a 500 satisfied it) → now asserts 404.
- **Fixed** — wrong entitlement feature key was invisible to every test → new pin reads the closure cell.
- **Fixed** — `arm_real_ai_quota` didn't patch `E2E_EXEMPT_EMAILS` itself → an exemption in the ambient env could silently un-meter the test user.
- **Fixed** — the fixture referenced this module's `noop_ai_quota` instead of deriving the route-captured one by name → the documented double-execution could install an override that never fires.
- **Fixed** — drift test's bare `next(iter(...))` could raise `StopIteration` instead of asserting.
- **Answered, not changed** — "the 429 test doesn't prove metering happens *before* the AI spend." It does, by construction: the first request 404s out of the handler before any AI call yet still consumes quota, so a counter that only rose after a successful generation could never produce the second request's 429. Documented in the test rather than mocking 10 different AI service entry points.

## Known limitations

- **Monthly-window enforcement is not exercised here** — `arm_real_ai_quota` disables it so the daily cap alone drives the 429 and the `X-AI-Quota-Period: day` header is deterministic. Monthly is covered by `test_ai_usage_quota.py::test_quota_monthly_window_enforced`.
- **Day-boundary flake, theoretical** — unlike `arm_real_rate_limiter`, the clock isn't frozen: the quota keys off `datetime.now(...)` buckets, not the `deps.time.time` seam. Exposure is the sub-10ms window between a test's two requests straddling midnight UTC. Judged not worth patching the module-wide `datetime`; rationale is in the fixture docstring.
- **The pin is order-insensitive** — it reads `dependant.dependencies` as a set, so it won't notice quota and entitlement swapping positions. Both still run before the handler, so this is cosmetic.
- Pre-existing `tests/conftest.py` ruff findings (25) are untouched; the new files are ruff-clean.

## Verification

- `1199 passed, 11 skipped` — full backend suite (excluding `test_e2e_no_mocks.py` / `services/test_ai_service.py`, which hang without network locally; CI covers them).
- New files pass `ruff check`.


frankbria:
## Third-party review (opencode / GLM) — on the record

Two passes ran against this branch: pre-PR on `34231a0`, and this second pass on final HEAD `85cd800` scoped to checking the fixes.

### Pass 1 (`34231a0`) — 6 findings

| # | Finding | Disposition |
|---|---|---|
| 1 | `test_entitled_plan_is_not_blocked` asserted `!= 402`, which a 500 or a deleted handler also satisfies | **Fixed** — asserts `404` (request must reach the handler's book lookup) |
| 2 | Wrong entitlement *feature key* on a route was invisible to every test | **Fixed** — new `test_every_ai_route_gates_on_its_own_feature` reads the closure cell |
| 3 | `arm_real_ai_quota` didn't patch `E2E_EXEMPT_EMAILS` itself | **Fixed** — patched in the fixture, so an ambient-env exemption can't silently un-meter the test user |
| 4 | Fixture referenced the module-global `noop_ai_quota` instead of the route-captured one | **Fixed** — key derived by name from the live routes, killing the conftest double-execution landmine |
| 5 | Drift test's bare `next(iter(...))` could raise `StopIteration` instead of asserting | **Fixed** — guarded |
| 6 | 429 test doesn't prove metering happens *before* the AI spend | **Answered, not changed** — it does by construction: the first request 404s out of the handler before any AI call yet still consumes quota, so a counter that only rose after a successful generation could never produce the second request's 429. Documented in the test rather than mocking 10 AI service entry points. |

Finding 2 is the substantive one and was outside the issue's stated AC. It matters because all 10 decorators are near-identical, and today every plan allows every feature — so a copy-pasted wrong key passes the name check, the restricted-plan 402 test, and the free-plan test, and would only surface the day a paid tier splits the features.

### Pass 2 (`85cd800`) — verdict: all fixes sound, no new defects

> **(1) `_entitlement_features` closure read — works & robust.** `feature` is a true freevar with a cell; `co_freevars` / `__closure__` are positional-parallel tuples. Graceful degradation: if a future refactor drops the closure, the guard skips it, `features` stays `set()`, and the assertion surfaces `(expected, set())` — loud, named, not a silent pass.
>
> **(2) key-by-name + teardown — correct & leak-free.** Deriving the key by walking the live routes is *necessary*, not just defensive. `patches` accumulates setup + per-`arm` patches, all stopped in reverse; `armed_keys` popped idempotently.
>
> **(3) tightened 404 assertions — coupling correct and justified.** Every handler opens with the book lookup before ownership (403) or any AI work; bodies are `Body(default={})` so `json={}` is accepted. Legitimate future breakage would fail loudly — "that's the test doing its job, not brittle coupling."
>
> **(4) No new defects.** The `walk_effective_routes` extraction is behavior-preserving; both rate-limiter call sites updated consistently.

Pass 1 also raised two items recorded as **Known Limitations** in the PR body rather than fixed: the monthly quota window isn't exercised here (covered by `test_ai_usage_quota.py::test_quota_monthly_window_enforced`), and the pin is order-insensitive within `dependencies=[...]` (cosmetic — both deps still run before the handler).

frankbria:
## Demo verification — acceptance criteria → outcome evidence

Test-only backend change, so the demo is executable: every guard is shown *failing* when the thing it protects is removed, and the baseline is shown *not* catching it.

### Baseline: the hole is real (not assumed)

Deleted `Depends(get_entitlement_checker("enhance_text"))` from `books.py:2714`, then ran only the tests that existed before this PR:

```
$ uv run pytest tests/test_api/test_entitlement_gate.py \
                tests/test_api/test_ai_usage_quota.py \
                tests/test_api/test_rate_limit_routes.py -q
============================== 22 passed in 0.81s ==============================
```

Paywall enforcement silently gone from an endpoint; suite green. That is issue #340.

---

### AC 1 — "Extend the completeness pin to assert every AI endpoint declares BOTH deps"

**Evidence — the pin fails RED and names the exact route.**

Mutation: drop the quota dep from `generate-draft` (`books.py:2512`).
```
FAILED ...TestAIGateWiringCompleteness::test_every_ai_route_declares_the_usage_quota
FAILED ...TestQuotaExhaustionPerRoute::test_call_past_the_cap_gets_429[POST /api/v1/books/{book_id}/chapters/{chapter_id}/generate-draft]
========================= 2 failed, 32 passed in 0.71s =========================
```

Mutation: drop the entitlement dep from `enhance-text` (`books.py:2714`) — the same deletion the baseline above missed.
```
FAILED ...TestAIGateWiringCompleteness::test_every_ai_route_declares_an_entitlement_check
FAILED ...TestAIGateWiringCompleteness::test_every_ai_route_gates_on_its_own_feature
FAILED ...TestEntitlementDenialPerRoute::test_restricted_plan_gets_402[POST /api/v1/books/{book_id}/chapters/{chapter_id}/enhance-text]
========================= 3 failed, 31 passed in 0.73s =========================
```

Bonus mutation (beyond the AC): gate `analyze-summary` on `"generate_toc"` — a plausible copy-paste, and *behaviourally invisible* today since every plan allows every feature.
```
FAILED ...TestAIGateWiringCompleteness::test_every_ai_route_gates_on_its_own_feature
========================= 1 failed, 33 passed in 0.72s =========================
```
Caught by the new feature-key pin and by nothing else.

`app/api/endpoints/books.py` restored after each mutation; `git diff --quiet` confirmed clean.

### AC 2 — "restricted-user→402 and real-quota→429 route-level tests parameterized across all 10 AI endpoints"

**Evidence — 34 tests, green, with all 10 endpoints individually named in the test ids.**

```
$ uv run pytest tests/test_api/test_ai_gate_routes.py -q
============================== 34 passed in 0.66s ==============================
```

Breakdown: 4 wiring pins + (10 × 402 denial) + (10 × entitled-plan reaches handler) + (10 × real-quota 429). The 429 tests drive the genuine `check_quota` against real Mongo-backed counting and assert the production contract — `X-AI-Quota-Limit: 1`, `X-AI-Quota-Period: day`, and the "limit reached" body.

### No collateral damage

```
$ uv run pytest tests/ -q   # minus test_e2e_no_mocks.py + services/test_ai_service.py (need network locally; CI covers them)
================ 1199 passed, 11 skipped, 12 warnings in 55.93s ================
```

The `conftest` quota-fake reshaping touches every test in the suite; nothing regressed.

Issue #50, opened 2025-12-18
Fix accessibility issues - ensure WCAG 2.1 Level AA compliance


frankbria:
## Problem
The application has partial accessibility implementation with gaps in ARIA labels, keyboard navigation, screen reader support, and color contrast.

## Current Behavior
- Some interactive elements lack ARIA labels
- Keyboard navigation incomplete in some areas
- Screen reader announcements inconsistent
- Some color contrasts below WCAG standards

## Expected Behavior
- Full WCAG 2.1 Level AA compliance
- All interactive elements keyboard accessible
- Proper ARIA labels throughout
- Screen reader friendly navigation
- Sufficient color contrast everywhere

## Implementation Guidance
1. Run accessibility audit with axe DevTools or Lighthouse
2. Review existing accessible components as patterns (reference CLAUDE.md note about WCAG 2.1 compliance)
3. For each issue found:
   - Add missing ARIA labels
   - Ensure keyboard focus management
   - Test with screen reader (NVDA/JAWS)
   - Fix color contrast issues
4. Reference accessible patterns in `frontend/src/components/chapters/ChapterEditor.tsx`

## Files Likely Involved
- All components in `frontend/src/components/`
- Focus on interactive components: buttons, forms, modals, tabs
- Check all color definitions in Tailwind config

## Acceptance Criteria
- [ ] Lighthouse accessibility score >95
- [ ] All interactive elements have ARIA labels
- [ ] Full keyboard navigation works (tab, arrow keys, enter, escape)
- [ ] Screen reader announces all important actions
- [ ] Color contrast ratios meet WCAG AA standards
- [ ] Focus indicators clearly visible
- [ ] Skip links available for navigation
- [ ] Automated accessibility tests pass

## Blocked by
Issue #44 (Fix broken UI elements)

PR #374, opened 2026-07-27
[P1.9] test: delete hollow tests + real-browser rich-text formatting E2E (#341)

### LINKED ISSUE #341: [P1.9] Hollow/tautological tests + untested rich-text editing inflate green CI
**Phase:** P1.9 — pre-launch. No dependencies.

**Problem:** Several tests validate nothing about the product. `MockSystemFlow.test.tsx` asserts jest mocks return their configured values (renders nothing, prints "All core workflows validated"). `test_debug_chapter_questions.py` has zero asserts (only prints) — a collected test that cannot fail. `TestInfrastructureIntegration.test.tsx`/`example.test.tsx` test the harness. Separately, TipTap is globally mocked in unit tests and E2E only inserts draft text, so no test confirms Bold/Italic/Heading actually produce formatted output despite it being a headline capability.

**Evidence:** `src/__tests__/MockSystemFlow.test.tsx:183`; `tests/test_debug_chapter_questions.py:83`; `src/__tests__/TestInfrastructureIntegration.test.tsx:57`; `src/jest.setup.ts:265` + `RichTextEditor.test.tsx:154` (`toHaveBeenCalledTimes(7)` against a mock).

**Impact:** Pass-count overstates confidence; a total break in create→TOC→draft or in formatting commands ships green.

**Acceptance criteria:**
- [ ] Delete the tautological/zero-assert/harness tests (#200/#210 precedent).
- [ ] Add one real-browser E2E that applies Bold/Heading/List via the toolbar and asserts the resulting marks/HTML.
- [ ] Drop the brittle mock call-count assertion.

frankbria:
Closes #341.

## The problem

Four tests inflated the pass count while validating nothing, and the headline rich-text editing capability had no test that could prove it works.

### 1. Deleted the hollow tests (#200/#210 precedent)

| File | Why it validated nothing | Why deleting is safe |
|---|---|---|
| `frontend/src/__tests__/MockSystemFlow.test.tsx` | Mocked `bookClient` wholesale, rendered no component, then asserted the mocks returned their own configured values. Printed "All core workflows validated". | Zero product coverage — every call hit a `jest.mock`. Real journey coverage lives in `src/e2e/complete-authoring-journey.spec.ts`. |
| `frontend/src/__tests__/TestInfrastructureIntegration.test.tsx` | Asserted `typeof describe === 'function'` and that the router mock returns the mock. Rendered only inline throwaway components. | Zero product coverage; `grep` confirms **no importer** of its exported `testUtils` / `renderWithSuspense` / `renderWithErrorBoundary`. |
| `frontend/src/__tests__/example.test.tsx` | Scaffold: renders a local `<Hello>`, asserts it says hello. | Zero product coverage. |
| `backend/tests/test_debug_chapter_questions.py` | **Zero asserts**, `print()` only — collected by pytest, unable to fail. | Same service call covered with real asserts by `test_services/test_question_generation_service.py`; the near-identical sibling `test_debug_questions.py` makes the same call with 3 asserts. |

**Root-cause sweep** (per the #210 "sweep the class, not just the cited file" precedent): scanned all of `backend/tests` for collected-but-assertless files. Three hits, triaged individually — one deleted (above), `test_e2e_no_mocks.py` **kept** (no asserts of its own but delegates to `SystemE2ETest.run()`, whose `raise_for_status()` calls give it teeth), and `test_system_e2e_simplified.py` **fixed** (below).

### 2. New real-browser formatting E2E

`frontend/src/e2e/editor-formatting.spec.ts` — TipTap is globally mocked in jsdom (`src/jest.setup.ts`) because it cannot run there, so the unit layer can only ever observe that *a command ran*, never that formatted output appeared. The existing E2E only inserted plain draft text.

The new spec drives a real browser with real TipTap against the real backend, clicks the actual toolbar, and asserts the resulting marks in **both** the live DOM and the auto-saved backend HTML — formatting that renders but never persists is still broken. Covers Bold, Heading 2, Bullet List, and toggle-off. Deterministic seeded books (no AI), so it runs in CI without an OpenAI key.

This lands in `src/e2e/`, the root `playwright.config.ts` `testDir`, which `tests.yml` runs as a required check — so it genuinely gates.

### 3. Dropped the brittle mock call-count assertion

`RichTextEditor.test.tsx` asserted `expect(mockRunFn).toHaveBeenCalledTimes(7)` against a shared mock, with the comment "the mock chain gets called more times". A magic number that passed whether or not each button was wired to the right command. Replaced with per-command assertions: clicking Bold must call `toggleBold`, H1 must call `toggleHeading({level: 1})`, and so on. This layer honestly proves only *button → correct command*; real output is proven by the E2E above.

### 4. A test that printed "Has TOC: False" and reported PASSED

`backend/tests/test_system_e2e_simplified.py` generated a TOC, verified the book had none, and passed. **Two defects cancelling out:** `verify_system()` read `book["table_of_contents"]` — the *Mongo document's* field name, which `BookResponse` does not declare, so `.get()` returned `{}` unconditionally and the check could never be true in any state — and `run()` discarded the return value, so it never mattered. Now reads the TOC from `GET /books/{id}/toc` and asserts it.

## Verification

**Every new assertion was mutation-checked** — a test that cannot fail is exactly what this issue is about:

| Mutation | Result |
|---|---|
| Wire the Bold button to `toggleItalic` in `ChapterEditor.tsx` | Both Bold E2E tests RED; Heading/List stay green |
| Same mutation, jest layer | `the Bold button runs the toggleBold command` RED, other 6 green |
| Skip TOC generation in the simplified backend E2E | `AssertionError: Workflow completed but the book has no persisted table of contents` |

- **Frontend**: 128 suites, 2240 passed / 5 skipped. Coverage **rose** — statements 88.93 → 89.00, functions 85.20 → 85.56, lines 90.16 → 90.24, branches unchanged — confirming the deleted tests carried no product coverage. CI gate (85/85/75/85) green with headroom.
- **Backend**: 1198 passed / 11 skipped.
- **New E2E**: 4/4 passed against real backend + Mongo + chromium.
- `tsc --noEmit` clean; `eslint` 0 errors (the 403 pre-existing warnings are untouched and none are in the changed files).

## Review

Cross-family review by **opencode (GLM-5.2)** — no Critical or Major findings. It independently verified the four load-bearing assumptions (toolbar `title=` attributes, AI mock method-name match, `/toc` response shape, `BookResponse` lacking `table_of_contents`) and traced each assertion for vacuity, including the save-race question: the typing-triggered save produces `<p>text</p>`, which is not a substring of any polled fragment, so an intermediate un-formatted save cannot satisfy the poll.

Of 5 Minor findings, adopted 1: the standalone `'<ul>'` poll was weak (any list anywhere would satisfy it), now folded into the text-specific `<ul><li><p>text</p></li></ul>` — re-run and confirmed green rather than assumed.

## Known limitations

- The exact-substring fragments (`<h2>text</h2>`, `<ul><li><p>text</p></li></ul>`) couple to TipTap's serialization; a StarterKit bump could break them. They fail **loudly**, so this is maintenance cost, not a trust hole — and loosening them would weaken the one thing the spec exists to prove.
- The E2E relies on Playwright's accessible-name fallback to `title` (the live toolbar has no `aria-label`). Empirically confirmed by the passing run and the mutation going red for the right reason.

## Follow-up filed separately

`frontend/src/components/chapters/EditorToolbar.tsx` has **zero production importers** — the toolbar that actually ships is inlined in `ChapterEditor.tsx`. Its suite asserts `aria-label` on every icon-only button (#50, WCAG 2.1 AA) against a component no user ever sees, while the live inline toolbar has `title=` and **no `aria-label`**. Same hollow-test class as this issue plus a real a11y gap, but it needs its own change rather than doubling this PR.


frankbria:
Follow-up from the review scope note now filed as #375 (dead `EditorToolbar.tsx` + missing `aria-label` on the live toolbar).

Issue #351, opened 2026-07-23
[P3.1] Dead-code & scratch sweep across frontend and backend


frankbria:
**Phase:** P3.1 — hygiene (checklist-style; one focused session). No dependencies.

**Problem:** Multiple orphaned/scratch artifacts inflate the tree and mislead.
- **Frontend:** dead `userClient.ts` (zero importers, noted dead in #244) built on a misnamed `authFetch` placeholder that does no auth (`lib/utils.ts:12`); dead `TiptapDemo.tsx` (309 lines); test-only `responsiveHelpers.ts` under `src/lib`; wrong-stack `react-router-dom` devDep; no-op `useOptimizedClerkImage`.
- **Backend:** assertion-free `test_debug_questions*.py` (Clerk `clerk_id` fields); ~622 lines of skipped Clerk-era route tests (`test_password_change`/`test_email_verification`/`test_profile_picture`); dead `get_user_by_clerk_id` alias + `.get('clerk_id')` fallbacks; `.ps1` scratch scripts; scratch markdown in `backend/` root; uncollected `archived_tests/`.
- **Docs:** `AGENTS.md` names Vitest (actual: Jest) and `npm run type-check` (actual: `typecheck`); stale agent reports at repo root; auto-generated `CURRENT_SPRINT.md` that can't regenerate without gitignored `.beads`.

**Evidence:** `frontend/src/lib/api/userClient.ts:18`, `lib/utils.ts:12`, `components/chapters/TiptapDemo.tsx:1`; `backend/tests/test_debug_questions.py:1`, `tests/test_api/test_routes/test_password_change.py:7`, `db/database.py:36`; `AGENTS.md:19,60`.

**Impact:** Grep/diff noise, misleading dead auth/test surfaces, docs that misdirect agents.

**Acceptance criteria:**
- [ ] Delete the dead modules/tests/scripts/aliases (git preserves history); verify with zero-importer grep + `tsc --noEmit`/pytest collect.
- [ ] Fix `AGENTS.md` tool names.
- [ ] Move or gitignore auto-generated `CURRENT_SPRINT.md` and stale root reports.

frankbria:
Additional dead-code item found while implementing #337 (PR #368):

[...]

Issue #332, opened 2026-07-23
[P0.3] Password reset is advertised but sends no email in production (fake success)


frankbria:
**Phase:** P0.3 — launch blocker (account recovery silently broken). No dependencies.

**Problem:** Sign-in advertises password reset (`auth/sign-in/page.tsx:171` links `/auth/forgot-password`). The reset email is sent fire-and-forget and `forgot-password/page.tsx:58` unconditionally shows success. `sendPasswordResetEmail` only sends when `EMAIL_SERVICE_PROVIDER` is `resend`/`sendgrid`; unset in production it throws `Unknown email provider`, which the void IIFE swallows.

**Evidence:**
- `frontend/src/lib/auth.ts:144` (`void (async () => { try { ... } catch { console.error } })()`)
- `frontend/src/lib/email.ts:~215`
- `EMAIL_SERVICE_PROVIDER`/`EMAIL_SERVICE_API_KEY`/`EMAIL_FROM_ADDRESS` appear in neither `frontend/.env.example` nor the deploy workflows.

**Impact:** A user requesting account recovery sees a "check your email" screen but never receives an email — account recovery is silently broken for a SaaS.

**Acceptance criteria:**
- [ ] Document the three `EMAIL_*` vars in `.env.example`; set them as staging/prod secrets.
- [ ] Verify a real delivery on staging.
- [ ] Surface a soft error/log when the provider is unconfigured, so the dead-end is observable rather than a fake success.

Issue #384, opened 2026-07-29
[P3.17] VoiceTextInput dictation test is flaky under --coverage, red-lighting unrelated PRs


frankbria:
**Phase:** P3.17 — hygiene. No dependencies.

**Problem:** One test passes uninstrumented and fails under coverage, so it fails PRs at random regardless of what they change.

```
● VoiceTextInput Component › Continuous dictation (accumulation)
  › covers interim + multiple final results via the mock sequence
```

`.github/workflows/tests.yml` runs the suite twice — `Run unit tests`, then `Check coverage threshold` with `--coverage`. The test passes in the first and fails in the second, in the same job, on the same commit.

**Evidence:** PR #371 ([run 30185423875](https://github.com/frankbria/auto-author/actions/runs/30185423875)) changed only `frontend/scripts/package-lock.json` and `tasks/todo.md` — it cannot affect frontend behaviour — and still went red:

```
Run unit tests:            Test Suites: 130 passed, 130 total
                           Tests:       5 skipped, 2241 passed, 2246 total
Check coverage threshold:  FAIL src/__tests__/components/VoiceTextInput.test.tsx
                           Test Suites: 1 failed, 129 passed, 130 total
```

Coverage itself passed comfortably in that run — 88.91 stmts / 79.02 branch / 85.17 funcs / 90.14 lines against the 85/85/75/85 floor — so the gate failed purely on the flake. Re-running the same branch later passed all checks.

**Cause:** Istanbul rewrites every function to record hits, which inflates execution time enough to reorder `setTimeout`-vs-microtask scheduling. The test drives a mocked `SpeechRecognition` interim/final result sequence and asserts on ordering it does not control, so the instrumentation overhead changes the outcome.

**Impact:** Random red CI on unrelated PRs. Each occurrence costs a re-run and, worse, trains reviewers to treat a red `Frontend Tests` as noise — which is exactly how a real failure gets waved through.

**Acceptance criteria:**
- [ ] Make the mock `SpeechRecognition` sequence deterministic — drive results explicitly (fake timers or an awaited emit helper) rather than relying on scheduling order.
- [ ] Confirm by running `npm test -- --coverage src/__tests__/components/VoiceTextInput.test.tsx` repeatedly (≥20×) with zero failures.
- [ ] While in the file, check the sibling dictation tests for the same pattern.

_Found while clearing the Dependabot backlog (#360–#382); it was the sole cause of #371's red CI._

PR #386, opened 2026-07-29
docs(deploy): rewrite deployment guides to the real stack; drop Clerk-era scripts (#343)

### LINKED ISSUE #343: [P2.2] Deployment docs describe the wrong stack (Postgres/Clerk) + instruct baking an AWS secret into the client bundle
**Phase:** P2.2 — post-launch fast-follow. No dependencies.

**Problem:** Both deployment guides are migration-stale and would misconfigure a real deploy. Root `DEPLOYMENT.md` documents PostgreSQL + Clerk + Vercel/Docker (actual stack is MongoDB + better-auth + PM2/nginx) and instructs `NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY=` — any `NEXT_PUBLIC_` var is inlined into the client bundle, leaking IAM credentials to every browser. `.github/DEPLOYMENT.md` marks six Clerk secrets as Required and never mentions `BETTER_AUTH_SECRET`, which the backend hard-rejects at startup if unset. Two orphaned Clerk-era deploy scripts (`scripts/deploy.sh`, `deploy-fixed.sh`) would deploy a broken auth config if run.

**Evidence:** `DEPLOYMENT.md:44,217,223`; `.github/DEPLOYMENT.md:38-50,104-111`; actual workflow `deploy-staging.yml:234,250-252`; `scripts/deploy.sh:54`.

**Impact:** An operator provisioning from these docs configures the wrong datastore/auth (deploy health-check fails) or leaks an AWS secret into the client bundle.

**Acceptance criteria:**
- [ ] Rewrite both docs to the real stack (MongoDB, `BETTER_AUTH_SECRET`/`BETTER_AUTH_URL`, PM2/nginx, `uv sync`).
- [ ] Never store AWS secrets in `NEXT_PUBLIC_`.
- [ ] Delete or rewrite the orphaned Clerk deploy scripts.
- [ ] Cross-check the secrets list against `deploy-staging.yml`.

frankbria:
Closes #343.

## Problem

Both deployment guides were migration-stale and would misconfigure a real deploy:

- **`DEPLOYMENT.md`** documented PostgreSQL + Clerk + Vercel/Docker and an Alembic migration chain. The actual stack is MongoDB + better-auth + PM2/nginx, and the repo contains no `backend/alembic/`, no `Dockerfile`, and no Vercel config. It also instructed setting `NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY=` — Next.js inlines every `NEXT_PUBLIC_*` variable into the client bundle **at build time**, so following that line publishes an IAM secret to every visitor via View Source.
- **`.github/DEPLOYMENT.md`** marked six Clerk secrets as Required and never mentioned `BETTER_AUTH_SECRET`, which the backend hard-rejects at startup — a deploy configured from that document fails its own health check.
- **`scripts/deploy.sh` / `scripts/deploy-fixed.sh`** still wrote a Clerk env, including a hardcoded Clerk tenant URL.

## What changed

**Rewrote `DEPLOYMENT.md`** against the only live deploy path. MongoDB instead of PostgreSQL, better-auth instead of Clerk, `uv sync` instead of `uv pip install -r requirements.txt`, PM2 release-symlink deploy instead of Vercel/Docker, Python 3.13 per `pyproject.toml`, Alembic section removed. The `NEXT_PUBLIC_AWS_*` block is replaced by an explanation of *why* build-time inlining makes the prefix public by definition, pointing AWS credentials at the backend `.env` where the workflow already writes them.

**Rewrote `.github/DEPLOYMENT.md`** with the secrets table cross-checked line-by-line against `deploy-staging.yml` and `e2e-staging-tests.yml`. Adds `BETTER_AUTH_SECRET`, `SENTRY_DSN_BACKEND`/`SENTRY_DSN_FRONTEND`, `EMAIL_SERVICE_*` (without which password reset silently no-ops, #332), and `TEST_USER_EMAIL`/`TEST_USER_PASSWORD` — flagged as **`staging` environment** secrets, since a workflow without `environment: staging` receives empty strings. The rendered `.env` samples now match what the workflow writes, including that `BACKEND_CORS_ORIGINS` is comma-separated and **not** a JSON array (a quoted array gets mangled by the dotenv reader into invalid JSON and crashes `Settings()` at startup).

**Deleted both deploy scripts.** Verified safe: `deploy-staging.yml` inlines its own `bash -s` heredoc (L190–412) and never invokes them; no workflow, `package.json` script, pre-commit hook, or CI config references either.

**Added a mechanical guard.** `scripts/check-secrets.sh` — the existing pre-commit scanner — now rejects a secret assigned to a `NEXT_PUBLIC_*` name, so the AC "never store AWS secrets in `NEXT_PUBLIC_`" is enforced rather than merely documented.

**Bannered three operator-facing docs** that still instruct running the deleted scripts, including one targeting a decommissioned server IP.

## Demo — outcome evidence

The guard is the only executable change. Three cases, staged and run through `./scripts/check-secrets.sh`:

| Case | Input | Result |
|------|-------|--------|
| A | `NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY=wJalr…` (the exact line #343 reported) | **exit 1 — BLOCKED** |
| B | `NEXT_PUBLIC_STRIPE_SECRET_KEY: ${{ secrets.STRIPE }}` (yaml colon form) | **exit 1 — BLOCKED** |
| C | Prose: "Never set `NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY` in the frontend env." | exit 0 — allowed |

Case C matters: without the trailing `[=:]` requirement, the guard would reject the very documentation warning against the anti-pattern.

Repo-wide checks, all clean:
- No live reference to the deleted scripts in `.github/`, `scripts/`, `*.json`, or `.pre-commit-config.yaml`.
- No `NEXT_PUBLIC_*` secret assignment anywhere in the repo (including `archive/`).
- No Clerk / Postgres / Vercel / Alembic left in either rewritten guide except explicit negations.
- `bash -n scripts/check-secrets.sh` passes; the full staged changeset passes `check-secrets.sh`.

## Review

Independent pre-PR review by **opencode (GLM)** on the full `main...HEAD` diff, prompted to verify every documented secret, env var, and file path against `deploy-staging.yml`, `e2e-staging-tests.yml`, both `.env.example` files, `pyproject.toml`, and `tests.yml`.

**Critical: none. Major: none.** It independently confirmed all 9 deploy-blocking secrets are documented, that both rendered env-file templates reproduce the workflow's `echo` block verbatim, that the new regex is valid ERE with no false positives on existing repo content, and that deleting the scripts is safe.

Three Minors, two fixed in c38f9cb:
- `FRONTEND_URL` was described as doing "double duty" when it actually fans out to five consumers — corrected and each one named.
- The regex deliberately omits a bare `_KEY` suffix; that omission is now documented in the script, with the reason (matching it would false-positive on `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY` and similar intentionally-public keys).
- Not changed: the reviewer flagged "Staging uses MongoDB Atlas" as unverifiable from repo files. It is correct — staging runs against Atlas and its IP allowlist is a real operational trip-hazard when the server IP changes — it simply isn't derivable from committed code.

## Known limitations

- **`deploy-staging.yml:27` pins `PYTHON_VERSION: '3.9'`** while `pyproject.toml` requires `>=3.13`. Harmless today, because the runner never installs backend dependencies (they are installed server-side by `uv sync`), but it is misleading and worth its own issue. Not fixed here — this PR is scoped to documentation.
- **`deploy-staging.yml:26` uses Node 18** while `tests.yml` tests on Node 20. Same reasoning.
- **Eight further docs still reference the deleted scripts** — `STAGING_DEPLOYMENT_ANALYSIS.md`, `CI_CD_RECOVERY_PLAN.md`, `deployment-pipeline-summary.md`, `deployment-pipeline-diagram.md`, `production-readiness-gaps.md`, `PM2-DEPLOYMENT-FIXES.md`, `analysis/deployment-review.md`, `fixes/2025-10-24-cors-error-fix.md`. These are dated analyses and incident records rather than procedures, so they were left alone; the three that tell a reader to *run* something were bannered. A docs-archival sweep is a reasonable follow-up (and overlaps #351's dead-code sweep).


frankbria:
## Demo verification — every acceptance criterion mapped to outcome evidence

### AC1 — Rewrite both docs to the real stack (MongoDB, `BETTER_AUTH_SECRET`/`BETTER_AUTH_URL`, PM2/nginx, `uv sync`)

```
$ grep -inE "clerk|postgres|vercel|alembic|netlify|docker" DEPLOYMENT.md .github/DEPLOYMENT.md
DEPLOYMENT.md:32:There is no PostgreSQL, no Docker image, and no Alembic migration chain in this
DEPLOYMENT.md:199:Clerk, long after the migration to better-auth).
```

The only two survivors are explicit negations — a sentence telling the reader those things do **not** exist, and the explanation of why the deleted scripts were wrong. No instruction anywhere in either guide now references the old stack.

Independently confirmed by the pre-PR reviewer: both rendered env-file templates in `.github/DEPLOYMENT.md` reproduce the workflow's `echo` block **verbatim, including ordering**.

### AC2 — Never store AWS secrets in `NEXT_PUBLIC_`

Repo-wide, including `archive/`:

```
$ grep -rInE 'NEXT_PUBLIC_[A-Z0-9_]*(SECRET|PRIVATE_KEY|PASSWORD|CREDENTIALS?)[A-Z0-9_]*[[:space:]]*[=:]' . \
    --exclude-dir=node_modules --exclude-dir=.git --exclude-dir=.next
(no matches)
```

And enforced going forward, not merely documented — three cases staged and run through `./scripts/check-secrets.sh`:

| Case | Input | Result |
|------|-------|--------|
| A | `NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY=wJalr…` — the exact line #343 reported | **exit 1, BLOCKED** |
| B | `NEXT_PUBLIC_STRIPE_SECRET_KEY: ${{ secrets.STRIPE }}` — yaml colon form | **exit 1, BLOCKED** |
| C | Prose: "Never set `NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY` in the frontend env." | exit 0, allowed |

Case C is the reason the pattern requires a trailing `[=:]`: without it the guard would block the documentation that warns against the anti-pattern.

### AC3 — Delete or rewrite the orphaned Clerk deploy scripts

Deleted (`-438` lines). Safety verified by looking for a caller rather than assuming there is none:

```
$ grep -rn "deploy\.sh\|deploy-fixed" .github/ scripts/ *.json *.yaml *.yml .pre-commit-config.yaml
(no matches)
```

`deploy-staging.yml` inlines its own `bash -s` heredoc at L190–412 and never shells out to either script. Remaining mentions live only in `docs/`; the three that instruct a reader to *run* something are now bannered as superseded.

### AC4 — Cross-check the secrets list against `deploy-staging.yml`

Mechanical, both directions — every `secrets.X` in the two live workflows against the documented table:

```
26 secrets referenced by deploy-staging.yml + e2e-staging-tests.yml
26 documented in .github/DEPLOYMENT.md
 0 undocumented
 0 documented-but-unused
```

API_URL, AWS_ACCESS_KEY_ID, AWS_REGION, AWS_S3_BUCKET, AWS_SECRET_ACCESS_KEY, BETTER_AUTH_SECRET, BYPASS_AUTH, CLOUDINARY_API_KEY, CLOUDINARY_API_SECRET, CLOUDINARY_CLOUD_NAME, DATABASE_NAME, DISCORD_WEBHOOK_URL, EMAIL_FROM_ADDRESS, EMAIL_SERVICE_API_KEY, EMAIL_SERVICE_PROVIDER, FRONTEND_URL, HOST, MONGODB_URI, OPENAI_API_KEY, SENTRY_DSN_BACKEND, SENTRY_DSN_FRONTEND, SLACK_WEBHOOK_URL, SSH_KEY, TEST_USER_EMAIL, TEST_USER_PASSWORD, USER — all ✅.

### Other checks

- `bash -n scripts/check-secrets.sh` — passes.
- The full staged changeset passes `./scripts/check-secrets.sh` (one pre-existing false positive surfaced during development, from the scanner's `password.*"…"` pattern matching ordinary prose; resolved by rewording the sentence, not by weakening the scanner).

frankbria:
## Review round 2 — CodeRabbit triage (9 findings)

All 9 accepted and addressed in 2865c82. Two were worth more than the wording fix they looked like:

**Leaking secrets through troubleshooting output (major).** `.github/DEPLOYMENT.md` told an operator debugging a failed health check to `cat` the generated backend `.env`. That file holds the database URI, `BETTER_AUTH_SECRET`, the OpenAI key, and any AWS credentials — and troubleshooting output is exactly what gets pasted into tickets and screen shares. Replaced with names-only (`cut -d= -f1`) and a set/unset check that never prints a value. A PR about not leaking secrets shipping a "print all the secrets" command would have been an unfortunate way to close this issue.

**`kill -9` by port on a shared host (major).** I had written `lsof -ti:3002 | xargs -r kill -9` and then, directly beneath it, a warning that the VPS is shared and the port may not be ours. The warning doesn't help when the command is right there to copy. Now: identify the owner (`lsof -i:3002 -sTCP:LISTEN`, `pm2 list`), restart through PM2 if it's ours, leave it alone if it isn't.

**Scanner truncation (major).** `check-secrets.sh` only ever scanned the first 10KB of a file's added lines, so whether a leak was caught depended on where in the file it happened to land — a real weakness the new `NEXT_PUBLIC_` rule inherited. Raised to 100KB to match the scanner's own file-size guard. Verified against this changeset, which rewrites two documents well past the old cap: no new false positives, and the guard demo still passes all three cases.

**Runbooks: banner → archive.** CodeRabbit argued a banner still leaves an executable procedure sitting in `docs/`. Fair. Rather than deleting them, I moved `MANUAL_DEPLOYMENT_QUICKSTART.md` and `CI_CD_RECOVERY_QUICKSTART.md` to `archive/deployment/` — the repo's established home for read-only history per `CLAUDE.md`. History preserved, but they're no longer where an operator looks for a procedure. Relative links updated and resolution verified.

**The rest** — audit/dev-server blocks now use subshells or separate terminals so they work from the repo root; the CORS curl sends a real preflight; `docs/AUTOMATION_SETUP.md` now states the scanner rule's exact coverage and that `_KEY`/`_TOKEN` are excluded by design.

---

## ⚠️ `Frontend Tests` failed on the previous run — not from this PR

Run [30417605696](https://github.com/frankbria/auto-author/actions/runs/30417605696): 1 failed of 2245, `VoiceTextInput.test.tsx:196`, in the `Check coverage threshold` step. This PR changes no frontend code.

That is **#384** — but investigating it turned up something the issue gets wrong, so I'm fixing that first and will report there. In short: it is not only a test-timing artifact. `VoiceTextInput.tsx` mirrors `value` into a ref via `useEffect`, and the effect cannot run between two `onresult` events that arrive in the same tick — so the second read sees the stale value and the first finalized segment is discarded. That is the #330 bug at a shorter timescale, and it is reachable by real dictation, not just by Istanbul's instrumentation overhead. Details and a failing test to follow on #384.

Merging this PR is gated on that going green.

Issue #349, opened 2026-07-23
[P2.8] Live editor toolbar and several forms lack accessible state/labels (the #50 a11y fix landed on dead code)


frankbria:
**Phase:** P2.8 — post-launch fast-follow. No dependencies.

**Problem:** The toolbar users actually see is inlined in ChapterEditor with only `title=` on icon-only toggles — no `aria-pressed`, no `role="toolbar"`. The #50 fix (aria-label/aria-pressed/role) was applied to `EditorToolbar.tsx`, which has zero importers (dead code). Related a11y gaps: the clarifying-questions answer textarea is placeholder-only (no label/`aria-labelledby`), summary submit/validation errors aren't `role=alert` and the textarea lacks `aria-invalid`/`aria-describedby`, and QuestionDisplay/ChapterEditor save-status/error aren't in live regions; a few muted labels fail 4.5:1 contrast.

**Evidence:** `frontend/src/components/chapters/ChapterEditor.tsx:531,482`; `frontend/src/components/toc/ClarifyingQuestions.tsx:206,167`; `frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:162,196,205`; `frontend/src/components/chapters/questions/QuestionDisplay.tsx:530,579`.

**Impact:** Screen-reader users can't perceive formatting state, field prompts, autosave results, or errors (WCAG 4.1.2/1.3.1/1.4.3).

**Acceptance criteria:**
- [ ] Render the fixed EditorToolbar (deleting the inline copy) or add `role=toolbar`+`aria-pressed`.
- [ ] Label the question textarea via `aria-labelledby`.
- [ ] Add `role=alert`/`aria-live`/`aria-invalid`/`aria-describedby` to the summary and save-status regions.
- [ ] Bump `text-gray-500`-on-`gray-800` to ≥4.5:1; retarget the a11y test at the rendered toolbar.

PR #391, opened 2026-07-29
perf(editor): memoize the toolbar, code-split TipTap, drop the unused react-query dep (#347)

### LINKED ISSUE #347: [P2.6] Editor runtime perf: full re-render per keystroke, no request cache, editor not code-split
**Phase:** P2.6 — post-launch fast-follow. No dependencies.

**Problem:** Three compounding client costs in the core editor. (1) The 836-line ChapterEditor subscribes to the editor instance directly, so all ~15 toolbar buttons + 4 AI-dialog triggers re-render on every ProseMirror transaction. (2) There is no client request cache — `@tanstack/react-query` is a declared dep but referenced only by a jest mock, so every chapter-tab switch re-fetches full chapter HTML. (3) TipTap/ProseMirror is statically imported into the book-page route, not code-split.

**Evidence:** `frontend/src/components/chapters/ChapterEditor.tsx:152,531-726`; unused `frontend/src/lib/react-query.ts` (only importer); `frontend/src/components/chapters/TabContent.tsx:5`.

**Impact:** Character-by-character reconciliation on long chapters, repeated full-HTML refetch on every tab switch, and a large eager editor bundle on route entry.

**Acceptance criteria:**
- [ ] Extract a memoized toolbar reading state via `useEditorState({editor, selector})`.
- [ ] Adopt react-query (keyed by bookId+chapterId) for `getChapterContent`/`getUserBooks`, or remove the unused dep.
- [ ] `next/dynamic({ssr:false})` the editor in TabContent.

frankbria:
Closes #347.

## 1. Toolbar re-rendered on every keystroke

`useEditor` re-renders its host on every ProseMirror transaction, and the fourteen formatting buttons read `editor.isActive()` inline — so all of them plus their icon components reconciled per character typed.

**The fix reuses code that was already in the repo.** `EditorToolbar.tsx` existed, complete with accessibility tests from #50, imported by nothing but its own test. That is the dead code #349 refers to ("the #50 a11y fix landed on dead code"). Rather than write a second toolbar, this wires the existing one in and converts it to `useEditorState({editor, selector})` + `memo`, deleting the 196-line inline duplicate from `ChapterEditor`.

The subscription is what makes `memo` safe — with `memo` alone the active states, and therefore `aria-pressed`, would silently go stale. Two tests pin both halves.

**Side effect worth naming:** wiring it up lands #50's accessibility work on live code for the first time — `role="toolbar"`, per-button `aria-label`, `aria-pressed`, and 44px touch targets (WCAG 2.5.5) where the inline copy had 32px and no ARIA. **The formatting buttons are visibly larger as a result.** That was #50's intent; flagging it because it is a visual change, not a silent one. #349 remains open for the "several forms" half.

## 2. TipTap was not code-split

Both routes hosting an editor imported it statically. One shared `next/dynamic` boundary (`LazyChapterEditor`) now covers both rather than duplicating the loading state.

**First Load JS, measured with `next build` on `main` and on this branch:**

| Route | `main` | This branch | Change |
|---|---|---|---|
| `/dashboard/books/[bookId]` | 497 kB | **374 kB** | **−123 kB (−25%)** |
| `/dashboard/books/[bookId]/chapters/[chapterId]` | 413 kB | **206 kB** | **−207 kB (−50%)** |

The second route is the deprecated redirect page, which was paying for the entire editor bundle in order to redirect away two seconds later.

## 3. react-query removed

It was a declared dependency with **zero** runtime usage; the only `src/lib/react-query.ts` was a jest mock — `jest.fn()` sitting in production source — imported by nobody. Per the AC's explicit either/or, and confirmed with the repo owner, the dep and that file are deleted rather than adopted.

Chapter-tab switches still refetch full HTML; that is the acknowledged cost of this choice and is noted as follow-up. Putting a read cache in front of an editor that already has autosave plus a localStorage fallback needs invalidation-after-save designed deliberately, not bolted onto a perf pass.

---

## A regression I shipped, and the hollow test that missed it

Review caught that the undo/redo buttons read `editor.can().chain().focus().undo().run()` inline for `disabled`. Once the toolbar is memoized it re-renders only when the selector's output changes — and that capability read was not in the selector. So typing makes undo possible, nothing in the selector changes, no re-render happens, and **the button stays greyed out permanently.** Undo unreachable from the toolbar. Fixed in 3c2f45d; this is precisely the trap my own comment warned about two lines above the bug.

The more interesting part: **my first regression test passed against the buggy code.** My `useEditorState` mock in `jest.setup.ts` re-rendered on every transaction, whereas the real hook runs on `useSyncExternalStoreWithSelector` and bails out when the new selector output equals the old. Under the permissive mock the inline read got re-evaluated too, so the test proved nothing — and it was hiding exactly the class of bug memoization introduces.

The mock now reproduces the bail-out semantics. Mutation check with it corrected:

```
bug reintroduced  → ✕ keeps undo/redo enablement live across transactions (#347)
fix restored      → ✓ 21 passed
```

Before the mock fix, that same mutation passed. Worth stating plainly: without fixing the mock, this PR would have shipped a broken undo button behind a green test.

## Declined finding, with evidence

CodeRabbit also asked that `ChapterEditor` be removed from the redirect page entirely rather than lazily loaded — reasonable on its face, since the page's job is to redirect. **Declining: it would delete live E2E coverage.** `src/e2e/style-transformation.spec.ts:76` navigates to that exact route and drives the full editor there:

```ts
await page.goto(`/dashboard/books/${BOOK_ID}/chapters/${CHAPTER_ID}`);
await expect(page.locator('.tiptap')).toContainText(ORIGINAL);
await page.getByRole('button', { name: /transform style/i }).click();
```

`draft-generation.spec.ts` and `responsive.spec.ts` also target it. Those specs leaning on a deprecated route is its own smell, but removing the component would remove the coverage — a separate decision from a perf pass.

## Verification

| Check | Result |
|-------|--------|
| Frontend suite (coverage) | **128/128 suites, 2244 passed**, 5 skipped, thresholds met |
| `tsc --noEmit` | clean |
| `npm run lint` | 0 errors (416 pre-existing warnings) |
| `next build` | succeeds; bundle deltas above |
| Mutation check (undo/redo) | fails with bug, passes with fix ✅ |
| Mutation check (parent re-render) | fails without `memo` ✅ |
| CodeRabbit | 2 findings — 1 fixed, 1 declined with evidence |

`RichTextEditor.test.tsx`'s ten button→command tests pass against the new toolbar **unchanged**, which is the evidence the swap preserved behaviour. Mock surfaces gained `on`/`off` because the component now subscribes; no assertion was altered.


frankbria:
## CI caught a miss my own grep created

`Frontend Tests` failed on the first run:

```
FAIL src/__tests__/ChapterQuestionsEndToEnd.test.tsx
  ● Test suite failed to run
    Cannot find module '@tanstack/react-query' from 'src/__tests__/ChapterQuestionsEndToEnd.test.tsx'
```

Four suites import it — `ChapterQuestionsEndToEnd`, `ChapterQuestionsMobileAccessibility`, `ChapterQuestionsEdgeCases`, `ChapterQuestionsPerformance`.

I claimed "zero runtime usage" on the strength of a grep that **excluded `__tests__`**, and it passed locally because `node_modules` still held the package after I removed it from `package.json`. Two mistakes stacked: a search scoped to exclude the very files that used it, and a verification run against a stale dependency tree. CI's clean `npm ci` found both immediately. Re-verified under `npm ci` this time.

The substance holds, though — every one of the four used it identically: a `QueryClientProvider` wrapping components that never call a react-query hook. Ceremony from the adoption that never happened, which is the same reason the dependency is going. Wrappers removed, `queryClient.clear()` in `afterEach` dropped or folded into the `jest.clearAllMocks()` it was already sitting next to. **No assertion changed.**

```
128/128 suites, 2244 passed, 5 skipped — against a clean npm ci
tsc --noEmit clean
```

`review / review` also passed on the previous commit. Re-running the full gate now.

PR #392, opened 2026-07-29
fix(voice): release the mic on unmount, stop advertising unsupported dictation, humanize errors (#348)

### LINKED ISSUE #348: [P2.7] Voice input UX/privacy gaps: mic left live on unmount, unsupported-browser button, raw error codes, missing a11y
**Phase:** P2.7 — post-launch fast-follow. No dependencies. (Complements the P0.1 data-loss fix.)

**Problem:** Several defects across both SpeechRecognition surfaces. (1) No unmount cleanup calls `recognition.stop()`, so navigating away while recording leaves the mic live. (2) The summary-page "🎤 Voice Input" button renders enabled to unsupported browsers (Firefox/Safari/insecure context) and only errors on click, unlike VoiceTextInput which tracks `isSupported`. (3) Routine events (`no-speech`, `aborted`) surface as scary raw red errors; denied permission gives no guidance. (4) The summary voice surface has no live-region announcements and discards interim results.

**Evidence:** `frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:57-60,77,103,174,181`; `frontend/src/components/chapters/VoiceTextInput.tsx:143`.

**Impact:** Privacy (stuck-live mic), false capability advertisement, confusing errors that make working dictation look broken, and voice-only affordances inaccessible to screen readers.

**Acceptance criteria:**
- [ ] Add unmount `stop()` in both surfaces.
- [ ] Detect support on mount and hide/disable the voice button with guidance.
- [ ] Map/suppress benign error codes and give actionable copy for `not-allowed`.
- [ ] Add `aria-live` status + `role=alert` + interim transcript to the summary surface (mirror VoiceTextInput).

frankbria:
Closes #348.

## 1. The stuck-live microphone — worse than the issue described

The issue reports that no unmount cleanup calls `recognition.stop()`, so navigating away mid-dictation leaves the mic live. True, and fixed. But review surfaced a second source that my first pass missed, and it is the more serious one:

```ts
// before
await navigator.mediaDevices.getUserMedia({ audio: true });   // result discarded
```

That `MediaStream`'s audio tracks are what actually hold the microphone open, and **`recognition.stop()` does not touch them.** The stream was requested and dropped on the floor, so the mic stayed live for the life of the page — *including after the user pressed Stop*. The recording indicator stayed lit while they sat there believing dictation had ended.

Both surfaces now stop recognition on unmount (reading the ref, not state, so they cannot capture a stale recorder; handlers detached first because `stop()` fires `onend`, which would otherwise `setState` after unmount), and `VoiceTextInput` holds the stream in a ref and stops its tracks on **both** paths — explicit stop and unmount.

## 2. False capability advertisement

The summary page's "🎤 Voice Input" button rendered enabled in Firefox/Safari and over plain HTTP, failing only once clicked. It is now gated on a probe that requires `navigator.mediaDevices.getUserMedia` as well as the constructor — **Chrome over plain HTTP exposes `SpeechRecognition` but cannot reach the mic**, so a constructor check alone still advertises something that does not work. Unsupported browsers get guidance instead of a dead button.

## 3. Raw error codes

Both surfaces rendered the API's code straight through — `Error recording audio: no-speech`. `no-speech` fires whenever a listening window passes without audio, and `aborted` whenever the user stops or navigates, so **routine use painted red banners**, which is how you train people to ignore the error region entirely. Those two are now suppressed; everything else maps to copy that says what to do, with `not-allowed` explaining where to grant permission.

Shared as `src/lib/voice/speechRecognitionErrors.ts` rather than duplicated — the two components have very different shapes but identical error semantics.

## 4. Summary surface accessibility

It had no live region at all: a voice-only affordance a screen-reader user could not monitor. Added an `aria-live` status, `role="alert"` on the error banner, `aria-pressed` on the toggle, and display of interim results — which were already being computed and then discarded, so the page looked frozen while listening. Mirrors what `VoiceTextInput` already did.

## Tests

Assertions target the device, not the state: `track.stop()` on a mocked stream, because the track is the thing holding the mic. Asserting `isRecording === false` would have passed throughout the entire bug.

**Mutation checks:**

| Mutation | Result |
|---|---|
| Delete the unmount effect | ✕ both mic-release tests fail |
| Remove track release from the stop path only | ✕ stop test fails, unmount test still passes |

The second one matters: it shows the two release paths are independently guarded rather than one covering for the other.

Plus 19 tests on the shared error module, including that no mapped message ever contains the raw code it came from.

## One pre-existing test updated

`VoiceTextInput.test.tsx` asserted the old copy verbatim — `'Error recording audio: network'` — which is the behaviour this issue exists to remove. Updated to the new intent, and strengthened: it now also asserts the raw code is absent. The test's purpose (a real failure still surfaces to the user) is unchanged.

## Verification

| Check | Result |
|-------|--------|
| Full frontend suite (coverage) | **130/130 suites, 2271 passed**, thresholds met |
| `tsc --noEmit` | clean |
| `npm run lint` | 0 errors |
| CodeRabbit | 2 findings, both fixed — one was the `getUserMedia` leak above |

### DIFF
--- a/frontend/src/__tests__/components/VoiceTextInput.test.tsx
+++ b/frontend/src/__tests__/components/VoiceTextInput.test.tsx
@@ -135,10 +135,16 @@ describe('VoiceTextInput Component', () => {
       // Wait for recognition to start
       expect(mockRecognition.start).toHaveBeenCalled();
 
-      // Wait for error to appear
+      // A real failure still surfaces — but as actionable copy, not the raw
+      // code. #348 replaced "Error recording audio: network" because echoing the
+      // API's code tells the user nothing about what to do. getAllByText because
+      // the message renders in both the banner and the sr-only live region.
       await waitFor(() => {
-        expect(screen.getByText('Error recording audio: network')).toBeInTheDocument();
+        expect(
+          screen.getAllByText(/needs an internet connection/i).length
+        ).toBeGreaterThan(0);
       }, { timeout: 2000 });
+      expect(screen.queryByText(/Error recording audio/i)).not.toBeInTheDocument();
     });
   });
 
--- a/frontend/src/app/dashboard/books/[bookId]/summary/page.tsx
+++ b/frontend/src/app/dashboard/books/[bookId]/summary/page.tsx
@@ -11,6 +11,11 @@ import {
   countSummaryCharacters,
   getSummaryReadinessError,
 } from '@/lib/constants/summary-readiness';
+import {
+  describeSpeechError,
+  isSpeechRecognitionSupported,
+  SPEECH_UNSUPPORTED_MESSAGE,
+} from '@/lib/voice/speechRecognitionErrors';
 
 export default function BookSummaryPage() {
   const router = useRouter();
@@ -20,6 +25,12 @@ export default function BookSummaryPage() {
   const [isLoading, setIsLoading] = useState(false);
   const [isListening, setIsListening] = useState(false);
   const [error, setError] = useState('');
+  // Advertised capability must match reality: this button used to render enabled
+  // in Firefox/Safari and over plain HTTP, and only failed on click (#348).
+  const [voiceSupported, setVoiceSupported] = useState(false);
+  // Interim results were thrown away, so nothing showed until a phrase
+  // finalised and the surface looked frozen while listening.
+  const [interimTranscript, setInterimTranscript] = useState('');
   const saveTimeout = useRef<NodeJS.Timeout | null>(null);
   const lastSaved = useRef('');
   const [summaryHistory, setSummaryHistory] = useState<unknown[]>([]);
@@ -76,11 +87,35 @@ export default function BookSummaryPage() {
   // Speech recognition setup
   const recognitionRef = useRef<SpeechRecognition | null>(null);
 
+  useEffect(() => {
+    setVoiceSupported(isSpeechRecognitionSupported());
+  }, []);
+
+  // Release the microphone if this page unmounts mid-dictation (#348).
+  // Handlers are detached first so stop()'s onend cannot setState after unmount.
+  useEffect(() => {
+    return () => {
+      const recognition = recognitionRef.current;
+      if (!recognition) {
+        return;
+      }
+      recognition.onresult = null;
+      recognition.onerror = null;
+      recognition.onend = null;
+      try {
+        recognition.stop();
+      } catch {
+        // Already stopped — nothing to release.
+      }
+      recognitionRef.current = null;
+    };
+  }, []);
+
   const startListening = () => {
     const SpeechRecognitionCtor =
       (window.SpeechRecognition || window.webkitSpeechRecognition) as typeof SpeechRecognition | undefined;
-    if (!SpeechRecognitionCtor) {
-      setError('Speech recognition is not supported in your browser.');
+    if (!voiceSupported || !SpeechRecognitionCtor) {
+      setError(SPEECH_UNSUPPORTED_MESSAGE);
       return;
     }
     setError('');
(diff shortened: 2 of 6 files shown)


frankbria:
## Third round on the same defect — reviewer was right again

GLM flagged that `onerror` ends the session without releasing the `MediaStream`. Correct, and auditing every session-ending path found I had fixed **two of four**:

| Path | Before this commit |
|---|---|
| user presses Stop | ✅ released (previous commit) |
| unmount | ✅ released (previous commit) |
| `onerror` — network/permission failure | ❌ **leaked** |
| `onend` — service timeout / end of utterance | ❌ **leaked** |

The two I missed are the ones the *browser* drives rather than the user, and they're arguably the worse ones: `isRecording` goes false, so the Stop button disappears — leaving a lit recording indicator and **no control that would turn it off**.

Also fixed a leak on retry: `startRecording` overwrote `micStreamRef` without releasing what was already there, orphaning the previous stream's tracks for the page's lifetime with no handle left to stop them.

`releaseMicStream` is now a stable `useCallback` hoisted above the recognition factory, so all four paths call one function, and `startRecording` releases before acquiring. That's what it should have been initially — my previous commit was two patches at two of four sites, which is precisely how the third and fourth got missed.

**Mutation check, per path:**

```
release removed from onerror → ✕ releases the stream when recognition errors
release removed from onend   → ✕ releases the stream when recognition ends on its own
restored                     → ✓ 11 passed
```

Each fails only its own test, so no test is covering for another.

```
130/130 suites, 2274 passed, 5 skipped
tsc --noEmit clean
```

Worth recording how this went: the issue described one symptom (no unmount `stop()`), and the actual defect was a `MediaStream` nobody ever disposed, across four exit paths. Three review rounds to bottom out. The lesson I'll carry: when the fix is "release a resource," enumerate the exits *first* and assert on the resource, not on the state flag that merely accompanies it.

frankbria:
## A regression I introduced, caught before merge

GLM flagged that the re-entrancy guard I added in `df10392` checks `recognitionRef.current`, which **neither `onend` nor `onerror` ever clears**. Verified — only explicit stop and unmount cleared it.

`onend` fires routinely: a pause, a service timeout, end of utterance. So after the first such end the Start button returned while every click was a silent no-op. **Dictation dead until the component unmounted**, with no error to explain why. That is a worse outcome than the leak the guard was added to fix — I'd have shipped a broken feature to close a privacy bug.

The self-critique worth recording: a guard is only as good as the invariant it reads. I added a check on `recognitionRef` without auditing who maintains it — the same mistake as the original defect in different costume. The original was "assumed the mic was released"; this was "assumed the handle was current". Both are trusting state I hadn't traced every writer of.

Both handlers now clear the handle alongside the stream they already released.

**Tests** drive a session to `onend` and to `onerror`, then assert a subsequent Start **actually calls `getUserMedia` again**. Asserting the button is merely present would have passed throughout the bug — it was visible and inert, which is the whole problem.

```
handle no longer cleared → ✕ allows a new recording after onend ends the session
                           ✕ allows a new recording after onerror ends the session
restored                 → ✓ 15 passed
```

```
130/130 suites, 2278 passed, 5 skipped
tsc --noEmit clean
```

---

### Running tally on this one resource

| # | Escape route | Found by |
|---|---|---|
| 1 | no unmount `stop()` | the issue |
| 2 | `getUserMedia` stream never disposed at all | review |
| 3 | `onerror` ends session, keeps stream | review |
| 4 | `onend` ends session, keeps stream | me, auditing #3 |
| 5 | re-entrant start orphans the first stream | review |
| 6 | unmount during the `await` | me, fixing #5 |
| 7 | guard bricks restart (my regression) | review |

Five review rounds. The issue described one symptom; the defect was a resource whose ownership was spread across seven paths with no single acquire/release pair. It now has one.

frankbria:
## StrictMode would have broken dictation in every dev session

GLM caught that `unmountedRef` is set `true` in cleanup and **never reset**. React StrictMode double-invokes effects in development — mount → cleanup → mount — which is Next.js's default for `next dev`. So the flag latched `true` on that first teardown, and `startRecording`'s guard then aborted every recording. Dictation dead for the whole dev session, silently.

Fixed by resetting at the top of the effect, which is the correct place: the component is mounted there by definition.

**The mutation check is the interesting part:**

```
reset removed → ✕ still records after StrictMode double-invokes the effects
                ✓ 16 others pass
```

Sixteen tests covering unmount, error, natural end, restart, re-entrancy, failed start — none could see it. Only a test that wraps the component in `StrictMode` reproduces the double-invoke, because that *is* the defect's trigger.

What this says about my own guard, added two commits earlier: it closed a real leak, but "have I unmounted?" tracked in a write-once ref is only correct if mount happens once — and in development it doesn't. That's now the third time in this PR a fix of mine introduced state whose maintenance I hadn't fully traced. The consistent lesson: **every new piece of state needs its full write-set audited, not just the write that motivated it.**

```
130/130 suites, 2280 passed, 5 skipped
tsc --noEmit clean
```

### Full tally for this one resource

| # | Escape route / defect | Found by |
|---|---|---|
| 1 | no unmount `stop()` | the issue |
| 2 | `getUserMedia` stream never disposed at all | review |
| 3 | `onerror` keeps the stream | review |
| 4 | `onend` keeps the stream | me, auditing #3 |
| 5 | re-entrant start orphans the first stream | review |
| 6 | unmount during the `await` | me, fixing #5 |
| 7 | guard bricks restart (my regression) | review |
| 8 | `start()` throw leaves a dead handle | review |
| 9 | StrictMode latches the unmount flag (my regression) | review |

Six review rounds. The issue described one symptom; the real defect was a resource with no single owner, plus three regressions I introduced while giving it one.

PR #393, opened 2026-07-30
fix(a11y): label the question field, announce save status, fix dark-mode contrast (#349)

### LINKED ISSUE #349: [P2.8] Live editor toolbar and several forms lack accessible state/labels (the #50 a11y fix landed on dead code)

frankbria:
Closes #349.

## Scope note: two of the four ACs were already delivered

- **AC1 (render the fixed EditorToolbar, delete the inline copy)** — done in **#347**. That PR found `EditorToolbar.tsx` sitting unreferenced with #50's accessibility tests attached, and wired it in. Landing it is what put `role="toolbar"`, per-button `aria-label`/`aria-pressed`, and 44px touch targets on live code for the first time.
- **AC3, in part** — done in **#348**: the summary error banner is `role="alert"` and dictation status is a live region.

What remained is below.

## Labelling

The clarifying-questions answer field had only a placeholder. **Placeholder text is not an accessible name** — it disappears the moment the user types, and several screen readers skip it — so the field announced as unlabelled and the user couldn't tell which question they were answering. Now labelled by the question heading via `aria-labelledby`.

## Save status was purely visual

`QuestionDisplay`'s autosave indicator and `ChapterEditor`'s footer status are now announced. A save *failure* is `role="alert"` — a possibly-unsaved answer is worth interrupting for — while a routine "Saved" during typing is polite.

The summary textarea now carries `aria-invalid` and `aria-describedby` pointing at both its readiness error and its help text; the error was previously visible text with nothing associating it to the field. Polite rather than assertive, since it re-evaluates on **every keystroke** and an alert would interrupt continuously.

## Contrast

`text-gray-500` on `gray-800` is ~2.7:1. Raised to `gray-400` (~5:1) where it's body text on a dark surface — 8 occurrences across the TOC components and `BookCreationWizard`'s dark-mode override.

**Deliberately not changed**, after checking each of the 26 occurrences rather than sweeping:
- **Disabled controls** (`disabled:text-gray-500` ×2) — WCAG 1.4.3 explicitly exempts inactive components, and the dimming *is* the affordance.
- **Auth pages** — already carry `dark:text-gray-400` overrides, so they're fine in the mode that matters.
- **Icons** — non-text, governed by the 3:1 rule, and all accompanied by text labels.
- **Generic theme-token components** (`text-muted-foreground` et al) — need a theme-wide audit, not a spot fix.

A blanket find-and-replace would have been wrong in at least the first two categories.

## Review round: live regions that would never have announced

CodeRabbit caught that my live regions were **mounted together with their content**. That's the classic way to get silence — a screen reader must already be observing a `role="status"` node to notice it mutate, so inserting region and text in the same commit frequently announces nothing at all. The feature I added would have been decorative.

Both surfaces now keep one always-mounted `sr-only` announcer, with the visual indicators `aria-hidden` so the status isn't read twice. `role="alert"` is left inserted-on-demand on the failure branch, because for alerts insertion *is* the trigger.

The announcer wording is intentionally distinct from the visual labels ("Your response could not be stored" vs "Save failed") — that keeps existing `getByText` queries unambiguous, and full sentences read better than terse visual chips.

`role="alert"` also moved off the error *container* onto the message itself: the container holds the Retry button, whose label was being announced as part of the alert.

## A correction to my own change

My first commit promoted the question prompt `h3` → `h2`, claiming the wizard's `h1` was followed directly by an `h3`. **That was wrong.** The component renders its own `h2` section header, so the question belongs *under* it at `h3` — my change made it a sibling of its own section.

The real violation was only the `h4` "Question Overview", which renders in a branch where the question `h3` does not, so it followed the `h2` directly. The `h4` → `h3` fix was sufficient; the `h2` promotion is reverted. Settled by running axe both ways rather than reasoning about the markup.

## Verification

| Check | Result |
|-------|--------|
| Full frontend suite (coverage) | **131/131 suites, 2282 passed**, thresholds met |
| `tsc --noEmit` | clean |
| `npm run lint` | 0 errors |
| axe on ClarifyingQuestions | no violations |
| Mutation check | removing `aria-labelledby` fails both new tests, including axe |

### DIFF
--- a/frontend/src/app/dashboard/books/[bookId]/summary/page.tsx
+++ b/frontend/src/app/dashboard/books/[bookId]/summary/page.tsx
@@ -255,6 +255,10 @@ export default function BookSummaryPage() {
                 )}
               </div>
             </div>
+            {/* aria-invalid tells assistive tech the field is in an error
+                state, and aria-describedby points at the message explaining why
+                — previously the readiness error was visible text with nothing
+                associating it to the field (#349). */}
             <textarea
               id="summary"
               value={summary}
@@ -263,6 +267,8 @@ export default function BookSummaryPage() {
               className="w-full bg-background border border-border rounded-md py-2 px-3 text-foreground"
               placeholder="Describe your book's main concepts, structure, and key points that should be organized into chapters..."
               required
+              aria-invalid={Boolean(inputError)}
+              aria-describedby={inputError ? 'summary-input-error summary-help' : 'summary-help'}
             ></textarea>
 
             {/* Dictation status, announced politely. A voice-only affordance with
@@ -294,7 +300,17 @@ export default function BookSummaryPage() {
               <span>{countSummaryCharacters(summary)} / {SUMMARY_MIN_CHARACTERS} characters</span>
             </div>
             {inputError && (
-              <div className="text-red-400 text-xs mt-1">{inputError}</div>
+              // Polite, not alert: this fires on every keystroke while the
+              // summary is below the minimum, and an assertive region would
+              // interrupt the user continuously as they type.
+              <div
+                id="summary-input-error"
+                role="status"
+                aria-live="polite"
+                className="text-red-400 text-xs mt-1"
+              >
+                {inputError}
+              </div>
             )}
             <div id="summary-help" className="text-xs text-muted-foreground mt-2">
               <div>Guidelines: Aim for 1-3 paragraphs. Include the main idea, genre, and any key themes or characters. At least {SUMMARY_MIN_WORDS} words and {SUMMARY_MIN_CHARACTERS} characters are required to generate a table of contents — the more detail you give, the better the result.</div>
--- a/frontend/src/components/BookCreationWizard.tsx
+++ b/frontend/src/components/BookCreationWizard.tsx
@@ -150,7 +150,7 @@ export function BookCreationWizard({ isOpen, onOpenChange, onSuccess }: BookCrea
                       disabled={isSubmitting}
                     />
                   </FormControl>
-                  <FormDescription className="text-muted-foreground dark:text-gray-500">
+                  <FormDescription className="text-muted-foreground dark:text-gray-400">
                     Provide a short summary or description of your book project.
                   </FormDescription>
                   <FormMessage className="text-destructive dark:text-red-400" />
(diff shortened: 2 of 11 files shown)


frankbria:
### Note: `security/snyk` failure is an account quota, not a finding

```
security/snyk (frankbria)   fail   You have used your limit of private tests
```

That is Snyk's plan limit being exhausted, not a vulnerability detected in this diff. Nothing in this PR touches dependencies — it changes ARIA attributes, Tailwind text colours, and one heading level.

The repo's own supply-chain gate (`Security Audit`, which runs `pip-audit` + `npm audit` against `security-baseline.json`) is unaffected and reports separately.

I can't clear the quota from here, and I won't merge with `--admin`. If Snyk turns out to be a required check, it needs an account-side reset before this can land.

PR #399, opened 2026-07-30
build(deps): move the whole @tiptap family to v3 — closes #385

### LINKED ISSUE #385: [P3.18] Two frontend major dep bumps are blocked on coordinated migrations (tiptap v3, zod 4)
**Phase:** P3.18 — dependency maintenance. No dependencies.

**Problem:** Dependabot keeps proposing two major bumps that cannot merge as single-package changes. Both were reviewed and closed (#381, #380) with the analysis below; Dependabot will reopen them on each new upstream release, so this issue exists to stop the analysis being re-derived every time.

---

### 1. `@tiptap/*` → v3 (closed as #381)

v3 unbundles the extensions from `starter-kit`, so every command disappears from the chain, and a duplicated `@tiptap/core` breaks type identity:

```
ChapterEditor.tsx(536,50): error TS2551: Property "toggleBold" does not exist on type "ChainedCommands".
ChapterEditor.tsx(154,7):  error TS2769: No overload matches this call.
  Type "@tiptap/core/dist/Mark".Mark<any, any> is not assignable to type
       "@tiptap/starter-kit/node_modules/@tiptap/core/dist/index".Mark<any, any>
```

Also affected: `toggleItalic`, `toggleStrike`, `toggleHeading`, `toggleBulletList`, `toggleOrderedList`, `toggleBlockquote`, `toggleCodeBlock`, `undo`, `redo`, `setHorizontalRule` — in both `ChapterEditor.tsx` and `EditorToolbar.tsx`.

Root cause: `frontend/package.json` pins ~10 siblings at `^2.x` (`@tiptap/react` plus `extension-{blockquote,bullet-list,character-count,heading,link,list-item,ordered-list,placeholder,strike,underline}`). Bumping `starter-kit` alone cannot work.

**Acceptance criteria:**
- [ ] Move the entire `@tiptap/*` family to v3 in one PR.
- [ ] Register the now-unbundled extensions explicitly in `ChapterEditor.tsx` / `EditorToolbar.tsx`.
- [ ] Verify `npm run typecheck` is clean and the editor E2E passes (this surface is what #341 added real browser coverage for).

---

### 2. `zod` → v4 (closed as #380)

`@hookform/resolvers` is pinned `^3.10.0`, and its v3 zod resolver only speaks zod 3. Under zod 4 it *throws* out of the resolver instead of returning field errors, so validation breaks at runtime rather than rendering inline messages:

```
● BookCreationWizard › Validation Tests › validates required title field
  ZodError: [ { "code": "custom", "path": ["title"], "message": "Title is required" } ]
    at values (node_modules/@hookform/resolvers/zod/src/zod.ts:71:9)
```

Five `BookCreationWizard` tests fail this way. Classic peer-range-satisfied-but-runtime-incompatible: npm resolves the install cleanly and the resolver crashes when invoked.

**Acceptance criteria:**
- [ ] Bump `@hookform/resolvers` to >=5 (routes zod 4 through a separate entry point) in the same PR as zod 4.
- [ ] Audit every `zodResolver` call site.
- [ ] Verify the `BookCreationWizard` validation tests pass.

---

**Impact:** Low urgency — neither is a security fix, and both current versions are healthy. The cost of leaving it untracked is repeated rediscovery, not risk.

frankbria:
Closes #385 (with #398, the zod half).

## Why single-package bumps can't work

`package.json` pinned **twelve** `@tiptap` packages, so bumping `starter-kit` alone left eleven siblings at `^2.x`. The resulting duplicated `@tiptap/core` produced exactly the errors the issue records — `Mark<any, any> is not assignable to Mark<any, any>` is two copies of the same type from two copies of the package, and the missing `toggleBold`/`toggleItalic`/… commands follow from the same split.

## Two findings that contradict the issue's analysis

Both worth recording, since the issue exists specifically to stop this being re-derived:

**Only five of the twelve packages were ever imported** — `react`, `starter-kit`, `extension-{underline,placeholder,character-count}`. The other seven (`blockquote`, `bullet-list`, `heading`, `link`, `list-item`, `ordered-list`, `strike`) were pinned but never referenced anywhere in `src/`; StarterKit already bundled them. They're **removed rather than upgraded** — they contributed nothing except the version skew that broke the bump.

**v3's StarterKit bundles *more* than v2's, not less** — 24 packages, including `underline` and `link`. The issue states *"v3 unbundles the extensions from starter-kit, so every command disappears from the chain"*. That describes the partial-bump symptom, not v3. With the family moved together, every command resolves and **no extension needed registering** — so AC 2 turns out not to apply.

## What the migration actually required

- `@tiptap/pm` promoted to a direct dependency: v3 makes it a peer with an *exact* version requirement, and leaving it transitive is what npm couldn't resolve.
- The lockfile's `@tiptap` subtree had to be pruned before npm would re-resolve — it kept anchoring on `extension-character-count@2.27.1` and reporting a conflict against a peer it was itself pinning.

**No source changes.** The diff is `package.json` + `package-lock.json`.

## Verification

```
$ npx tsc --noEmit ; echo $?
0
```

The issue's `TS2551` (`toggleBold does not exist on ChainedCommands`) and `TS2769` (`No overload matches this call`) are both gone.

| Check | Result |
|-------|--------|
| `@tiptap/core` instances in the tree | **1** (`3.29.2`) — was the root cause |
| Distinct `@tiptap` versions | **1** (`3.29.2`) |
| `tsc --noEmit` | exit 0 |
| Full frontend suite (coverage) | **133/133 suites, 2287 passed**, thresholds met |
| Editor tests (`RichTextEditor`, `EditorToolbar`, `PreferenceDefaults`) | 38 passed |
| Production build | compiles; editor routes byte-identical in size to v2 (374 kB / 206 kB First Load) |
| Editor E2E (added by #341) | runs in CI on this PR |

## Acceptance criteria

- [x] Move the entire `@tiptap/*` family to v3 in one PR — done, and the seven unused packages dropped.
- [x] Register the now-unbundled extensions explicitly — **not needed**; v3's StarterKit bundles them. Recorded above so this isn't re-investigated.
- [x] `npm run typecheck` clean and editor E2E passes.


frankbria:
## E2E caught a real v3 behaviour change

The one check that could. The unit suite mocks tiptap entirely, so only a browser driving the real editor sees this:

```
expect(page.locator('.tiptap p')).toHaveCount(0)
  unexpected value "1"
at editor-formatting.spec.ts:101 — 'Heading 2 converts the paragraph to an <h2>'
```

**Cause:** StarterKit v3 bundles `trailingNode` and enables it unless explicitly set to `false`. It keeps an empty paragraph at the end of the document so there's always somewhere to click below the last block — a sensible default in the abstract.

**Disabled deliberately.** This editor's HTML is what gets persisted and exported, so leaving it on would append `<p></p>` to every chapter a user saves, and to every PDF/DOCX/EPUB generated from one.

That's a content change, and a dependency upgrade is the wrong place to make one: it would arrive silently, apply retroactively to every future save, and be attributed to a version bump rather than a decision. Turning it on is a product call that deserves its own issue and a plan for existing documents.

The jest mock for `starter-kit` gained `.configure`, since the extension is now configured rather than passed bare.

```
133/133 suites, 2287 passed, 5 skipped
tsc --noEmit exit 0
```

This is also a small vindication of #341's investment in real-browser editor coverage — the migration typechecked clean, passed 2287 unit tests, and built successfully while changing what every save writes to the database.

PR #443, opened 2026-08-11
fix(auth): upgrade better-auth 1.4.9 → 1.6.26, clearing GHSA-p6v2-xcpg-h6xw

### LINKED ISSUE #411: [P2.9] better-auth 1.4.9 → 1.6.26: live high-severity advisory, blocked on Auth<> generic becoming invariant
Dependabot PR #405 bumps `better-auth` 1.4.9 → 1.6.25. It **cannot merge**: `Frontend Tests` (a required check) fails on typecheck because better-auth 1.6.x changed the `Auth<>` generic to be invariant over its options object.

Supply-chain triage on the PR came back clean — sole maintainer `bekacru` unchanged, no new lifecycle scripts, version published 2026-07-23. This is a **compatibility** blocker, not a supply-chain one, so the PR is held rather than closed.

## Failures

```
src/lib/auth.ts(110,5): error TS2322: Type 'Auth<{ appName: string; database: (options: BetterAuthOptions)
  => DBAdapter<BetterAuthOptions>; plugins: [{ id: "two-factor"; ... }]; emailAndPassword: {...};
  session: {...}; advanced: {...}; }>' is not assignable to type 'Auth<BetterAuthOptions>'.

src/app/api/auth/[...all]/route.ts(20,34): error TS2345: Argument of type 'Auth<BetterAuthOptions> | null'
  is not assignable to parameter of type '{ handler: (request: Request) => Promise<Response>; }
  | ((request: Request) => Promise<Response>)'.
```

## Why it's worth doing rather than pinning

Two minor versions of accumulated fixes, several security-relevant:

- Apple OAuth now sends the PKCE code challenge during authorization (token exchange previously failed) — better-auth#10294
- Google One Tap no longer creates users when sign-up is disabled on the Google provider — better-auth#10479
- `get-session` now returns `no-store` cache-control headers, preventing stale session data being served — better-auth#10222
- magic-link and email-OTP send endpoints now validate the `Origin` header on cookieless requests, preventing cross-origin abuse — better-auth#10368

None of these are advisory-tracked, and the repo currently has Dependabot **security alerts disabled**, so there is no automated signal that would escalate this on its own.

## Tasks

- [ ] Read the better-auth 1.5.0 and 1.6.0 release notes for the `Auth<>` generic change and the intended migration (likely `Auth<typeof authOptions>`, or dropping the explicit annotation at `auth.ts:110` and letting it infer).
- [ ] Fix the null-narrowing at `route.ts:20` — `toNextJsHandler` no longer accepts `Auth | null`.
- [ ] `cd frontend && npm run typecheck && npm test` green against 1.6.25.
- [ ] Verify session/JWT flow end-to-end against staging. The `no-store` header change and the `sameSite="lax"` CSRF fix from #341 both touch cookie behavior.
- [ ] Push to `dependabot/npm_and_yarn/frontend/better-auth-1.6.25` (or a branch that supersedes it) and merge PR #405.

## Priority note

Slotted at P3.19 alongside #385 (frontend major bumps blocked on coordinated migrations) — same shape of work. Pull this forward to P2 if magic-link, email-OTP, Google One Tap, or Apple OAuth are enabled for real users, since the fixes above stop being theoretical at that point.

Blocks: PR #405.

frankbria:
Closes #411. Supersedes #405 (Dependabot's 1.6.25 bump, which could not merge — it targets the stale patch and carries none of the type work).

## Why

`better-auth` 1.4.9 is inside the vulnerable range of [GHSA-p6v2-xcpg-h6xw](https://github.com/advisories/GHSA-p6v2-xcpg-h6xw) (**high**): versions `< 1.4.17` key rate-limit buckets on individual IPv6 addresses, so the limiter is bypassable by rotating within a prefix. Rate limiting is core, not plugin-gated, so this applies as configured here.

The other 19 published advisories for this package are gated behind `oidc-provider` / `mcp` / `api-key` / `organization` / `admin` / OAuth-social, none of which this app enables — checked against `frontend/src/lib/auth.ts` (`emailAndPassword` + `twoFactor()` + `mongodbAdapter`).

Targets **1.6.26**, not the 1.6.25 in #405 — 1.6.26 is current stable (1.7.x is still RC).

## The type fix

1.6 made `Auth<TOptions>` invariant over its options, so the singleton no longer accepted its own assigned value:

```ts
let authInstance: ReturnType<typeof betterAuth> | null   // = Auth<BetterAuthOptions>, widened
authInstance = betterAuth({ ...concrete options... })    // = Auth<{ appName; plugins: [TwoFactorPlugin]; ... }>
```

`tsc` traced the mismatch to `twoFactorEnabled`: the `twoFactor()` plugin extends the user schema, so the widened type is a genuinely *different* shape, not a looser one.

Fix — extract the call into a factory and type the cache as its return:

```ts
function createAuthInstance(database: Db) { return betterAuth({ /* … */ }) }
let authInstance: ReturnType<typeof createAuthInstance> | null = null
```

**This is more than a compile fix.** The old annotation was erasing the plugin's types all along — under `Auth<BetterAuthOptions>`, `$Infer.Session.user.twoFactorEnabled` and `api.enableTwoFactor` were invisible on 1.4.9 too. The factory restores 2FA type safety that had been silently lost.

**The second reported error fixed itself.** The issue lists `route.ts:20` as a separate failure, but `Auth<BetterAuthOptions> | null is not assignable` was a *symptom* of the widened cache leaking into `getAuth()`'s return type. `src/app/api/auth/[...all]/route.ts` is unchanged. (`toNextJsHandler`'s signature is byte-identical between 1.4.9 and 1.6.26 and is structural — it only wants `{ handler }`.)

## Dependency resolution — why `overrides` and not `--legacy-peer-deps`

npm could not resolve 1.6.x at all. better-auth declares an optional peer on `@tanstack/react-start`; npm explored it into `@vitejs/plugin-react` → `@rolldown/plugin-babel`, which needs `@babel/core ^7.29` against our transitive `7.28.5`.

`--legacy-peer-deps` would have worked but disables peer checking for the entire tree. Pinning the floor in the existing `overrides` block is narrower, and matches how this repo already handles `js-yaml`.

Verified the optional-peer chain is **not** installed — `@tanstack/react-start`, `@vitejs/plugin-react` and `@rolldown/plugin-babel` are all absent from the lockfile.

Import paths are unchanged: `better-auth/adapters/mongodb` is now a one-line re-export of `@better-auth/mongo-adapter`, and `twoFactor` still ships from `better-auth/plugins`.

## Runtime verification — and why it was necessary

**Nothing in CI exercises this upgrade.** Three independent reasons:

- `jest.config.cjs` maps `better-auth/react`, `/client` and `/client/plugins` to hand-written mocks, and `lib/auth.ts` has no unit tests at all
- `tsconfig.json` excludes every test path from `tsc`
- the CI E2E job runs with `BYPASS_AUTH=true` / `E2E_ALLOW_BYPASS=1`, so it skips better-auth entirely

So a green pipeline would have proven nothing here. The following was exercised by hand against a real MongoDB and a production build, with **no auth bypass**:

| Check | Result |
|---|---|
| `POST /api/auth/sign-up/email` | 200; sets `better-auth.session_token` + separate `better-auth.session_data` |
| `POST /api/auth/sign-in/email` | 200 |
| Wrong password | 401 |
| `GET /api/auth/get-session` with signed cookie | 200 |
| `twoFactorEnabled` on returned user | present — plugin schema live |

### The backend contract was the real risk

`backend/app/core/better_auth_session.py` does **not** validate a JWT. It splits the session cookie on `.` and queries better-auth's own Mongo collection directly, hardcoding better-auth's internal storage contract. A change there would fail silently as a blanket 401 — invisible to typecheck, tests, and bypassed E2E alike.

Every assumption it makes was verified intact under 1.6.26:

| Backend assumption | Verified |
|---|---|
| Collection named `session` | exists |
| `find_one({"token": <pre-dot part>})` | **FOUND** |
| Token stored raw, byte-equal to cookie segment | true |
| Fields `token` / `userId` / `expiresAt`, camelCase | all present |

(The session token alphabet is `a-z A-Z 0-9`, so it contains no dot and `split('.')[0]` is unambiguous.)

### Cross-version login — the one that would have locked everyone out

Created an account under **1.4.9**, reinstalled **1.6.26**, rebuilt, and signed that account in:

```
sign-in HTTP 200
twoFactorEnabled backfilled on legacy user: True -> False
find_one({token}) -> FOUND
```

Legacy scrypt hashes remain valid (params are byte-identical across versions) and legacy users without the newer `twoFactor` fields read cleanly rather than being locked out.

## Also included

`scripts/check-secrets.sh` (separate commit, `d447a4d`) — the `password` pattern was rejecting an **unmodified** line, because the refactor moved it and the scanner only inspects added (`+`) lines:

```
const { sendPasswordResetEmail } = await import("@/lib/email");
```

It matched "Password" inside the identifier, then `.*`, then the quoted module path. It's the loosest pattern in the set — its value side accepts any 8+ characters, where api-key/secret/token each demand a 32+ char alphanumeric run. Narrowed to require an assignment operator, the same fix already applied to the `NEXT_PUBLIC_` pattern. Verified in both directions: 6 real-secret forms still caught, 5 false-positive forms cleared. (`--no-verify` was **not** used; the fix is in the working tree, so the corrected pattern gates its own commit.)

## Known limitations

- **Upstream regression, not introduced here and not fixed here**: 1.4.9 compared password hashes with a constant-time helper; both 1.6 implementations use a plain `===` on the derived key. This is a timing-side-channel regression in better-auth itself. Worth an upstream issue; out of scope for this PR.
- `freshAge` now derives from `createdAt` rather than `updatedAt`. It gates only `unlinkAccount`, which this app never calls, so it is inert today — but it would matter if account deletion is added later.
- The `as AuthHandlers` cast in the auth route is now unnecessary. Left in place to keep this diff scoped to the upgrade.
- Verification above was against a local Mongo, not staging. `npm run test:e2e:staging` uses real auth and should be run post-deploy — that path still can't be exercised pre-merge, since staging runs the deployed build.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -37,7 +37,7 @@
         "@tiptap/pm": "^3.29.2",
         "@tiptap/react": "^3.29.2",
         "@tiptap/starter-kit": "^3.29.2",
-        "better-auth": "^1.4.7",
+        "better-auth": "^1.6.26",
         "class-variance-authority": "^0.7.0",
         "clsx": "^2.0.0",
         "date-fns": "^4.1.0",
@@ -158,12 +158,12 @@
       }
     },
     "node_modules/@babel/code-frame": {
-      "version": "7.27.1",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.27.1.tgz",
-      "integrity": "sha512-cjQ7ZlQ0Mv3b47hABuTevyTuYN4i+loJKGeV9flcCgIK37cCXRh+L1bd3iBHlynerhQ7BhCkn2BPbQUL+rGqFg==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.29.7.tgz",
+      "integrity": "sha512-Aup7aUOfpbAUg2ROOJN6Iw5f9DMBlzu0mIkm/malLQFN/YQgO48wCj0Kxa3sEHJvPVFg7siR+qRInwXd2qhQKw==",
       "license": "MIT",
       "dependencies": {
-        "@babel/helper-validator-identifier": "^7.27.1",
+        "@babel/helper-validator-identifier": "^7.29.7",
         "js-tokens": "^4.0.0",
         "picocolors": "^1.1.1"
       },
@@ -172,29 +172,29 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.28.5",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.28.5.tgz",
-      "integrity": "sha512-6uFXyCayocRbqhZOB+6XcuZbkMNimwfVGFji8CTZnCzOHVGvDqzvitu1re2AU5LROliz7eQPhB8CpAMvnx9EjA==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.29.7.tgz",
+      "integrity": "sha512-locTkQyKvwIEgBzVrn8693ebc97F2U8ZHjbXwDXJ5Fn2TCpNwTlKcaKLkdHop5c/icOFE7qt7Q9JC5hnKNa6Gg==",
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
       }
     },
     "node_modules/@babel/core": {
-      "version": "7.28.5",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.28.5.tgz",
-      "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
+      "version": "7.29.7",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.29.7.tgz",
+      "integrity": "sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==",
       "license": "MIT",
       "dependencies": {
-        "@babel/code-frame": "^7.27.1",
-        "@babel/generator": "^7.28.5",
-        "@babel/helper-compilation-targets": "^7.27.2",
-        "@babel/helper-module-transforms": "^7.28.3",
-        "@babel/helpers": "^7.28.4",
-        "@babel/parser": "^7.28.5",
-        "@babel/template": "^7.27.2",
-        "@babel/traverse": "^7.28.5",
-        "@babel/types": "^7.28.5",
+        "@babel/code-frame": "^7.29.7",
+        "@babel/generator": "^7.29.7",
+        "@babel/helper-compilation-targets": "^7.29.7",
+        "@babel/helper-module-transforms": "^7.29.7",
+        "@babel/helpers": "^7.29.7",
+        "@babel/parser": "^7.29.7",
+        "@babel/template": "^7.29.7",
+        "@babel/traverse": "^7.29.7",
+        "@babel/types": "^7.29.7",
         "@jridgewell/remapping": "^2.3.5",
         "convert-source-map": "^2.0.0",
         "debug": "^4.1.0",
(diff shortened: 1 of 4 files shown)


frankbria:
## Pre-PR cross-family review (Phase 7c-1)

**Primary reviewer (opencode / GLM) was unavailable** — it failed with a server-side error (`UnknownError`, ref `err_4d34f414`) rather than returning findings. Fell back to `codex review --base main` per the review-invariants rule.

### Finding — [P2] Scanner blind spot, **accepted and fixed** in 459c1d3

codex answered the exact question it was asked to probe: *construct a real secret the old pattern caught and the new one misses.* It did:

```
password-value: "supersecret"
"password-field": "supersecret"
```

The narrowing in d447a4d used `[A-Za-z0-9_]*` as the name continuation, which cannot cross a hyphen — so after `password` the match could never reach the assignment operator, and those YAML/JSON-style keys silently stopped being flagged. Real regression, introduced by me, in a security check. Fixed by allowing `-` and `.` in that class.

Worth noting `db-password: "..."` was never affected (there the hyphen *precedes* the word), so the gap was narrower than it first looks — but still real.

Re-verified in both directions after the fix: **10 real-secret forms caught** (including both of codex's cases), **6 false-positive forms clean** (including the original `sendPasswordResetEmail` line that started this).

### Not accepted

No other Critical/Major findings were reported. One thing I checked independently and am **not** fixing here, because it is pre-existing and out of scope:

`getAuth()` assigns `client` at `auth.ts:108` *before* `await client.connect()` at `:109`, so two concurrent callers can each build an auth instance, last write winning. This predates the upgrade. The change actually improves the symptom slightly — returning the local `instance` means each caller gets the object it created, where previously both read the shared `authInstance` and one could receive the other's. Flagging rather than widening this PR's scope.

<sub>Third-party review is on the record twice per the workflow; this is the pre-PR pass. The repo's own `review / review` check (glm) also runs on this PR — unlike on Dependabot PRs, it can authenticate here.</sub>

frankbria:
### Correction to the PR body

I wrote, of the secret-scanner commit:

> (`--no-verify` was **not** used; the fix is in the working tree, so the corrected pattern gates its own commit.)

The first half is true — no `--no-verify`. **The second half is wrong.** `check-secrets.sh` explicitly skips itself (lines 65-73):

```bash
if [ "$FILE" = "scripts/check-secrets.sh" ]; then
    continue
fi
```

with the rationale that the PATTERNS array is by construction a list of strings that look exactly like the secrets it hunts for, so every rule would flag its own definition. That is a sound design decision — but it means nothing gated that commit, and I shouldn't have implied the fix validated itself.

The actual evidence for the pattern change is unchanged and stands on its own: the explicit two-direction test, now covering **10 real-secret forms caught** and **6 false-positive forms clean** after the codex fix in 459c1d3.

Correcting it here rather than silently, since it was a claim about how the change was verified.

Issue #427, opened 2026-08-10
[P2.11] Replace PM2/rsync deploy CI with Docker image builds + Compose on the VPS


frankbria:
## Problem

`.github/workflows/deploy-staging.yml` (449 lines) deploys by SSH-ing to the shared VPS, `npm ci`/`uv sync` **on the box**, symlinking `releases/<id>` → `current`, rendering `ecosystem.config.template.js`, then `pm2 delete && pm2 start && pm2 save`. `deploy-production.yml.disabled` (259 lines) is the same shape.

Problems this causes today:
- Build happens on the shared VPS, so deploys depend on the box's node/python/uv versions, its disk, and whatever else is running there.
- No immutable artifact — a rollback means re-running the old release dir and hoping its `node_modules` still match.
- `pm2 delete` before `pm2 start` is a hard stop/start, not a zero-downtime swap.
- Port collisions are managed by convention (8000 backend / 3002 frontend) on a box shared with other apps.
- The symlink + PM2 + nginx trio has drifted out of sync before (see `docs/PM2-DEPLOYMENT-FIXES.md`, `docs/STAGING_DEPLOYMENT_ANALYSIS.md`).

## Proposal

Move deploys to containers. Build images in CI, ship images to the box, run them with Docker Compose. No orchestrator — this is one small VPS.

**Scope:**
1. `backend/Dockerfile` — multi-stage, `uv sync --frozen` (no `--extra test`), non-root user, uvicorn entrypoint.
2. `frontend/Dockerfile` — multi-stage Next.js build, `output: "standalone"`, non-root user.
3. `docker-compose.yml` (+ `docker-compose.staging.yml` override) — backend, frontend, published on the existing host ports so nginx upstreams don't change. Mongo stays on Atlas; no Mongo container.
4. `.dockerignore` for both apps.
5. Rewrite `deploy-staging.yml`: build + push images to GHCR tagged with the commit SHA → SSH → `docker compose pull && docker compose up -d` → existing `/health` smoke check → keep the last N images.
6. Port `deploy-production.yml.disabled` to the same shape (stays disabled).
7. Update `docs/STAGING-DEPLOYMENT.md` / `.github/DEPLOYMENT.md`; mark the PM2 docs historical.

**Out of scope:** Kubernetes, containerizing local dev, containerizing CI test jobs, moving Mongo off Atlas.

## Acceptance criteria

- [ ] Both images build in CI and push to GHCR tagged `sha-<short>` + `staging`.
[...]

PR #485, opened 2026-08-12
fix(deploy): route the staging deploy over Tailscale

frankbria:
Dispatching the container deploy (the last unproven piece of #427) exposed why **no automated deploy has worked for about two weeks** — and it isn't specific to containers.

## The blocker

The box's UFW allows port 22 from exactly two sources:

```
22/tcp             ALLOW  70.172.0.0/17   # Frank Cox subnet - never block
22 on tailscale0   ALLOW  Anywhere
```

GitHub-hosted runners are on neither, so SSH from Actions times out.

**Evidence this predates #427:**
- The PM2 workflow this replaced failed on **all 8 of its runs on 2026-08-12** (03:02–04:49)
- The newest release directory on the box is dated **2026-07-30**
- My own cutover succeeded only because it ran from the allowed subnet

So the container migration didn't break automated deploys — it surfaced that they were already broken.

## The fix

Join the tailnet. The alternative — punching GitHub's published IP ranges through the firewall — is a large, permanent widening of SSH exposure on a machine that also runs **other people's applications**, to accommodate a transient runner.

Adds a  step before SSH setup, and routes every scp/ssh at  so the tailnet address is used when configured, with the public address as fallback.

## Two secrets are needed, and I can't create them

| Secret | Value |
|---|---|
|  | an **ephemeral, pre-approved** auth key for the runner |
|  | the box's tailnet address — , or its MagicDNS name |

Without  the step is skipped and the deploy fails at the scp with a connection timeout — **loudly, at the same place it fails today**, rather than appearing to work.

## What is and isn't proven

**Proven:** the deploy itself. Three container deploys and a rollback, all healthy, verified publicly through unchanged nginx.

**Unproven:** only the network path from a GitHub runner, which needs the secrets above.

### DIFF
--- a/.github/workflows/deploy-staging-containers.yml
+++ b/.github/workflows/deploy-staging-containers.yml
@@ -56,6 +56,30 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v7
 
+      # The box's UFW allows port 22 from exactly two places: the owner's home
+      # subnet (70.172.0.0/17, commented "never block") and the tailscale0
+      # interface. GitHub-hosted runners are on neither, so SSH from Actions
+      # times out. This is not new and not specific to containers — the PM2
+      # workflow this replaced failed the same way on all 8 of its runs on
+      # 2026-08-12, and the newest release directory on the box is dated
+      # 2026-07-30. Automated deploys have been broken for ~2 weeks.
+      #
+      # Joining the tailnet is the sanctioned route in rather than punching
+      # GitHub's IP ranges through the firewall, which would be a large and
+      # permanent widening for a transient runner.
+      #
+      # Requires a TAILSCALE_AUTHKEY secret (an ephemeral, pre-approved auth key
+      # is the right kind). Without it this step is skipped and the deploy fails
+      # at the scp with a connection timeout — loudly, not silently.
+      - name: Connect to Tailscale
+        if: env.TS_AUTHKEY != ''
+        env:
+          TS_AUTHKEY: ${{ secrets.TAILSCALE_AUTHKEY }}
+        uses: tailscale/github-action@v3
+        with:
+          authkey: ${{ secrets.TAILSCALE_AUTHKEY }}
+          version: latest
+
       - name: Setup SSH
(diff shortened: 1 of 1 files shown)


frankbria:
Dispatching the container deploy — the last unproven piece of #427 — exposed why **no automated deploy has worked for about two weeks**. It isn't specific to containers.

## The blocker

The box's UFW allows port 22 from exactly two sources:

```
22/tcp             ALLOW  70.172.0.0/17   # Frank Cox subnet - never block
22 on tailscale0   ALLOW  Anywhere
```

GitHub-hosted runners are on neither, so SSH from Actions times out.

**Evidence this predates #427:**

- The PM2 workflow this replaced failed on **all 8 of its runs on 2026-08-12** (03:02–04:49)
- The newest release directory on the box is dated **2026-07-30**
- My own cutover succeeded only because it ran from the allowed subnet

So the container migration didn't break automated deploys — it surfaced that they were already broken.

## The fix

Join the tailnet. The alternative — punching GitHub's published IP ranges through the firewall — is a large, permanent widening of SSH exposure on a machine that also runs **other people's applications**, to accommodate a transient runner.

Adds a `Connect to Tailscale` step before SSH setup, and routes every scp/ssh at `secrets.STAGING_TS_HOST || secrets.HOST`, so the tailnet address is used when configured with the public address as fallback.

## Two secrets are needed, and I can't create them

| Secret | Value |
|---|---|
| `TAILSCALE_AUTHKEY` | an **ephemeral, pre-approved** auth key for the runner |
| `STAGING_TS_HOST` | the box's tailnet address — `100.124.242.39`, or its MagicDNS name |

Without `TAILSCALE_AUTHKEY` the step is skipped and the deploy fails at the scp with a connection timeout — **loudly, at the same place it fails today**, rather than appearing to work.

## What is and isn't proven

**Proven:** the deploy itself. Three container deploys plus a rollback, all healthy, verified publicly through unchanged nginx.

**Unproven:** only the network path from a GitHub runner, which needs the two secrets above.

PR #495, opened 2026-08-13
[P2.16] fix(toc): reject TOC payloads missing the wrapper instead of saving an empty TOC

### LINKED ISSUE #492: [P2.16] PUT /books/{id}/toc returns 200 while saving an empty TOC when the payload lacks the "toc" wrapper
Found while fixing the staging E2E suite (#488). A client that sends the wrong body shape gets a **200 and an empty table of contents**, with no indication anything went wrong.

## The defect

`backend/app/api/endpoints/books.py`, `update_book_toc`:

```python
data: dict = Body(...)
toc_data = data.get("toc", {})          # missing key -> {}
chapters = toc_data.get("chapters", [])  # -> []
```

`data` is an untyped `dict`, so FastAPI validates nothing. A body of

```json
{"chapters": [ ...50 chapters... ], "total_chapters": 50}
```

resolves `toc` to `{}`, `chapters` to `[]`, passes both `isinstance` guards, and saves an empty TOC — returning **200**.

## Confirmed on staging

The E2E suite PUT 50 chapters, asserted `status < 400` (which passed), then found nothing:

```
book: Large TOC Edge Book 1786594812649
table_of_contents chapters: 0
toc_items: 0
```

## Why this matters beyond the test

- **A 2xx is being returned for a request that accomplished nothing.** Any client with a slightly wrong shape — a version skew, a refactor, a third-party integration — silently loses the user's chapter structure and is told it succeeded.
- Worth checking whether this **overwrites an existing TOC** with the empty one. If so a malformed request destroys real work. I did not verify that path, and it should be checked before deciding severity.
- The absent `toc` key is indistinguishable from a deliberate "save an empty TOC", so the endpoint cannot currently tell a mistake from an intent.

## Suggested fix

- Reject when `toc` is absent: `400` rather than defaulting to `{}`. An explicit empty TOC can still be sent as `{"toc": {"chapters": []}}`.
- Better: replace `data: dict = Body(...)` with a Pydantic model so FastAPI validates the shape and returns 422 automatically. That is the pattern the rest of this module already uses.
- Add a regression test: a flat (unwrapped) payload must not return 2xx.

## Priority

P2.16 — silent, user-facing data loss on a core save path, but it requires a malformed request rather than firing on the normal flow, so below the live-advisory tier.

The E2E side is already handled: the spec now wraps the payload correctly **and** asserts a read-back, so a 2xx can no longer stand in for persistence there.

frankbria:
Closes #492

## The defect, confirmed

`PUT /api/v1/books/{id}/toc` read its body as an untyped `dict` and defaulted both keys when absent:

```python
toc_data = data.get("toc", {})           # missing key -> {}
chapters = toc_data.get("chapters", [])  # -> []
```

The issue left one question open — whether this *overwrites* an existing TOC. It does. `_update_toc_internal` (`backend/app/db/toc_transactions.py:155`) writes `{**toc_data, ...}` as the whole `table_of_contents`, so a wrong-shaped body:

1. resolves to an empty TOC,
2. **destroys the book's real chapters**,
3. bumps the version, and
4. returns **200 `success: true`**.

So this is silent, user-facing data loss reported as success, not merely a no-op — the severity assumption in the issue holds.

## The fix

Both keys are now required rather than defaulted; a missing one is a `400`. An intentional clear is still expressible as `{"toc": {"chapters": []}}`, which the existing chapter validation already accepted (now pinned by a test).

## Design note — why not a Pydantic model

The issue suggested replacing `data: dict = Body(...)` with a Pydantic model. I kept the hand-rolled validation deliberately: FastAPI would return **422** where this endpoint currently returns **400** for four already-tested malformed shapes (`toc` not an object, `chapters` not a list, chapter not an object, chapter missing a title). That is a public error-contract change wider than the reported bug, and the frontend branches on status. The keys-required guard closes the data-loss path in 12 lines without touching the contract.

## Tests (TDD — RED first)

Three added to `TestUpdateToc`, all failing-then-passing against the fix:

- `test_unwrapped_payload_is_rejected_and_preserves_toc` — saves a real TOC, PUTs the exact flat shape from the issue, asserts `400` **and reads back** that the original chapters survived. This is the regression that would have caught #492.
- `test_toc_without_chapters_key_is_rejected` — `{"toc": {"total_chapters": 5}}` is malformed, not "save empty".
- `test_explicit_empty_chapters_still_saves` — pins the intentional-clear path against future over-tightening.

Also removed a redundant function-local `logging.getLogger` that shadowed the module logger at `books.py:91`.

## Verification

- Backend: **1236 passed, 9 skipped**, coverage **92%** (floor 85%).
- Pre-commit gates run and green: secrets, backend ruff, unit tests, coverage.
- Every in-repo caller already sends the wrapped shape — `bookClient.updateToc` (`frontend/src/lib/api/bookClient.ts:851`), `e2e/helpers/testData.ts:78`, `tests/e2e/staging/edge-cases.spec.ts:39` — so no client is broken by requiring the keys.

## Third-party review

opencode (z.ai GLM) reviewed the branch diff pre-PR: **no Critical, no Major**. It confirmed independently that no in-repo client breaks and that the data-loss path is closed.

## Known limitations

The reviewer flagged one **Minor, latent** issue outside this fix's scope: `_update_toc_internal` still replaces the TOC wholesale, so a partial update sending only `chapters` would reset `total_chapters` / `estimated_pages` / `structure_notes` (and `GET` would then report `total_chapters: 0` alongside real chapters). No current caller does this — the frontend always resends the full object — so it is not live. Filed separately rather than widened into this PR.


frankbria:
## Third-party review — opencode (z.ai GLM), pre-PR on the branch diff

Diff unchanged since the review ran (`ecdb6a7`). Verbatim output:

---

The diff removes a function-local `import logging; logger = ...` but keeps the `logger.info(...)` call. Let me verify `logger` exists at module scope, or the happy path now 500s.## Review: PUT `/books/{id}/toc` required-keys fix

**No Critical. No Major. One Minor (latent).**

### (1) Legitimate client breakage — **None**
The canonical client `bookClient.updateToc` (frontend/src/lib/api/bookClient.ts:812) types `chapters` as required and always sends `{ toc }`. Both e2e helpers (`testData.ts:78`, `edge-cases.spec.ts:38`) wrap in `toc`. No in-repo caller sends the previously-defaulted shapes. Requiring the keys aligns the contract with the only client that exists.

### (2) Data-loss path closed — **Yes, for the reported bug. One residual Minor**
The chapters-wipe path is closed: `toc` and `chapters` are both required + type-checked before any write, and `_update_toc_internal` spreads `toc_data` (toc_transactions.py:155-161), so `chapters` is always present. The only way to reach an empty TOC now is the documented `{"toc":{"chapters":[]}}` (tested).

- **Minor — wholesale replace still drops non-`chapters` fields.** `_update_toc_internal` still does `{**toc_data, ...}`, so a caller that sends `{"toc":{"chapters":[...]}}` without `total_chapters` / `estimated_pages` / `structure_notes` silently resets them (GET then defaults `total_chapters` to 0 at books.py:1326 while real chapters exist — a wrong count). This is the same root pattern as #492 applied to metadata fields. Not live today (the frontend resends all fields), but it's latent and will bite the next partial-update caller. Either preserve prior fields (`{**current_toc, **toc_data, ...}`) or compute `total_chapters` from `len(chapters)`.

The `logger` cleanup is safe — `logger` is module-level at books.py:91 (the removed lines were redundant shadowing).

### (3) Tests — **Meaningful regressions**
`test_unwrapped_payload_is_rejected_and_preserves_toc` is the gold-standard regression: it reproduces the exact #492 flat-body shape, asserts 400, *and* verifies via GET that the prior chapters survived — it fails on the old code and passes on the fix. The `chapters`-absent test covers the second guard; the empty-`chapters` test pins the intentional-clear contract against future over-tightening. No gaps worth blocking on.

frankbria:
## Demo — the data loss reproduced, then closed

Full narrated run: [`docs/demos/2026-08-13-issue-492-toc-payload-validation.md`](../blob/fix/492-toc-payload-validation/docs/demos/2026-08-13-issue-492-toc-payload-validation.md). Every step runs against **real MongoDB** through the real app; steps [3] and [5] are `GET` read-backs, so the evidence is what is in the database, not what the handler claimed.

**Before the fix** — an author with 3 saved chapters, then one wrong-shaped request:

```
[1] Author saves a real TOC     -> 200  chapters_count=3  version=2
[2] Client PUTs 50 chapters flat -> 200  body={"book_id":"...","toc":{},"version":3,...}
[3] Read back from MongoDB       -> 0 chapters: []
    version now                  -> 3
```

Their work is gone, the version was bumped so the optimistic lock treats it as a legitimate edit, and the client was told it succeeded. This answers the question the issue left open: the malformed request does not no-op, it **overwrites real work**.

**After the fix** — same client, same payload:

```
[2] Client PUTs 50 chapters flat -> 400  body={"detail":"Request body must contain a 'toc' object"}
[3] Read back from MongoDB       -> 3 chapters: ['Why Systems Fail', 'The Silent 200', 'Designing for Loud Failure']
    version now                  -> 2
[4] Deliberate clear             -> 200  chapters_count=0
[5] Read back after clear        -> 0 chapters
```

The bad request changed nothing; the deliberate clear still works. The endpoint can now tell a mistake from an intent.

### Acceptance criteria

| From the issue | Evidence |
|---|---|
| Reject when `toc` is absent — 400, not a defaulted `{}` | Step [2] after: `400 "Request body must contain a 'toc' object"` |
| Explicit empty TOC still sendable as `{"toc": {"chapters": []}}` | Steps [4]/[5]: `200`, `chapters_count=0`, reads back as 0 |
| Regression test: a flat payload must not return 2xx | `test_unwrapped_payload_is_rejected_and_preserves_toc` — asserts status **and** read-back; verified failing on pre-fix code |
| *(open question)* Does it overwrite an existing TOC? | **Yes** — step [3] before the fix. Now preserved. |

### Docs

While syncing docs I found `docs/api-toc-endpoints.md` documented the request body **unwrapped** — the destructive shape itself. Anyone integrating from the docs would have hit this bug. Corrected in `fb87bc7`, along with the endpoint's real 400 `detail` strings and its 409.

Full suite after both commits: **1236 passed, 9 skipped, 92% coverage**.

Issue #517, opened 2026-08-25
[P2.19] Dependabot has no PR grouping and auto-merge is disabled, so batches serialize and Dependabot silently self-closes valid bumps


frankbria:
## Problem

Dependency batches on this repo are **expensive and occasionally lossy**, because three settings compound:

1. `.github/dependabot.yml` defines **no `groups:`** — every package gets its own PR (`open-pull-requests-limit: 5` per ecosystem, so up to 15 open at once).
2. `main` branch protection has `required_status_checks.strict: true` — branches must be up to date, so every merge flips all other open PRs to `BEHIND`.
3. **Repo-wide auto-merge is disabled** — `gh pr merge --auto` fails with `GraphQL: Auto merge is not allowed for this repository (enablePullRequestAutoMerge)`.

Together these force a strictly serial update-branch → full CI cycle → merge loop, with no way to queue it.

## Measured cost

The 2026-08-25 sweep: 14 open PRs, 9 merged. Each merge required a rebase and a complete CI cycle (Frontend Tests, Backend Tests, E2E Playwright, Security Audit) at **~10 minutes apiece** — roughly **2h20m of wall-clock and 9 full CI runs** to land what is, in total, four one-line workflow edits, two Python version pins, and three lockfile bumps.

## The lossy part — this is the real motivation

During that rebase storm, with four PRs all touching `.github/workflows/build-images.yml` (#498, #501, #504, #506), **Dependabot closed #504 itself** two minutes after #498 merged:

> Looks like docker/setup-buildx-action is up-to-date now, so this is no longer needed.

`main` still pinned `@v3` at that moment. The claim was simply wrong. Dependabot deleted the branch on close, so `gh pr reopen 504` failed with `Could not open the pull request`, and the bump had to be recreated by hand as #515.

**Nothing surfaces this.** A self-closed Dependabot PR is indistinguishable from an intentionally-declined one. Had nobody re-read the file on `main`, the repo would have silently kept `setup-buildx-action@v3` until some later scheduled run happened to re-notice — and the same failure mode applies to a bump that is patching an advisory.

## Proposed fix

1. **Add `groups:` to `.github/dependabot.yml`** — at minimum one group per ecosystem (`github-actions`, `npm` dev vs prod, `uv`). This collapses N PRs touching one file into one PR, which removes the rebase storm at its source and cuts CI runs proportionally.
[...]

Issue #439, opened 2026-08-10
[P2.13] Dependabot's uv ecosystem writes backend bumps to the generated requirements.txt, so transitive backend updates are silent no-ops


frankbria:
This is the follow-up c4eb4c2 (#426) explicitly called for and that was never filed:

> Follow-up worth its own change: point the uv ecosystem in .github/dependabot.yml at the lockfile so future backend bumps are not no-ops.

It has now recurred. On 2026-08-10 Dependabot opened four more backend PRs with the identical defect — #430 (charset-normalizer), #431 (wsproto), #432 (pyzmq), #434 (typing-extensions). All four were closed as no-ops.

## The defect

Dependabot's `uv` ecosystem edits `backend/requirements.txt` for **transitive** dependencies. That file is a generated artifact — its own header records the command that produces it:

```
# This file was autogenerated by uv via the following command:
#    uv export --all-extras --no-emit-project --no-hashes --format requirements-txt -o requirements.txt
```

Nothing installs from it. Deploy and CI both run `uv sync` against `uv.lock`, and the audit gate feeds pip-audit from `uv export`, which also reads `uv.lock`. So a PR touching only that file:

- changes no installed version,
- clears no advisory,
- leaves the generated file disagreeing with its source until the next `uv export` silently reverts it.

Verified for all four of the 2026-08-10 PRs — none appears in `pyproject.toml`, and each sits at the pre-bump version in `uv.lock`:

| Package | uv.lock | requirements.txt (PR) | In pyproject? |
|---|---|---|---|
| charset-normalizer | 3.4.2 | 3.4.9 | no |
| wsproto | 1.2.0 | 1.3.2 | no |
| pyzmq | 27.0.0 | 27.1.0 | no |
| typing-extensions | 4.14.1 | 4.16.0 | no |

**Direct** dependencies behave correctly — #436 (boto3 1.43.40 → 1.43.56) touched `pyproject.toml`, `uv.lock`, and the export together, because boto3 is declared in `pyproject.toml`. The split is transitive-vs-direct, not random.

## Why this matters beyond noise

The visible symptom is churn — recurring PRs that must be triaged and closed. The real risk is the inverse case: **a backend transitive dependency with a genuine advisory would produce a PR that looks like a fix and applies nothing.** Merging it would close the alert while leaving the vulnerable version installed. Combined with #413 (Dependabot security alerts off, Snyk out of quota), the backend currently has no dependency-update path that can be trusted to actually take effect.

[...]

PR #521, opened 2026-08-25
fix(deps): hide the generated requirements export from Dependabot (#512)

### LINKED ISSUE #512: [P2.17] Dependabot uv config opens unmergeable no-op PRs for backend transitive deps
## Problem

Dependabot's `uv` ecosystem config produces **unmergeable no-op PRs for backend transitive dependencies**. It edits only `backend/requirements.txt` — a generated export of `uv.lock` — leaving `pyproject.toml` and `uv.lock` untouched.

Consequences:
1. **The bump changes nothing installed.** `uv sync` resolves from `uv.lock`; the export is not an input.
2. **It fails the required `Security Audit` check** on the lock/export sync gate, whose own message says: *"If a dependency bump only touched requirements.txt, apply it to pyproject.toml/uv.lock instead — editing the export alone changes nothing that is installed."*

Three PRs in the 2026-08-25 Dependabot sweep were exactly this and had to be closed: #499 (flask-cors), #503 (tqdm), #505 (pydantic-core).

Direct dependencies pinned in `pyproject.toml` are handled correctly — #500 (pytest-asyncio) and #502 (pydantic-settings) touched all three files and merged green. The defect is specific to **transitives**, which Dependabot only ever sees in the export.

## Why this matters beyond noise

If a real CVE/GHSA ever lands on a backend transitive, Dependabot will open a PR that is **structurally unmergeable** and red on a required check. The team would have to notice the advisory some other way and patch by hand under time pressure. This is the security-patching path being quietly broken, which is why it sits above the open P3 product bugs.

## Options

1. Point the `uv` ecosystem at `backend/pyproject.toml` / `backend/uv.lock` as the manifest rather than the export, so Dependabot writes to the real source of truth.
2. If the ecosystem cannot express that, exclude transitives from the `uv` entry (`allow: dependency-type: direct`) and rely on `scripts/audit_gate.py` + `security-baseline.json` to surface transitive advisories, with a documented manual bump procedure.
3. Add an automated post-bump step that regenerates the export from the lock so the sync gate is satisfied.

Option 1 is the correct fix if supported; option 2 is the honest fallback.

## Acceptance criteria

- [ ] A Dependabot PR for a backend **transitive** either updates `pyproject.toml`/`uv.lock` (and the regenerated export) together, or is not opened at all.
- [ ] No Dependabot backend PR lands with `Security Audit` red on the lock/export sync gate.
- [ ] If option 2 is chosen, the manual transitive-bump procedure is documented in `docs/references/quality-standards.md` (or `CLAUDE.md`), including:
      `cd backend && uv export --all-extras --no-emit-project --no-hashes --format requirements-txt -o requirements.txt`
- [ ] Verified by observing the next scheduled Dependabot run, not just by config inspection.

## Context

Found during a supply-chain triage sweep of 14 open Dependabot PRs. None of the three affected PRs carried an advisory, so nothing is currently unpatched.

frankbria:
Closes #512.

## The defect

`backend/requirements.txt` is a `uv export` of `uv.lock`. Dependabot's `uv`
ecosystem parsed it as a **manifest**, and transitive dependencies appear *only*
there — so a transitive bump produced a PR that edited the export and nothing else.

Confirmed across the 2026-08-25 sweep by the files each PR actually touched:

| PR | dependency | kind | files touched | outcome |
|----|-----------|------|---------------|---------|
| #500 pytest-asyncio | direct | pyproject.toml, uv.lock, requirements.txt | merged green |
| #502 pydantic-settings | direct | pyproject.toml, uv.lock, requirements.txt | merged green |
| #499 flask-cors | transitive | **requirements.txt only** | closed, red |
| #503 tqdm | transitive | **requirements.txt only** | closed, red |
| #505 pydantic-core | transitive | **requirements.txt only** | closed, red |

`uv sync` resolves from the lock, so those three changed nothing installed, and each
failed the lock/export sync gate. Upstream bug: `dependabot-core#13912`.

## The fix

`exclude-paths: ["requirements.txt"]` on the `uv` entry. The update job then sees
`pyproject.toml` and `uv.lock` only, so a transitive bump is either a real lockfile
bump or no PR at all. (Scope: this hides the export from update *scans*, not from the
dependency graph — see Known limitations.)

This is the issue's option 1, and it is the part that matters past the churn: an
advisory landing on a backend transitive previously produced a **structurally
unmergeable** PR on the security-patching path.

## The trade-off, left manual on purpose

A direct-dep PR now bumps `pyproject.toml`/`uv.lock` and leaves the export stale, so
the sync gate asks for one `uv export` on the branch (the failing check prints the
command; its wording is updated here, since the common case inverts).

I designed and then discarded the workflow that would have pushed that regeneration
automatically: GitHub's recursion guard means a **`GITHUB_TOKEN` push creates no new
workflow run**, so a bot-pushed commit would leave the required checks un-run on the
new head commit, needing a manual re-run or a separate token to unblock — worse than
the 30-second manual command it replaces. Documented in
`docs/references/quality-standards.md` alongside the manual transitive-bump procedure
(issue AC 3).

## Tests

`scripts/test_dependabot_config.py` (2 tests, TDD — written failing first):

1. pins the exclusion;
2. asserts every `exclude-paths` glob resolves to a real file under its entry's
   `directory`. That guards the one way this fix breaks silently — patterns are
   directory-relative, so a repo-relative `"backend/requirements.txt"` matches
   nothing and Dependabot carries on unchanged while the config looks right.

**Mutation-verified**: changing the pattern to `"backend/requirements.txt"` turns
both tests red; reverting turns them green.

The `Security Audit` step now runs `pytest scripts/` rather than naming one file, so
new script tests are picked up. It currently collects `test_audit_gate.py` (29) and
`test_dependabot_config.py` (2) — **31 passed**.

## Verification

- The documented regen command was run locally and reproduces the committed export
  byte-for-byte (`diff` clean after filtering uv's self-referential output-path
  comment, the same way CI does).
- `exclude-paths` semantics confirmed against GitHub's current options reference:
  real key, globs relative to `directory`.
- Cross-family review: **codex** — *no findings*, having independently verified the
  `exclude-paths` semantics and the `pytest scripts/` collection set. (opencode/GLM
  was attempted first and returned an `UnknownError` server error; codex is the
  documented fallback.)

## Considered and rejected

Deleting `backend/requirements.txt` outright would remove the whole failure class —
nothing in CI, the Dockerfile or the deploy installs from it. Rejected: #439 and
`DEPLOYMENT.md` keep it deliberately for tooling that cannot read a lock, and
`scripts/run-test-suite.js` / `scripts/validate-test-environment.js` read it.
Reversing that decision is a separate call, not this PR's to make.

## Measured, not asserted

`dependabot/cli` driving the published `ghcr.io/dependabot/dependabot-updater-uv`
image against this repo's `/backend` — twice, identical source tree, differing only in
the job's `exclude-paths`. Every `create_pull_request` it emitted, grouped by the files
that PR would change:

| files the PR would change | BEFORE | AFTER |
|---|---:|---:|
| `requirements.txt` only — **no-op** | **35** | **0** |
| `pyproject.toml` + `uv.lock` | 2 | 16 |
| `pyproject.toml` + `uv.lock` + `requirements.txt` | 14 | 0 |
| **total** | **51** | **16** |

The 35 include `flask-cors`, `tqdm` and `pydantic-core` — the three closed by hand on
2026-08-25. All 16 real bumps survive.

Two things this settled that I would otherwise have guessed at:

- **The issue's option 2 could not have worked.** The job already runs with
  `update-subdependencies: false` and produced those 35 anyway — Dependabot treats a
  pinned line in `requirements.txt` as a *direct* requirement of that file, ignoring
  the `# via` provenance comment. `allow: dependency-type: direct` would have changed
  nothing.
- **The trade-off is real and larger than I first estimated.** The surviving 16 touch
  `pyproject.toml` + `uv.lock` and *not* the export, so a full sweep asks for the regen
  on up to 16 PRs, not the 2 I extrapolated from the last batch.
  `open-pull-requests-limit: 5` caps how many are live at once; #517 (grouping) is the
  lever that collapses them into one PR and one regen.

Full write-up: `docs/demos/2026-08-25-issue-512-dependabot-uv-transitive-noop.md`.

## Known limitations

- **Issue AC 4 is not closed here.** It asks for confirmation from a live Dependabot
  run; the `uv` schedule is weekly, so that lands on the next sweep. The runs above are
  the closest substitute — same updater image, same repo, same job schema.
- **The manual export regeneration is new toil**, accepted for the reason above and
  sized in the table.
- **`exclude-paths` scopes update jobs, not the dependency graph.** The export still
  registers as a manifest for Dependabot *alerts* — visible today as one advisory
  reported twice, against both `backend/uv.lock` (direct) and `backend/requirements.txt`
  (unknown). Unchanged by this PR. Checking that also surfaced a genuine gap — the pip
  audit gate scans the production surface only, so a test-extra advisory can never fail
  CI — filed as **#522**.


frankbria:
## Third-party review (post-PR) — `codex`

opencode/GLM was tried first for both the pre-PR and post-PR passes and returned
`UnknownError: Unexpected server error` each time (refs `err_235ab995`, `err_0fa82efa`),
so this is the documented `codex` fallback. Disclosed rather than quietly substituted.

**Pre-PR pass** (code + config only): *no findings*. It independently confirmed
`exclude-paths` is a real key with `directory`-relative globs, and that widening the CI
step to `pytest scripts/` collects only the two intended test files.

**Post-PR pass** (full diff including docs): **2 findings, both adopted.**

---

**Medium — the `GITHUB_TOKEN` claim was too categorical.**
The rationale for not automating the export regeneration said "a `GITHUB_TOKEN` push
does not re-trigger workflows … checks that never run". The defensible version is
GitHub's recursion guard: a `GITHUB_TOKEN` push creates **no new workflow run**, so the
required checks sit un-run on the new head commit and need a manual re-run (or a
separate token) to unblock. Same conclusion, accurate premise. Reworded in
`.github/dependabot.yml`, `quality-standards.md`, `CHANGELOG.md`, the demo doc and this
PR's description.

**Minor — "Dependabot no longer sees `requirements.txt` at all" contradicted this PR's
own Known Limitations.**
`exclude-paths` scopes update **scans**, not the dependency graph; alerts still
attribute advisories to the export — which is exactly the duplicate `pytest` alert
recorded in the limitations section. The claim is now qualified everywhere it appeared:
`dependabot.yml`, `quality-standards.md`, `CHANGELOG.md`, `CLAUDE.md`, demo doc.

---

Both are wording-accuracy findings, not behaviour changes: no code, config or test
changed, and `pytest scripts/` stays at **31 passed**. Fixed in `ea6886d`.

frankbria:
## Demo — outcome evidence (Phase 11 gate)

Not "the YAML parses". The real `ghcr.io/dependabot/dependabot-updater-uv` image
(`sha256:4555a667…`), driven by `dependabot/cli` against `frankbria/auto-author`
`/backend`. **Two runs, identical source tree, differing only in the job's
`exclude-paths`.**

The updater never reads `.github/dependabot.yml` — GitHub parses that and passes the
result into the job — so putting `exclude-paths` in the job file exercises exactly the
code path this PR's config change will hit.

Every `create_pull_request` the updater emitted, grouped by the files that PR would
change:

| files the PR would change | BEFORE | AFTER |
|---|---:|---:|
| `requirements.txt` only — **no-op, changes nothing installed** | **35** | **0** |
| `pyproject.toml` + `uv.lock` | 2 | 16 |
| `pyproject.toml` + `uv.lock` + `requirements.txt` | 14 | 0 |
| **total proposed PRs** | **51** | **16** |

The 35 export-only proposals include all three that had to be closed by hand in the
2026-08-25 sweep — `flask-cors` (#499), `tqdm` (#503), `pydantic-core` (#505). All 16
real bumps survive, unchanged.

### Two things the measurement settled that I would otherwise have guessed at

**The issue's option 2 could not have worked.** The job already runs with
`update-subdependencies: false` — visible in the updater's own job dump — and produced
those 35 anyway. Dependabot treats a pinned line in `requirements.txt` as a *direct*
requirement of that file, ignoring the `# via` provenance comment `uv export` writes
above it. `allow: dependency-type: direct` would have changed nothing. Removing the
file from scope is what works.

**The trade-off is larger than I estimated from the last batch.** The surviving 16
touch `pyproject.toml` + `uv.lock` and **not** the export, so a full sweep asks for the
one-command regen on up to 16 PRs, not the ~2 I extrapolated. `open-pull-requests-limit: 5`
caps how many are live at once; #517 (grouping) is the lever that collapses them into
one PR and one regen. Recorded rather than smoothed over.

### Regression guard

```
$ uvx --with pytest --with pyyaml pytest scripts/ -q
31 passed

$ sed -i 's|"requirements.txt"|"backend/requirements.txt"|' .github/dependabot.yml
$ uvx --with pytest --with pyyaml pytest scripts/test_dependabot_config.py -q
2 failed
$ git checkout -- .github/dependabot.yml && <rerun>
2 passed
```

`exclude-paths` globs resolve against the entry's `directory`, so a repo-relative path
matches nothing and leaves the config *looking* correct while all 35 no-ops keep
coming. That is the failure the second test exists to catch.

### Not demonstrated

**Issue AC 4** asks for confirmation from a live scheduled Dependabot run. The `uv`
ecosystem is weekly, so that lands after merge. The runs above are the closest
substitute available before then.

Full write-up: `docs/demos/2026-08-25-issue-512-dependabot-uv-transitive-noop.md`.

Issue #537, opened 2026-08-26
[P2.17.1] Staging is fully down: Atlas rejects the shared MONGODB_URI, every login has failed since 2026-08-20


frankbria:
## What

Staging (https://dev.autoauthor.app) cannot authenticate to MongoDB Atlas. **Every** login fails, so the environment is effectively down and has been since **2026-08-20**, roughly 12:09Z–18:08Z UTC.

Both containers are fed the same credential — `docker-compose.yml:61` sets the frontend's `DATABASE_URL: ${MONGODB_URI}`, and `:36` sets the backend's `MONGODB_URI` from the same variable in `/opt/auto-author/.env` — so this is one broken secret taking down both services.

## Evidence

Backend readiness probe (`GET https://api.dev.autoauthor.app/api/v1/health`), three consecutive calls:

```
{"status":"unhealthy","checks":{"mongodb":"error: OperationFailure","config":"ok"}}  HTTP 503  t=3.28s
{"status":"unhealthy","checks":{"mongodb":"error: TimeoutError","config":"ok"}}      HTTP 503  t=5.24s
{"status":"unhealthy","checks":{"mongodb":"error: OperationFailure","config":"ok"}}  HTTP 503  t=4.77s
```

`config: ok` — the secrets are *present*, so this is not a missing env var. `OperationFailure` on a bare `ping` is Atlas rejecting the command (auth/authorization), not a network reach problem; an un-allowlisted IP would surface as `ServerSelectionTimeoutError`. The 3–5s latency against a sub-100ms healthy ping says it is failing at the handshake. The backend's non-DB routes are fine (`GET /api/v1/` → 200), which rules out the container being down.

Frontend auth, same cluster, same credential:

```
POST https://dev.autoauthor.app/api/auth/sign-in/email  ->  HTTP 500, empty body, t=0.56s
```

Scheduled `e2e-staging-tests.yml` runs (every 6h) — the transition is unambiguous:

| Run | Result |
|---|---|
| 2026-08-20T12:09:19Z | success |
| 2026-08-20T18:08:45Z | **failure** |
| …every run since (22 consecutive) | failure |

In the latest run ([32914166034](https://github.com/frankbria/auto-author/actions/runs/32914166034)) all 11 specs fail identically at the login step:

```
TimeoutError: page.waitForURL: Timeout 30000ms exceeded.
  55 |   await expect(page).toHaveURL(/\/dashboard/);
```

[...]

Issue #520, opened 2026-08-25
[P3.29] Retire the PM2 rollback path: two .disabled workflows, a stale DATABASE_NAME secret, and a rotation runbook that now misleads


frankbria:
> **Re-scoped 2026-09-06.** The security half of this issue is gone — see *What changed* below. What remains is dead-file cleanup, so the priority drops from P2.21 to P3.29.

## Problem

Staging cut over to containers on **2026-08-12** (#484). The PM2 path was deliberately kept as the rollback route, and #497 explicitly defers its removal:

> The `MONGODB_URI` / `DATABASE_NAME` GitHub secrets and the two `.disabled` PM2 workflows are left in place. The cutover is one day old and those workflows are the rollback path; deleting the secrets now would disarm it silently.

That reasoning was correct at the time. The problem is **where the follow-up is tracked**: #497 points at "#427 AC 8", and **#427 is closed**, so this cleanup had no open tracker.

## What changed since this was written

This issue's headline claim — *"a live Atlas credential sitting in GitHub secrets that nothing reads"* — is **no longer true**, and neither is its inventory. Verified against the API on 2026-09-06:

- **There are no repo-level `MONGODB_URI` or `DATABASE_NAME` secrets.** The repo level holds only `CLAUDE_CODE_OAUTH_TOKEN`, `TS_AUTH_SECRET`, `TS_CLIENT_ID`, `ZHIPU_API_KEY`. Both names exist solely as **`staging` environment** secrets.
- **`MONGODB_URI` is authoritative, not orphaned.** #545 added `sync-staging-env.yml`, which reads that secret and writes it to `/opt/auto-author/.env` on the box, recreating both containers and verifying the credential end-to-end. It is the *only* application secret with a live consumer. Deleting it would break the rotation path — the exact opposite of this issue's original recommendation. (#537 flagged the re-scope; the API confirms the repo-level copy it assumed never existed.)

So the "second copy of a live credential with no consumer" that justified P2 is not there. The blast-radius argument goes with it.

## What is actually left

1. **`DATABASE_NAME`** (`staging` environment, last updated 2025-11-22). Consumed only by `deploy-staging.yml.disabled:242,303,322,356`. `sync-staging-env.yml` does not write it. It is a database *name*, not a credential — no blast radius, just a stale entry that makes the secret list harder to read.
[...]

PR #545, opened 2026-08-26
feat(ci): sync MONGODB_URI from the staging secret to the box (#537)

frankbria:
Part of #537.

## Problem

Application secrets live in `/opt/auto-author/.env` **on the box**, and until now **no workflow wrote that file**. So when the rotated `MONGODB_URI` was pushed to the `staging` environment secret, GitHub dutifully updated the secret's `updated_at` — and the running containers kept the old, rejected credential.

From the outside those two states are identical. That is what made #537 expensive: the natural next thought after updating a secret is "the credential is fixed", and nothing contradicts it. Staging had been down six days.

Either the secret is wired to something, or it should not exist. This wires it.

## What it does

`workflow_dispatch` → connect over Tailscale (same pinned action and tag pair as the deploy) → write the key → `up -d --force-recreate` → **verify against MongoDB for real**.

Verification is a live `ping` from inside the running backend container plus the readiness probe — not "the file was written". `/health` returns only an exception type, so the ping is what yields a diagnosable message.

## Scope is one key, deliberately

The box's `.env` holds **26** keys; only a subset have GitHub secrets at all (`AWS_*`, `CLOUDINARY_*`, `SENTRY_DSN` and others exist only on the box). Two failure modes ruled out the obvious "render the whole file from secrets" design:

1. It would **silently delete** every key with no corresponding secret.
2. It would let a **stale** GitHub secret overwrite a newer value on the box — turning this workflow into an outage source rather than a fix for one.

So: one key, updated in place, everything else carried through byte-for-byte.

## Credential handling

- **stdin, never argv.** The remote script is the `ssh` command argument and the value arrives on stdin, so it never appears in the process list. GitHub masks secrets in logs; masking does nothing about `ps` on a box shared with six other apps.
- **Refuses rather than guesses** — empty/unset, not a `mongodb://`/`mongodb+srv://` string, or *containing a newline*. That last one matters: the remote `read -r` takes one line, so a multi-line secret would be silently truncated into a valid-looking wrong credential. Worst available failure mode; now impossible.
- **Timestamped mode-600 backup** before the write, pruned to the five most recent so credential-bearing files don't accumulate on a shared box.

## Verification

Guards, all five cases:

| input | result |
|---|---|
| empty | REFUSED |
| `hunter2` | REFUSED: not a connection string |
| multi-line | REFUSED: contains newline |
| valid `mongodb+srv://…` | ACCEPTED |
| valid `mongodb://…` | ACCEPTED |

The write logic, run against a realistic 10-key `.env` with awkward values:

- key set unchanged ✅
- every other key byte-identical ✅
- value containing `@ / ? & =` written **exactly** ✅
- mode 600 preserved on `.env` and the backup ✅

`shellcheck` on both embedded remote scripts: two info/style notes (`ls` on our own fixed-format filenames; an unused loop counter), no defects.

## Note for #520

#520 proposes **retiring** the dead `MONGODB_URI`/`DATABASE_NAME` secrets. This takes the opposite resolution to the same problem — making `MONGODB_URI` live rather than deleting it. If this merges, #520 should be re-scoped to the `.disabled` workflow cleanup and `DATABASE_NAME` alone. Flagging rather than silently invalidating it.

## Known limitations

- **Not yet run.** It needs a `workflow_dispatch` against live staging, which will change the credential on the box. Happy to dispatch it on request — that is also the fastest route to closing #537.
- **`DATABASE_NAME` is not synced.** It has not changed since 2025-11-22 and altering it would point the app at a different database. Risk without need.
- **The `staging` environment secret remains the source of truth only for this one key.** Documented in `quality-standards.md` so the asymmetry is discoverable rather than folklore.

### DIFF
--- a/.github/workflows/sync-staging-env.yml
+++ b/.github/workflows/sync-staging-env.yml
@@ -1,0 +1,175 @@
+name: Sync Staging Env
+
+# Makes the `staging` environment secret MONGODB_URI actually authoritative.
+#
+# Why this exists (#537): application secrets live in /opt/auto-author/.env on the
+# box, and until now NO workflow wrote that file. Updating the GitHub secret
+# therefore changed nothing that runs — a trap that cost a real outage its first
+# hour of debugging, because the secret's `updated_at` says the credential was
+# refreshed while the box still holds the old one. Either the secret is wired to
+# something or it should not exist; this wires it.
+#
+# Scope is deliberately ONE key. The box's .env holds 26 keys and only a subset
+# have GitHub secrets at all (AWS_*, CLOUDINARY_*, SENTRY_DSN and others exist
+# only on the box), so a whole-file render would silently delete working config.
+# Worse, a broad sync would let a *stale* GitHub secret overwrite a newer value on
+# the box — turning this workflow into an outage source. One key, updated in
+# place, everything else untouched. Add keys here only when a specific need shows
+# up, and only after checking the box's value is not the newer one.
+#
+# This does not replace deploy-staging-containers.yml. That ships images; this
+# ships one credential.
+
+on:
+  workflow_dispatch:
+    inputs:
+      recreate:
+        description: 'Recreate containers so the new value takes effect (compose injects env at container CREATION, so a plain restart keeps the old value)'
+        required: false
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+
+concurrency:
+  # Shared with the deploy workflow on purpose: both mutate /opt/auto-author and
+  # both bounce containers. Never let them interleave on one box.
+  group: deploy-staging
+  cancel-in-progress: false
+
+jobs:
+  sync:
+    name: Sync MONGODB_URI to the staging box
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    environment:
+      name: staging
+      url: https://dev.autoauthor.app
+
+    steps:
+      # Same access path as the deploy: the box firewalls :22 off the public
+      # internet except from the owner's home subnet, and the runner reaches it
+      # over Tailscale. Both tags are required — Tailscale matches the OAuth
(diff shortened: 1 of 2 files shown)


frankbria:
## Review triage — two real defects found and fixed

**The CI `review / review` check shows FAIL, but it found nothing.** Its `GLM bug-hunting review` step ran 20m19s and was **cancelled** on an internal timeout — `glm-review.yml` sets no `timeout-minutes`, and the posted comment's checklist is entirely unchecked. So it never produced findings. Not a signal about this diff, but also not review coverage, so I ran the documented fallback instead.

### 1. `IMAGE_TAG` unset — `recreate` would always fail *(codex, P1)*

`docker-compose.staging.yml` pins both images to `${IMAGE_TAG:?IMAGE_TAG is required}`. `deploy-staging-containers.yml` only exports it inside its own SSH session; nothing on the box persists it. Confirmed against the live box:

```
config FAILS without IMAGE_TAG - codex is right
config OK with IMAGE_TAG
```

The credential would have been written and then never applied — precisely the failure this workflow exists to prevent. Fixed by deriving the tag from the **running** container (`sha-de87eea`) rather than from `main`, so syncing a credential can never move staging onto a different build as a side effect.

### 2. Write-then-verify could take the whole site down *(mine, on re-read)*

`docker-compose.yml` declares the frontend `depends_on: backend: condition: service_healthy`. If the refreshed credential were *also* bad, `up -d --force-recreate` would wait on a backend that can never pass its healthcheck, abort, and leave the **frontend** down — converting "login broken, site up" into "site down", strictly worse than the state being repaired.

Reordered to **pre-flight first**: prove the new credential against the live cluster using the container that already exists, writing nothing. A bad secret now costs one failed run and changes nothing. The candidate value reaches it over stdin via `docker exec -i` rather than `docker exec -e`, keeping it out of argv on a box shared with six other apps.

Both in `52f217e`.

Issue #544, opened 2026-08-26
[P2.20.1] Public repo discloses staging host, DB username and cluster hostname in four committed files


frankbria:
## Problem

This repository is **public**, and several committed files disclose staging infrastructure identifiers that should not be in it. Deliberately not restated here — see the locations:

| File | What it discloses |
|---|---|
| `.github/DEPLOYMENT.md:56` | staging host address |
| `docs/demos/2026-07-10-issue-189-loopback-bind.md:20,75` | staging host address (×4 in command examples) |
| `docs/STAGING_VERIFICATION_RESULTS.md:51` | database username **and** managed-cluster hostname, together in one connection string |
| `docs/BACKEND_TEST_FAILURE_ANALYSIS.md:24` | managed-cluster shard hostname |

## Severity: real, but not a breach

Calibrating honestly rather than alarming:

- **No passwords are exposed.** The connection string in `STAGING_VERIFICATION_RESULTS.md` is masked.
- **The host address is arguably already public** — `dev.autoauthor.app` resolves to it, so that disclosure is close to zero marginal risk on its own.
- **The username + cluster hostname pair is the part that matters.** It is half a credential and it names a specific principal to attack. Combined with the cluster hostname it removes the discovery step from a credential-stuffing or password-spray attempt against that database.

Residual risk is bounded by the database's own IP allowlist and password strength, not by anything in this repo.

## Open question worth answering before closing

The password for the exact database user named in `STAGING_VERIFICATION_RESULTS.md:51` was found rejected on 2026-08-26 (`bad auth : authentication failed`, AtlasError 8000 — see #537) and has since been rotated. **Was that rotation routine, or a response to something?** If routine, this issue stays P2 hygiene. If not, it escalates and the exposure above becomes materially more interesting. Nobody should assume the benign answer without checking the cluster's access logs.

## Important: scrubbing the working tree is not sufficient

These identifiers are in **git history**. Deleting the lines in a new commit leaves every value retrievable from any clone or from GitHub's commit views. The two honest options:

1. **Rotate and accept** *(recommended)* — treat the disclosed identifiers as burned, ensure the credential is rotated (already done for the database user), scrub the working tree so it stops getting worse, and move on. History stays as-is.
2. **Rewrite history** — `git filter-repo` over the affected paths, force-push, and every clone and open PR breaks. On a public repo the values may already be mirrored, forked, or indexed, so this buys less than it costs.

**Do not rewrite history without the maintainer explicitly asking for it.**

[...]

Issue #211, opened 2026-07-04
[P3.5] Guard test-DB teardown against non-test URIs (data-loss risk)


frankbria:
**Problem:** `conftest.py` runs `drop_database`/`collection.drop()` in setup/teardown with the target derived purely from `TEST_MONGO_URI` and no assertion that it's local or `*test*`. An accidental Atlas URI (the project's Mongo is Atlas; the repo has a history of stray env vars) silently wipes real users/books/audit_logs.

**Evidence:** `backend/tests/conftest.py:104,131,289-291,412-414`.

**Acceptance criteria:** at import time, assert the resolved DB host is local and/or the name matches `*test*` before any drop; raise otherwise.

PR #553, opened 2026-08-27
fix(security): scrub live staging identifiers from tracked docs (#544)

frankbria:
Addresses #544 (AC2 + a guard). **AC1 and AC3 need Atlas console access and are left open** — see below.

## What was scrubbed

All four locations the issue named, replaced with `<staging-host>`, `<db-user>`, `<cluster-host>`:

| file | disclosed |
|---|---|
| `.github/DEPLOYMENT.md` | staging host address |
| `docs/demos/2026-07-10-issue-189-loopback-bind.md` | staging host address ×4 |
| `docs/STAGING_VERIFICATION_RESULTS.md` | db username **and** managed-cluster hostname, in one string |
| `docs/BACKEND_TEST_FAILURE_ANALYSIS.md` | cluster shard hostname |

Every substitution keeps the surrounding command or table readable.

## I swept the whole tracked tree rather than trusting the list — and the list was right

Worth recording so nobody re-runs it. The raw sweep flagged ~99 tracked lines, which looks alarming and isn't:

- **88 of them are the *previous* provider's now-dead IP** (retired in the #537 move), across `archive/`, `claudedocs/`, `frontend/claudedocs/`, `docs/` and `.beads/issues.jsonl`. A dead host's address is not a disclosure. (Distribution corrected after review — the note exists so nobody re-runs the sweep, so a wrong inventory defeats its own purpose.)
- **The current host appears in exactly the two files above** — 5 occurrences.
- **Two apparent "username + hostname" hits were documentation placeholders**: `staging_user:pa%24%24word@` in `docs/DATABASE_CONNECTION_STANDARD.md`, and an uppercase `USER:REDACTED@cluster0.abcde.mongodb.net` in the #211 demo.
- **`backend/.env` is untracked, gitignored, and has never been committed** (`git log --all -- backend/.env` → 0 commits). I checked this first, because a committed `.env` would have been a different and much worse issue.

So #544's triage was better than my sweep's first pass. The one thing the sweep adds is the dead-IP count, which is a reason *not* to widen the scrub, not a reason to.

## The guard

`scripts/test_no_staging_identifiers.py`, in the suite CI already runs (`pytest scripts/`). Mutation-checked: RED when a real cluster hostname is reintroduced, GREEN when it is not.

Deliberately narrow, because a noisy rule gets deleted — this repo already dropped `detect-private-key` from pre-commit for false positives:

- **Matches `*.mongodb.net` hostnames**, which have no legitimate reason to appear literally in a public repo. Known examples are allowlisted **by exact value**, so a real cluster appearing in those same files still fails.
- **A second test fails if an allowlist entry stops matching anything**, so a stale entry cannot sit there silently permitting one hostname forever. This shipped **broken** in the first commit — `git grep` matched the `ALLOWED_HOSTS` literal in the guard file itself, so it could never fail. It passed for exactly as long as the file was untracked, which is why the bug survived its own first run. Caught by the cross-family review, which proved it by deleting the hostname from both referencing files; fixed in `625bd72` and mutation-checked in both directions.
- **Does NOT match public IPv4.** #544 itself calibrated the host IP as close to zero marginal risk (`dev.autoauthor.app` resolves to it), and such a rule would fire on version strings, RFC 5737 test addresses, and those 88 dead-host lines.

### One attempt reverted

I first added a `check-secrets.sh` pattern matching Mongo URIs whose userinfo isn't a placeholder. It false-positived on **both** of the repo's own documented examples (case-sensitivity missed `USER:`, and `staging_user` isn't a form the regex knew). That is exactly how the previous hook got removed, so I reverted it rather than tuning a fragile regex.

## Remediation path — Option 1, "rotate and accept"

Per the issue's own framing. These values are in git history; scrubbing the working tree stops it getting worse but does not retract them. History rewriting is **not** done here — the issue says not to without the maintainer explicitly asking, and on a public repo the values may already be mirrored or indexed.

## Also found, filed separately

The review surfaced operator-**network** identifiers in three docs — a home dynamic IP named alongside the ISP, and a `/24` granted SSH in a `ufw` rule. Different category from #544's staging infrastructure, and lower severity (stale, dynamic, no credential value), so filed as **#554** rather than widened into this PR.

## Left open — needs Atlas console access

- [ ] **AC1** — was the 2026-08-26 credential rotation routine, or a response to something? The issue is explicit that nobody should assume the benign answer. This gates whether #544 stays P2 hygiene or escalates.
- [ ] **AC3** — confirm the cluster's network access list still restricts connections to the expected sources.

I can't reach the Atlas console, so **#544 should stay open after this merges** until those two are answered.

### DIFF
--- a/.github/DEPLOYMENT.md
+++ b/.github/DEPLOYMENT.md
@@ -53,7 +53,7 @@ same name feeds only the disabled PM2 path.
 | `SSH_KEY` | Private SSH key for server access | `-----BEGIN OPENSSH PRIVATE KEY-----…` |
 | `USER` | SSH username | `root` |
 | `STAGING_TS_HOST` | Tailscale hostname for the box; preferred over `HOST` | `staging-vps` |
-| `HOST` | Public hostname or IP — fallback when `STAGING_TS_HOST` is unset | `195.35.14.177` |
+| `HOST` | Public hostname or IP — fallback when `STAGING_TS_HOST` is unset | `<staging-host>` |
 | `TS_CLIENT_ID` | Tailscale OAuth client id — the deploy runs over Tailscale (#485, #489) | — |
 | `TS_AUTH_SECRET` | Tailscale OAuth client secret; needs the client's full tag set (#490) | — |
 
--- a/docs/BACKEND_TEST_FAILURE_ANALYSIS.md
+++ b/docs/BACKEND_TEST_FAILURE_ANALYSIS.md
@@ -21,7 +21,7 @@
 ### Error Pattern
 ```
 pymongo.errors.ServerSelectionTimeoutError: SSL handshake failed:
-ac-gdkptel-shard-00-XX.oxzhocn.mongodb.net:27017: [Errno 104] Connection reset by peer
+<cluster-host>:27017: [Errno 104] Connection reset by peer
 ```
 
 ### Affected Tests (12 in session service + 2 question generation)
--- a/docs/STAGING_VERIFICATION_RESULTS.md
+++ b/docs/STAGING_VERIFICATION_RESULTS.md
@@ -48,7 +48,7 @@ All critical infrastructure verified and operational:
 
 **MongoDB Connection**: Atlas (Cloud)
 ```bash
-DATABASE_URL=mongodb+srv://frankbria:***@briastrategygroup.oxzhocn.mongodb.net/
+DATABASE_URL=mongodb+srv://<db-user>:<redacted>@<cluster-host>/
 DATABASE_NAME=auto_author-staging
 ```
 
--- a/docs/demos/2026-07-10-issue-189-loopback-bind.md
+++ b/docs/demos/2026-07-10-issue-189-loopback-bind.md
@@ -17,7 +17,7 @@ LISTEN 0      511                              *:3002             *:*
 The host firewall (ufw, default-deny) already drops external 8000/3002 — the AC "OR" branch — but nothing in the repo documented it. From an off-box machine, direct port access times out while the nginx-proxied HTTPS endpoints serve normally:
 
 ```bash
-curl -sm 5 http://195.35.14.177:8000/api/v1/health -o /dev/null -w "direct 8000: %{http_code}\n" || echo "direct 8000: TIMEOUT/UNREACHABLE"; curl -sm 5 http://195.35.14.177:3002 -o /dev/null -w "direct 3002: %{http_code}\n" || echo "direct 3002: TIMEOUT/UNREACHABLE"; curl -sm 10 https://api.dev.autoauthor.app/api/v1/health -o /dev/null -w "via nginx api: %{http_code}\n"; curl -sm 10 https://dev.autoauthor.app -o /dev/null -w "via nginx app: %{http_code}\n"
+curl -sm 5 http://<staging-host>:8000/api/v1/health -o /dev/null -w "direct 8000: %{http_code}\n" || echo "direct 8000: TIMEOUT/UNREACHABLE"; curl -sm 5 http://<staging-host>:3002 -o /dev/null -w "direct 3002: %{http_code}\n" || echo "direct 3002: TIMEOUT/UNREACHABLE"; curl -sm 10 https://api.dev.autoauthor.app/api/v1/health -o /dev/null -w "via nginx api: %{http_code}\n"; curl -sm 10 https://dev.autoauthor.app -o /dev/null -w "via nginx app: %{http_code}\n"
 ```
 
 ```output
@@ -72,7 +72,7 @@ LISTEN 0      2048                     127.0.0.1:8000       0.0.0.0:*
 The app still works end-to-end through nginx (TLS + real response bodies), deploy health checks on localhost keep working, and direct external port access remains dead:
 
 ```bash
-echo "--- via nginx (off-box) ---"; curl -sm 10 https://api.dev.autoauthor.app/api/v1/health; echo; curl -sm 15 https://dev.autoauthor.app -o /dev/null -w "frontend via nginx: %{http_code}\n"; echo "--- deploy health checks (on-box, localhost) ---"; ssh staging 'curl -sf http://localhost:8000/api/v1/health >/dev/null && echo "localhost:8000 health OK"; curl -sf http://localhost:3002 -o /dev/null && echo "localhost:3002 OK"'; echo "--- direct external (must stay unreachable) ---"; curl -sm 5 http://195.35.14.177:8000/api/v1/health -o /dev/null -w "direct 8000: %{http_code}\n" || echo "direct 8000: TIMEOUT/UNREACHABLE"; curl -sm 5 http://195.35.14.177:3002 -o /dev/null -w "direct 3002: %{http_code}\n" || echo "direct 3002: TIMEOUT/UNREACHABLE"
+echo "--- via nginx (off-box) ---"; curl -sm 10 https://api.dev.autoauthor.app/api/v1/health; echo; curl -sm 15 https://dev.autoauthor.app -o /dev/null -w "frontend via nginx: %{http_code}\n"; echo "--- deploy health checks (on-box, localhost) ---"; ssh staging 'curl -sf http://localhost:8000/api/v1/health >/dev/null && echo "localhost:8000 health OK"; curl -sf http://localhost:3002 -o /dev/null && echo "localhost:3002 OK"'; echo "--- direct external (must stay unreachable) ---"; curl -sm 5 http://<staging-host>:8000/api/v1/health -o /dev/null -w "direct 8000: %{http_code}\n" || echo "direct 8000: TIMEOUT/UNREACHABLE"; curl -sm 5 http://<staging-host>:3002 -o /dev/null -w "direct 3002: %{http_code}\n" || echo "direct 3002: TIMEOUT/UNREACHABLE"
 ```
 
 ```output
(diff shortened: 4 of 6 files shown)


frankbria:
## Cross-family review — opencode (zai/glm-5.2), pre-merge pass

Read-only (diff inline, no `--auto`), and instructed not to echo identifier values. It verified independently rather than diff-reading: re-ran the whole-tree sweep itself, cross-checked tracked IPs against DNS, and mutation-tested the guard in an isolated worktree — including restoring the pre-scrub file contents to confirm the guard would have caught the original #544 disclosure.

Verdict `REQUEST_CHANGES` on one Major. Fixed in `625bd72`; triage below the review.

CI claim verified: `tests.yml:202` runs `uvx --with pytest --with pyyaml pytest scripts/ -q`. Now the independent sweep. I'll write a masking script so no identifier value is echoed:No live cluster hostname survives, and no tracked IPv4 currently resolves from `dev.autoauthor.app`. Now I need to classify the remaining IPv4s and locate them (masked), plus sanity-check DNS actually resolved:All IPv4 noise is classified: 88-occurrence IP is the archived/decommissioned provider host, the rest are SVG path coordinates, incident-doc mentions, and a log excerpt. Notably, at least two of the 88 live in `.beads/`, which the PR body's distribution claim omits. Let me confirm distributions and the allowlist references, then mutation-test the guard in an isolated worktree:Public resolvers disagree with local getent (likely sandbox interception) — I need to verify against resolver answers, not getent. Also: the dead IP appears in `.beads/` and `docs/` too, contradicting the PR body's distribution claim. Setting up the isolated worktree for guard mutation tests:Public-resolver DNS is blocked in this sandbox (local getent only — weak-positive evidence, will caveat). Worktree ready. Running the guard baseline + 4 mutations:Guard catches the original #544 disclosure (both locations). Now variants, the skip-prefix hole, and the allowlist-vacuity test:All four variant forms caught (2-label, 3-label shard, uppercase, srv URI). Now the skip-prefix hole and the allowlist-vacuity test:Unexpected — need the actual failure detail. Investigating what got staged and what the guard flagged:My earlier checkout left mutation-A's dirty files in place — the failure came from those, not the reports. More importantly: both skip-prefix dirs are **gitignored** with **0 tracked files**, so `SKIP_PREFIXES` currently guards nothing. Redoing C cleanly (force-add to simulate) and running D:Skip-prefix blind spot confirmed (latent — dirs currently gitignored). Now the allowlist-vacuity test — remove the allowlisted hostname from its two referencing files:Vacuity confirmed — the second test passes even with zero real references left; `git grep` self-matches the `ALLOWED_HOSTS` literal in the guard file itself. Final completeness checks (counts only, no values echoed), then reading the scrubbed spots in context:# Adversarial review — PR #553

## 1. Completeness (independent sweep)

I re-ran the sweep myself: `mongodb.net` hostnames at any label depth, mongodb URIs with userinfo, all public IPv4s cross-checked against DNS, Atlas org/shard substrings, and the DB username in URI context. **No live staging identifier survives in the tracked tree.** Only synthetic doc examples remain (`cluster.mongodb.net`-style placeholders, and the allowlisted fixture value in its two legit homes at `backend/tests/test_db_guard.py:17` and `docs/demos/2026-07-19-issue-211-test-db-teardown-guard.md:9`). The current host IP is at 0 occurrences (was 5). Caveat: public resolvers are blocked in this sandbox; DNS confirmation is via local resolver only (no tracked IP matched). The four issue-named locations are all scrubbed.

## Findings

- **Major** — `scripts/test_no_staging_identifiers.py:76-86` — `test_allowlisted_hosts_are_still_referenced` is vacuous: `git grep -lF <host>` always matches the `ALLOWED_HOSTS` literal in the guard file itself. **Proven**: I removed the hostname from both referencing files (occurrence count 0) and the test still passes. Failure scenario: the fixture/demo is later deleted or reworded → the stale entry silently whitelists that exact hostname forever — precisely the latent hole the test documents itself as preventing. It also makes the PR body claim "it caught two entries I had guessed at" untrue of the code as shipped (a manual pre-add grep may have caught them; this test cannot). Fix: exclude the guard file from the pathspec (`':!scripts/test_no_staging_identifiers.py'`) or filter it out of `hits`.
- **Suggestion** — `scripts/test_no_staging_identifiers.py:39` — `SKIP_PREFIXES` guards two dirs that are gitignored with zero tracked files (`.gitignore:47,49`), i.e. dead code; **proven** that force-adding a report containing a real cluster hostname passes the guard. Failure scenario: someone force-adds a Playwright trace (HTML artifacts plausibly embed backend error bodies with connection strings) and the guard is blind to exactly the artifact type most likely to leak one. Drop the prefixes or comment that they only matter for force-added reports.
- **Suggestion** — `docs/INCIDENT-2025-10-19-firewall-lockout.md:21,46`, `docs/fixes/2025-11-22-clerk-webhook-nginx-fix.md:20`, `tasks/lessons.md:112` — these retain non-staging *personal-network* IPs (an operator ISP-dynamic IP, a UFW allow-rule source prefix, log-excerpt client IPs). Out of #544's declared scope (not staging infra, and stale/dynamic), but they are real identifiers of the operator's network in a public repo; the PR's "~99 lines" sweep dismissed all IPv4 noise wholesale. Failure scenario: low-level doxxing/attack-surface mapping of the operator's home/office network despite a "scrubbed" security PR.
- **Suggestion** — PR body — the dead-IP distribution "across `archive/`, `claudedocs/` and `frontend/claudedocs/`" is incomplete: the 88 also span `.beads/issues.jsonl` and `docs/` (`docs/INCIDENT-2025-10-19-firewall-lockout.md`, `docs/STAGING_DEPLOYMENT_ANALYSIS.md`). Conclusion unaffected (all decommissioned-host refs), but the note exists specifically "so nobody re-runs it" — a wrong inventory defeats its own purpose.
- **Nitpick** — `docs/STAGING_VERIFICATION_RESULTS.md:52` — the database name is kept in the same block where the user/host were scrubbed; harmless (no credential value, out of the issue's four locations) but it is one more staging identifier a future sweep will trip over.
- **Nitpick** — `docs/STAGING_VERIFICATION_RESULTS.md:51` — password mask changed `***` → `<redacted>`, a third masking vocabulary alongside `<db-user>`/`<cluster-host>`; self-explanatory, but `***` was already non-identifying and consistent with the file's original style.

## Guard verification (priority 2, for the record)

Regex is sound: catches 2-label, 3-label shard (suffix-anchored), uppercase (`re.I`), hyphens, srv and `:27017` standard forms; boundaries hold against `mongodb.network`-style suffixes. **Mutation-tested**: restoring the pre-scrub file contents makes it FAIL at exactly the two original #544 locations — it would have caught the original disclosure; four synthetic variants all caught. CI wiring verified: `.github/workflows/tests.yml:202` runs `pytest scripts/`. The primary test is not trivially defeatable by any realistic accidental-reintroduction path; the defects above are the allowlist's missing integrity check and the latent skip-prefix blind spot.

## Scrub integrity (priority 3) and placeholders (priority 5)

No command, table, or example reads wrong: the DEPLOYMENT table row is well-formed, the error-quote and demo transcripts remain coherent, and the placeholder set matches the issue's AC vocabulary exactly (`<staging-host>` VPS vs `<cluster-host>` DB — distinct, not confusable). Remaining `ssh staging` / `staging-vps` are Tailscale-internal names, not publicly resolvable. No breakage found.

## Verdict

**REQUEST_CHANGES** — the scrub itself is complete and verified, but the guard's allowlist-integrity test is provably non-functional (self-matching `git grep`); fix that one test (and ideally resolve the two Suggestion items) before merge.

frankbria:
## Triage

Every finding verified against the repo before acting.

| # | Severity | Finding | Verified | Action |
|---|---|---|---|---|
| 1 | **Major** | `test_allowlisted_hosts_are_still_referenced` is vacuous — `git grep -lF` self-matches the `ALLOWED_HOSTS` literal | ✅ `git grep -lF` returns 3 files including the guard | **Fixed** in `625bd72` via `:!<path>`; mutation-checked — now FAILS when the host is unreferenced |
| 2 | Suggestion | `SKIP_PREFIXES` is dead code and a blind spot for force-added Playwright reports | ✅ both dirs gitignored, 0 tracked files | **Fixed** — skip removed entirely; nothing is skipped but the guard file |
| 3 | Suggestion | Operator-network IPs remain (home dynamic IP + ISP, a `/24` granted SSH) | ✅ confirmed | **Filed as #554** — different category from #544's staging infra; deliberately not widened into this PR |
| 4 | Suggestion | PR body's dead-IP distribution omits `.beads/` and `docs/` | ✅ | **PR body corrected** |
| 5 | Nitpick | DB name kept in the same block | Correct, and deliberate — no credential value, outside the issue's four locations | Skipped |
| 6 | Nitpick | `<redacted>` is a third masking vocabulary alongside `***` | Fair | Skipped — `***` reads as "a password was here", `<redacted>` matches the `<...>` set used for the other two substitutions in the same line |

### On finding 1 — the reviewer also caught me overclaiming

Its note that "it caught two entries I had guessed at" is **untrue of the code as shipped** is correct, and worth stating plainly. That check did fail for me, but only because the guard file was still **untracked** at that moment, so `git grep` could not see its own literal. The moment it was committed, the test became unfailable. The bug survived its own first run, and I credited it in the PR body for work it could not do.

### On finding 2

The reviewer's framing is better than my original: I skipped those directories as "noise", but a Playwright report HTML can embed a backend error body verbatim — which is precisely where a connection string would surface. Skipping nothing is both simpler and safer.

Guard is green, mutation-checked in both directions. Restores were done from a backup copy rather than `git checkout`, after that exact mistake reverted an uncommitted scrub earlier in this session.

Issue #556, opened 2026-08-27
[P0.10] better-auth 1.7 on main locks out every pre-1.7 account — total auth outage on deploy


frankbria:
## What

`better-auth` 1.6.26 → **1.7.1** (#543, commit `8c796a9`) is on `main`. Deploying it locks **every account created on ≤1.6 out of the application**. Sign-in returns `401 INVALID_EMAIL_OR_PASSWORD` and the server logs `WARN [Better Auth]: User not found` — while the user row is plainly still there, since sign-up rejects the same address with `422 USER_ALREADY_EXISTS_USE_ANOTHER_EMAIL`.

Accounts created *on* 1.7 sign in normally. The break is specific to credential records written by an earlier version.

Found by #550's staging verification. Staging is **already rolled back** to `sha-bae440a` and green; this issue is about `main`.

## Reproduction — three points on one account

`frank.bria+aa550-before2@pm.me`, created on staging at 15:48Z under 1.6.26, untouched between calls:

| stage | tag | better-auth | `POST /api/auth/sign-in/email` |
|---|---|---|---|
| before deploy | `sha-bae440a` | 1.6.26 | **200** |
| after deploy | `sha-8c796a9` | **1.7.1** | **401** |
| after rollback | `sha-bae440a` | 1.6.26 | **200** |

And the converse, same database:

| account | created on | on 1.7 | on 1.6 |
|---|---|---|---|
| created under 1.6.26 | 1.6.26 | ✘ 401 | ✓ 200 |
| created under 1.7.1 | 1.7.1 | ✓ 200 | ✓ 200 |

**The incompatibility is one-way** — 1.6 reads what 1.7 wrote, but 1.7 cannot read what 1.6 wrote. So this is not an ordinary forward migration; it is 1.7's credential lookup failing to match existing records.

Corroborating, at suite scale: staging E2E was green on `sha-bae440a` at 15:43–15:48Z ([33089344716](https://github.com/frankbria/auto-author/actions/runs/33089344716)), then **11 of 11 specs failed** on `sha-8c796a9` ([33090040109](https://github.com/frankbria/auto-author/actions/runs/33090040109)) — all in the shared login fixture, first attempt and both retries — and returned to `1 flaky, 10 passed` after rollback ([33092235853](https://github.com/frankbria/auto-author/actions/runs/33092235853)).

## Why this is P0

- It is a **total auth outage for the entire existing user base** the moment `main` is deployed anywhere.
[...]

Issue #539, opened 2026-08-26
[P2.19.1] Verify the first post-#517 Dependabot sweep actually groups and self-closes nothing


frankbria:
## What

#517 landed grouping (`.github/dependabot.yml`) and enabled repo-wide auto-merge in PR #538. Three of its four acceptance criteria were demonstrated with outcome evidence in [`docs/demos/2026-08-26-issue-517-dependabot-grouping.md`](../blob/main/docs/demos/2026-08-26-issue-517-dependabot-grouping.md). **AC3 was not, and structurally could not be.**

> - [ ] A batch of >= 3 dependency PRs lands without any being self-closed by Dependabot.

Proving that needs a real Dependabot sweep against the merged config. All three ecosystems are on `schedule: weekly`, so the first one runs after #538 merged (2026-08-26).

## Why it is worth actually checking

The self-close in #517 was the expensive part, not the CI cost. Dependabot closed #504 two minutes after #498 merged, claiming `docker/setup-buildx-action` was "up-to-date now" while `main` still pinned `@v3`, then deleted the branch so `gh pr reopen` failed and the bump had to be recreated as #515. A self-closed PR is indistinguishable from an intentionally-declined one.

Grouping should remove the collision at its source — one PR per ecosystem per sweep cannot race itself — but that is a prediction from a local harness, not an observation of GitHub's scheduler.

## Acceptance criteria

- [ ] The first post-#517 weekly sweep is observed; record how many PRs each ecosystem actually opened.
- [ ] Counts are consistent with the measured harness result (`uv` 13 → 4, `npm` ≥33 → 6, `github-actions` → 1 group). Investigate any material divergence — in particular whether `open-pull-requests-limit: 5` truncates the ungrouped major residue.
- [ ] Run the post-batch audit now documented in `docs/references/quality-standards.md` → **Dependency Batches** → *After every batch*, and confirm nothing was closed-but-unmerged while still stale on `main`.
- [ ] Confirm the npm indirect-dep residue matches what the demo predicts (standalone indirect bumps still arrive solo), or record how it differs.
- [ ] Exercise `gh pr merge --auto` on a grouped PR that can actually merge, closing the loop on #517 AC2 in production rather than the arm/disarm test.

## Commands

```bash
# What did the sweep open?
gh pr list --author "app/dependabot" --state open --json number,title \
  --jq '.[] | "#\(.number) \(.title)"'

# What did it close without merging? (the self-close check)
[...]

Issue #514, opened 2026-08-25
[P3.25] TypeScript 7 upgrade blocked by ts-jest peer range


frankbria:
## Problem

Dependabot proposes `typescript` 5.9.3 → 7.0.2 (#509, closed). It **cannot install**:

```
npm error ERESOLVE could not resolve
npm error While resolving: ts-jest@29.4.12
npm error Found: typescript@7.0.2
npm error Could not resolve dependency:
npm error peer typescript@">=4.3 <7" from ts-jest@29.4.12
```

`ts-jest@29.4.12` caps TypeScript below 7. Both `Frontend Tests` and `Build frontend image` fail at the install step, before any compilation. Not flaky, not rebaseable.

### Second, independent cap: `@typescript-eslint` 8.x (added from #555)

`ts-jest` is not the only blocker. Once `@typescript-eslint` moves 6.x -> 8.x — which
the `frontend-dev` group proposes — it caps TypeScript below 6.1:

```
npm error While resolving: auto-author-frontend@0.1.0
npm error Found: typescript@7.0.2
npm error Could not resolve dependency:
npm error peer typescript@">=4.8.4 <6.1.0" from @typescript-eslint/parser@8.68.0
npm error   dev @typescript-eslint/parser@"^8.67.0" from the root project
npm error   peer @typescript-eslint/parser@"^8.68.0" from @typescript-eslint/eslint-plugin@8.68.0
```

Measured 2026-08-27 against #548's manifest with `npm install --package-lock-only`.
This one is tighter than `ts-jest`'s (`<6.1.0` vs `<7`) and it is the cap that actually
fires first. Swapping the Jest transform therefore does **not** unblock TS 7 on its own —
`typescript-eslint` must also ship a TS 7 compatible major, or the ESLint TypeScript
plugin has to go.

## Additional consideration

TypeScript 7 is the Go-native port. The bump pulls **20 new packages** into the lockfile — `@typescript/typescript-{linux-x64,darwin-arm64,win32-x64,linux-arm64,…}` — per-platform native binaries. Expected for TS 7, but a real change in what executes during build, and worth auditing deliberately rather than accepting via an automated bump.

## Blocked on

**Both** of:

1. `ts-jest` shipping a TypeScript 7 compatible major, **or** migrating the Jest transform to an alternative (`@swc/jest`, or Next.js's built-in SWC transform) that does not gate on the TS version.
2. `typescript-eslint` shipping a major whose peer range admits TypeScript 7.

[...]

Issue #513, opened 2026-08-25
[P3.24] Migrate to Tailwind CSS v4 together with tailwind-merge v3


frankbria:
## Problem

`frontend/package.json` pins `"tailwindcss": "^3.3.0"` and `"tailwind-merge": "^2.1.0"`. Dependabot keeps proposing `tailwind-merge` 3.x (#510, closed), which **cannot be adopted independently**: v3.0.0 drops Tailwind CSS v3 support entirely.

Upstream is explicit. The v3.0.0 release notes: *"This release drops support for Tailwind CSS v3 in favor of support for Tailwind CSS v4. That means you should upgrade to Tailwind CSS v4 and tailwind-merge v3 together."* The v3.6.0 README: *"Supports Tailwind v4.0 up to v4.3 (if you use Tailwind v3, use tailwind-merge v2.6.0)."*

## Why the test suite will not protect this migration

`twMerge` failures are **silent**. Nothing in the frontend suite asserts the *output* of a merge, so a wrong result produces no test failure and no type error — only degraded styling at runtime. PR #510 had Frontend Tests and E2E Playwright fully green while being fundamentally incompatible. Any migration work here needs verification that does not lean on the existing suite.

## Scope

Paired upgrade, done together:
- `tailwindcss` 3.3 → 4.x (CSS-first config, `@theme` variables, PostCSS/Vite plugin change, removed utilities)
- `tailwind-merge` 2.6 → 3.x

`tailwind-merge` v3 breaking changes to work through:
- Theme scale keys now match the Tailwind v4 theme-variable namespace exactly
- `isLength` validator removed, split into `isNumber` and `isFraction`
- Config prefix must no longer include the combining `-`
- Tailwind v3 prefix *position* in a class no longer supported
- Custom separators no longer supported
- New mandatory `orderSensitiveModifiers` property when using `createTailwindMerge`
- Classes removed in Tailwind v4 are no longer recognised

## Acceptance criteria

- [ ] `tailwindcss` on 4.x and `tailwind-merge` on 3.x, upgraded in a single change.
- [ ] Every `twMerge`/`cn` call site audited against the breaking list above; any `createTailwindMerge` config supplies `orderSensitiveModifiers`.
- [ ] **Add unit tests asserting `twMerge` output** for the conflict cases the app actually relies on — the gap that let #510 go green.
- [ ] Visual verification of the main flows (dashboard, TOC wizard, chapter editor, export) via the `conducting-demo` skill; screenshots before/after.
[...]

PR #566, opened 2026-08-28
[#555] fix(deps): keep shipping build tools out of the dev group's majors

### LINKED ISSUE #555: [P2.19.2] frontend-dev group bumps tailwindcss majors on a false premise, and one blocked major strands twelve fine updates
Found while clearing the Dependabot queue after #534. **#548** is structurally unmergeable, and the reason is a flaw in the grouping added by #517 rather than anything wrong with the bumps.

## What happened

`.github/dependabot.yml` exempts the `frontend-dev` group from the no-majors rule:

```yaml
frontend-dev:
  # Dev deps never ship, so a major is a CI concern at worst.
  dependency-type: "development"
  patterns: ["*"]
```

So #548 arrived carrying **14 updates including five majors**, two of which are **already triaged as blocked**:

| bump | status |
|---|---|
| `typescript` 5.9.3 → 7.0.2 | closed once as #509, blocked per **#514** (`ts-jest@29` peers `typescript ">=4.3 <7"`) |
| `tailwindcss` 3.4.19 → 4.3.3 | blocked per **#513** (needs `tailwind-merge` v3 together) |

It fails at `npm ci`, before any compilation — and there is now a **second** independent cap on TypeScript 7 that #514 does not record:

```
peer typescript@">=4.8.4 <6.1.0" from @typescript-eslint/eslint-plugin@8.67.0
Found: typescript@7.0.2
```

So TS 7 is blocked by `ts-jest` *and* by `typescript-eslint` 8.x.

## Problem 1 — the group's justifying premise is false for `tailwindcss`

"Dev deps never ship" is true of eslint and jest. It is **not** true of `tailwindcss`, which sits in `devDependencies` but generates the CSS bundle every user downloads. `postcss` and `autoprefixer` are the same shape.

`dependency-type: "development"` is an npm-manifest classification, not a statement about what reaches production. Build-time toolchains that emit shipped artifacts are dev-classified but user-facing, and #517's exemption treats the two as identical.

This is the same argument #517 itself made for keeping production majors solo — "grouped PRs are cheap to merge and expensive to review or revert, which is the wrong trade for a bump that reaches users." Tailwind v3 → v4 is exactly such a bump; it is only in the batch because npm calls it a dev dependency.

## Problem 2 — one blocked major takes the whole group hostage

Twelve of the fourteen updates are unobjectionable (`@playwright/test`, `@types/react`, `postcss`, `autoprefixer`, `user-event`, …). None can merge, because two majors in the same PR cannot install. Dependabot will keep regenerating this PR in the same broken state on every sweep.

That is a variant of the failure #517 was created to fix. #517 removed the *rebase storm*; this is the same lossiness arriving as a permanently-red batch that hides a dozen fine updates behind two known blockers.

## Acceptance criteria

- [ ] `tailwindcss` (and any build-time toolchain that emits shipped artifacts — `postcss`, `autoprefixer`) no longer receives major bumps inside a grouped PR. Either move them to the `frontend-prod` group's `update-types: ["minor","patch"]` treatment via an explicit pattern, or add a dedicated group.
- [ ] Correct the `# Dev deps never ship` comment — it is the load-bearing justification and it is wrong as written.
- [ ] Decide how a group should behave when it contains a known-blocked major. Options: exclude the specific packages via `ignore` while #513/#514 are open, or split majors out of the dev group as well.
- [ ] Add the `@typescript-eslint` 8.x cap (`typescript >=4.8.4 <6.1.0`) to **#514**, which currently records only the `ts-jest` blocker.
- [ ] A regenerated `frontend-dev` PR installs cleanly.

## Tests

`scripts/test_dependabot_config.py` already asserts that no group reachable by a production dependency omits or widens `update-types`. That guard passes today because it trusts `dependency-type: "development"` as the definition of "cannot reach users". It should grow a case for build-time toolchains that ship artifacts despite that classification — otherwise this exact hole stays open by design.

## Note

Not fixed inline while clearing the queue: changing grouping policy is a deliberate call and belongs with the #517/#539 line of work, not in a sweep. #548 is left open and red rather than closed, since closing it loses the branch (the #504 lesson from #517).

frankbria:
Closes #555.

## The premise that was wrong

`.github/dependabot.yml` exempted the `frontend-dev` group from the no-majors rule on
one comment:

```yaml
frontend-dev:
  # Dev deps never ship, so a major is a CI concern at worst.
```

`dependency-type` classifies a **manifest section**, not a blast radius. `tailwindcss`,
`postcss` and `autoprefixer` sit in `devDependencies` and generate the CSS bundle every
user downloads — so `tailwindcss` 3 → 4 rode into #548 as one line of a 14-update batch,
on the same footing as an eslint bump.

## The fix

- New **`frontend-build`** group takes `tailwindcss` / `postcss` / `autoprefixer` at
  `minor`/`patch`; `frontend-dev` names them in `exclude-patterns`. Their majors arrive
  solo; their patches still batch.
- **`ignore`** on `typescript` and `tailwindcss` semver-major while #514 / #513 are open.
  Each entry names the issue whose closure removes it.
- Guard `test_build_time_toolchains_never_ride_a_major_taking_group`: any group that
  takes majors must not be able to match a shipping build tool.

## The wider fix was tried first, and measured wrong

Barring *every* group from grouping a major is one deleted condition and looks cleaner.
Reconstructing #548's manifest and running `npm install --package-lock-only` per package
shows three of the five now-solo majors **cannot install alone**:

| bump | alone |
|---|---|
| `@testing-library/jest-dom` 6→7 | OK |
| `jest-axe` 10→11 | OK |
| `@typescript-eslint/eslint-plugin` 6→8 | **FAIL** — peer `@typescript-eslint/parser@^8.68.0` |
| `eslint` 8→10 | **FAIL** — peer `eslint@"^7 \|\| ^8"` from `@typescript-eslint/parser@6.21.0` |
| `eslint-config-next` 15→16 | **FAIL** — peer `eslint@">=9.0.0"` |

The eslint family is peer-coupled and only moves together. Forcing every dev major solo
would have stranded four permanently-unmergeable PRs — #555's own complaint, relocated
rather than fixed. That is why the branch carries the first approach and then replaces
it; the intermediate commits are the measurement that changed the design.

So `frontend-dev` keeps its majors. The line is **what a package ships**, not which
manifest section it sits in.

## AC5 — a regenerated `frontend-dev` PR installs cleanly

#548 reconstructed under this config keeps **9 of 14** updates and resolves **exit 0**,
against `ERESOLVE` today. `eslint` 10.9.1, `eslint-config-next` 16.3.3 and both
`@typescript-eslint` packages at 8.68.0 cross together; `tailwindcss` holds at 3.4.19 and
`typescript` at 5.9.3, as intended.

Full evidence, per criterion: `docs/demos/2026-08-27-issue-555-dev-group-majors.md`.

## Also done

- **AC4 → #514.** It recorded only the `ts-jest@29` cap (`typescript ">=4.3 <7"`).
  Measured, `@typescript-eslint` 8.x peers `>=4.8.4 <6.1.0` — tighter, and the cap that
  actually fires. Swapping the Jest transform does **not** unblock TS 7 on its own.
  #514 updated with the output; #513 noted for the tailwindcss `ignore`.
- **A number in the issue corrected.** #555 says "twelve of the fourteen updates are
  unobjectionable". Measured against the manifest it is six — four range bumps plus two
  lockfile-only moves. The config comment carries the measured figure.

## Verification

- `pytest scripts/ -q` → **40 passed, 4 skipped** (runs in CI's `Security Audit` job).
- Mutation-checked both ways, so the guard is not vacuous:
  drop `exclude-patterns` → 1 failed; let `frontend-build` take majors → 2 failed.
  The new guard is also RED against `main`'s config.
- **Third-party review**: `codex review --base main` — *"I did not identify any actionable
  correctness issues in the diff."* opencode/GLM was tried first and failed twice with an
  upstream `UnknownError` (`err_1df95002`).

## Known limitations

- The three `npm warn ERESOLVE overriding peer dependency` lines in the AC5 resolution
  are **new** — `main`'s manifest emits zero. Warnings only; npm resolves and `npm ci`
  installs. They belong to whoever reviews the regenerated #548, not to this config
  change, and are recorded in the demo so they are not a surprise.
- This cannot be verified against a live sweep until the next weekly run. The manifest
  reconstruction is the strongest evidence available before then — same limitation
  #539 has.
- `open-pull-requests-limit` truncation is still unexercised (#539). With `tailwindcss`
  and `typescript` now ignored, this change makes truncation *less* likely, not more.
- `tasks/lessons.md` rides along: it was uncommitted on `main` at branch time and is the
  author's notes from #556. Docs-only, unrelated to this change.


frankbria:
## Third-party review (post-PR, Phase 10)

Diff unchanged since the pre-PR pass (`HEAD == REVIEWED_SHA`), so this posts that run rather than re-running the reviewer.

**Primary — opencode / GLM: unavailable.** Two attempts, both failed upstream before producing output:

```
Error: { "name": "UnknownError", "data": { "message": "Unexpected server error. Check server logs for details.", "ref": "err_1df95002" } }
```

**Fallback — `codex review --base main`: no findings.**

> The changes consistently update Dependabot grouping rules and their guard tests so development dependencies no longer group majors, with explicit ignores for known blocked majors. I did not identify any actionable correctness issues in the diff.

Codex did not check whether the now-solo majors can install individually. I verified that myself and it turned up the one real defect in the first approach — three of the five fail on their own peer ranges — which is why the design changed mid-branch. Evidence in the demo doc and the PR body.

frankbria:
`review / review` failed at 18m15s with **"Claude encountered an error after 17m 49s"** and none of its six checklist items completed — no findings were produced. That is #547's signature (cancelled around the 20m mark, posting a review that reads as completed), and it is the same GLM backend that failed my two direct opencode attempts today with `UnknownError` (`err_1df95002`). Infrastructure, not a finding.

Not a required check. The four that are — Frontend Tests, Backend Tests, E2E Tests (Playwright), Security Audit — all pass, and `codex review --base main` covered the diff with no actionable findings.

PR #594, opened 2026-09-06
fix(security): pin every GitHub Action to an immutable commit SHA

frankbria:
Closes #518.

26 `uses:` references across the four active workflows, spanning **10 distinct actions**, move from floating tags to 40-character commit SHAs with a trailing version comment. `tailscale/github-action` and the `glm-review` reusable workflow were already pinned.

Evidence: [`docs/demos/2026-09-06-issue-518-pin-actions-to-shas.md`](../blob/feature/518-pin-actions-to-shas/docs/demos/2026-09-06-issue-518-pin-actions-to-shas.md).

Hardening, not incident response — no action here is known to have been repointed. The change removes the capability; the guard removes the chance of it coming back.

## A tag is not a weaker pin, it is not a pin

`actions/setup-node@v7.0.0` was already a full semver tag and was exactly as mutable as `@v7` — any tag can be force-moved to a different commit by the upstream owner, or by whoever compromises that account, and every subsequent run executes the new code with no diff, no PR and no notification. Specificity is not immutability.

`build-images.yml` is why this was P2: `docker/login-action` receives the registry token and `build-push-action` publishes the images `deploy-staging-containers.yml` later pulls and runs on the box.

## Verification, in both directions

Every SHA was resolved from the tag the workflow named (`git/ref/tags`), then confirmed by listing the tags pointing back at that commit — a pin is only as good as the verification behind it.

**`codecov/codecov-action@v7` is an annotated tag.** Its ref's `.object.sha` is the *tag object*, not the commit; pinning that value would have produced a reference resolving to nothing. It was dereferenced through `git/tags/<sha>`. Nine of ten were lightweight and this one was not, so the resolve procedure is written into `quality-standards.md` rather than left as folklore.

## The guard, and the bug the reviewer found in it

`scripts/test_actions_are_sha_pinned.py` runs in `Security Audit`: every `uses:` SHA-pinned, every pin carrying a version comment, every workflow still parsing. Mutation-checked four ways with the guard staged first:

| mutation | result |
|---|---|
| `docker/login-action` reverted to `@v4` | ❌ not-pinned test |
| version comment stripped | ❌ comment test |
| guard repointed at an empty directory | ❌ `test_there_are_workflows_to_check` |
| unpinned `zz-probe.**yaml**` dropped in | ❌ not-pinned test |

[...]

Issue #547, opened 2026-08-27
[P3.26.1] glm-review is cancelled at ~20m on large diffs and posts an empty review that reads as completed


frankbria:
## Problem

On PRs with a large diff, `glm-review.yml`'s review step is **cancelled at ~20m25s** and posts a comment that *looks like a completed review but contains no findings*. The check goes red, so a reader has two ways to be misled at once:

- treat the red as a finding → chase a defect that was never reported;
- treat the posted comment as coverage → believe the diff was reviewed when the reviewer never got past step one.

Reproduced twice on #545, same diff, ~30 minutes apart:

| run | window | duration | conclusion |
|---|---|---|---|
| [33008885029](https://github.com/frankbria/auto-author/actions/runs/33008885029) | `20:08:36 → 20:29:00` | 20m24s | cancelled |
| [33011890917](https://github.com/frankbria/auto-author/actions/runs/33011890917) | `20:44:10 → 21:04:37` | 20m27s | cancelled |

Step-level: `GLM bug-hunting review: cancelled`, `Post GLM bug-hunting review: success` — so the *posting* step succeeds regardless, which is what publishes the empty review. In both cases the posted comment's checklist was **entirely unchecked**:

```
- [ ] Pull PR context and full diff
- [ ] Check existing PR comments to avoid duplicates
- [ ] Read the new workflow in full ...
```

## Why it matters

This is not cosmetic. On #545 the missing review was load-bearing: falling back to `codex review` surfaced a **P1** — a `${IMAGE_TAG:?...}` requirement that would have made the new workflow fail on its first real run. Had the empty GLM comment been read as "reviewed, nothing found", that would have shipped.

## Cause

`.github/workflows/glm-review.yml` sets **no `timeout-minutes`** at either job or step level, so nothing in the repo's control bounds the run — the ~20m25s cutoff comes from inside the action. The consistency of the two durations points at an internal limit rather than a flake.

## Not the same as #519

#519 covers this workflow failing **fast** on Dependabot PRs because the secret is withheld from bot PRs (visible here as the 14s and 5m41s failures). This issue is the opposite shape: a long run on a *human* PR with a large diff, cancelled mid-flight. Same workflow, different fix.

## Acceptance criteria

- [ ] Set an explicit `timeout-minutes` so the bound is the repo's, not an opaque internal one.
[...]

PR #596, opened 2026-09-06
fix(security): scrub the operator's own network from tracked docs

### LINKED ISSUE #554: [P2.20.2] Public repo discloses the operator's own network: home dynamic IP with ISP named, and a /24 granted SSH in a ufw rule
Found by the cross-family review on #553 (the #544 scrub). Deliberately **not** folded into that PR: #544 is scoped to *staging infrastructure* identifiers, and these are a different category — the **operator's own network**. Values are not restated here; see the locations.

## Locations

| file:line | what it discloses |
|---|---|
| `docs/INCIDENT-2025-10-19-firewall-lockout.md:46` | operator's home/office **dynamic IP, named together with the ISP** |
| `docs/INCIDENT-2025-10-19-firewall-lockout.md:21` | a `/24` source prefix used in a `ufw allow ... port 22` rule |
| `docs/INCIDENT-2025-10-19-firewall-lockout.md:6` | a previous provider's production server address |
| `docs/fixes/2025-11-22-clerk-webhook-nginx-fix.md:20` | client IP in a verbatim access-log excerpt |
| `tasks/lessons.md:112` | client IP in a uvicorn access-log excerpt |

## Why this is separate from #544, and lower severity

- **Not staging infra.** #544 was about the staging host and the Atlas cluster. Scrubbing these would not have made that issue any more complete.
- **Mostly stale.** The line naming the ISP describes a *dynamic* address from 2025-10-19 — near-certainly reassigned. The server address at :6 belongs to a provider that has since been retired.
- **No credential value.** Nothing here is half a credential the way the database username in #544 was.

## Why it is still worth doing

The `/24` line is the interesting one: it does not just name an address, it records **which prefix was granted SSH access**, which is a statement about the operator's network rather than about a machine that no longer exists. Combined with the ISP name on :46, a public repo is describing the maintainer's home network. That is an attack-surface and personal-privacy concern rather than an infrastructure one, which is exactly why it deserves its own issue instead of being quietly appended to a staging-scrub PR.

## Proposed fix

Same treatment as #544, Option 1 (scrub the working tree, do not rewrite history):

- [ ] Replace with `<operator-network>`, `<client-ip>`, `<former-prod-host>` as appropriate, keeping each doc readable.
- [ ] Remove the ISP name where it sits alongside an address.
- [ ] Decide whether `scripts/test_no_staging_identifiers.py` should grow a rule here. **Probably not** — a public-IPv4 rule was deliberately rejected in #553 because it fires on version strings, RFC 5737 addresses and the ~88 tracked references to the decommissioned host. A narrower rule (an IP on a line that also matches `ufw|allow from|ISP|Cox|dynamic`) may be worth it; a blanket one is not.

## Out of scope

History rewriting. Same reasoning as #544: these values are already in every clone, and on a public repo may be mirrored or indexed.

frankbria:
Closes #554. Sibling to #544 — different category: the maintainer's own network rather than staging infrastructure.

Evidence: [`docs/demos/2026-09-06-issue-554-scrub-operator-network.md`](../blob/feature/554-scrub-operator-network/docs/demos/2026-09-06-issue-554-scrub-operator-network.md).

## The issue's location table was wrong in both directions

So the scan was redone from scratch rather than worked as a checklist: eight distinct non-private IPv4-shaped strings across tracked files, each classified by reading the line it sits on.

**Over-reported twice.** `tasks/lessons.md` was listed as leaking a client IP — the value is `1.2.3.4`, already the canonical placeholder, so **no change was made**. The Clerk webhook log's "client IP" resolves to AWS eu-west-1: it is **Clerk's own egress**, not the operator's connection. Scrubbed anyway as valueless log noise, but the rationale is corrected so nobody misreads what was exposed.

**Under-reported once, and it is the one that mattered.** The incident doc carries a consumer **dynamic-DNS hostname** the table never listed. The issue's whole severity argument rests on staleness — a 14-month-old dynamic IP is near-certainly reassigned, the former provider is retired. **That argument does not apply to a DDNS hostname, which still resolves.** It is a live pointer to the connection it was set up for. The ISP name was also in two places, not the one listed.

Three of the four remaining scrub targets came from re-reading the file; only one came from the table.

## Deliberately not done

**87 of the 88 references to the decommissioned host remain.** Line 6 was replaced as the issue asked, but that is cosmetic consistency inside one document and **not a reduction in exposure** — it should not be read as one. Scrubbing 1 of 88 is theatre; the honest options are all or none. None is defensible for a dead machine at a retired provider whose references sit mostly in read-only `archive/`. **If that trade should go the other way it is a separate decision**, flagged rather than silently made.

## The guard

`scripts/test_no_operator_network_identifiers.py`, in `Security Audit`. Three rules, each requiring an IP **plus** context: a firewall grant naming a source address, a consumer dynamic-DNS hostname, and a public IP *near* wording calling it a home/office/dynamic address.

**No rule names an operator value** — a denylist for this category would embed the strings it exists to exclude, in the file that forbids them. Fixtures use RFC 5737 addresses.

The blanket public-IPv4 rule stays rejected, with the sweep as evidence: it would fire on three `frontend/public/*.svg` files whose path data parses as IPv4, plus the 88 dead-host references. A check that noisy gets switched off, which is worse than none.

Mutation-checked seven ways, including **three must-not-fire probes** (a documentation address, SVG digit runs, the `1.2.3.4` placeholder) and a vacuous-scan probe.

## The part worth reviewing: I reintroduced scrubbed values three times

All three were in prose *about* the work, not in the work.

1. The changelog entry restated the Clerk address — **caught by the new guard**.
2. The demo doc's verification block pasted the grep commands with arguments filled in, putting the `/24` prefix and ISP name into the document asserting they were gone, and falsifying its own claim. **Caught by `codex review`.**
3. The guard's own fixture named the real ISP. Found while fixing #2.

**The guard could not have caught 2 or 3**: a `/24` written as three octets matches no IPv4 pattern, and an ISP name is a proper noun no shape rule separates from prose. Shape rules catch shapes. That makes the review step load-bearing here, and the pattern is specific enough to name — **the highest-risk place to reintroduce a scrubbed value is the document explaining the scrub.**

Fixing #1 also exposed a real bug in the third rule: it ANDed *an IP anywhere on the line* with *the description anywhere on the line*, and a changelog entry here is one 3000-character line, so prose about the guard matched a placeholder mentioned paragraphs away. It now requires the two within 60 characters — the relationship it always meant. **The fix was to tighten the rule, not to allowlist the file**, which would have gone green in one line and left the rule wrong everywhere else.

## Verification

- Every scrubbed literal grepped repo-wide: **no match** for the residential IP and its `/24`, the DDNS subdomain, the ISP name, and the Clerk address
- The DDNS *vendor* name stays in the rule pattern and one fixture, deliberately: the secret is the subdomain, not the provider, and the rule cannot match what it may not name
- `uvx --with pytest --with pyyaml pytest scripts/ -q` — **74 passed, 4 skipped**
- Third-party review: `codex review --base main` — one P2 finding, fixed


frankbria:
## Third-party review — `codex review --base main`

One **P2** finding, and it was a real self-falsifying error:

> **[P2] Remove the reintroduced operator network prefix** — `docs/demos/2026-09-06-issue-554-scrub-operator-network.md:72`
> Because this repo is public and this change is specifically scrubbing the operator's `/24`, committing it here reintroduces the sensitive network prefix in the evidence document. A fresh `git grep` now matches this new line, so the stated repo-wide verification is no longer true; use a placeholder in the command/example instead.

Fixed. The verification block now reports results without pasting the literals, and says how to re-run it from the parent commit. Fixing it turned up a third instance — the guard's own fixture named the real ISP — so the module comment now states that a test sample must never carry a real operator value.

Worth recording why the guard was no help on either: a `/24` written as three octets matches no IPv4 pattern, and an ISP name is a proper noun no shape rule separates from prose. **Shape rules catch shapes.** The guard lowers the floor; review is load-bearing for this category.

frankbria:
## `review / review` red is #547 again — and a worse instance

Failed at **20m16s** having ticked **nothing**. Its progress comment never got past "Analyzing…":

- [ ] Gather full PR diff and context
- [ ] Verify scrub completeness
- [ ] Audit the guard's rule logic
- [ ] Check docs for self-defeating content
- [ ] Confirm the guard actually runs in CI (not silently skipped)

On #594 it at least completed four of five items before dying. Here it produced zero. Both runs land at ~20m1x s, so the cutoff is fixed and the amount of work that fits under it is not — noted on #547.

**Verified its last item myself, since nobody else did.** `tests.yml:219` runs `uvx --with pytest --with pyyaml pytest scripts/ -q`, which globs the directory, so a new file there is picked up without wiring. The `Security Audit` job log confirms it rather than assuming it:

```
Test the audit gate and dependabot config
74 passed, 4 skipped in 2.73s
```

**74**, against 59 on `main` before this branch — the 15 new tests are the ones added here. Not silently skipped.

Third-party review of record for this PR is `codex review --base main`, whose one P2 finding (the demo doc reintroducing the `/24`) is fixed. Merging on the four required checks.

Issue #558, opened 2026-08-27
[P2.25] Local staging E2E credentials are stale, so `npm run test:e2e:staging` cannot authenticate outside CI


frankbria:
## What

`frontend/tests/e2e/staging/.env.test` holds credentials that no longer work. Signing in with them against staging returns:

```
{"message":"Invalid email or password","code":"INVALID_EMAIL_OR_PASSWORD"}   HTTP 401
```

and the frontend container logs `WARN [Better Auth]: User not found` — so the address in that file has no account on staging Atlas at all. The file is dated 2026-01-01; the live credentials are the `TEST_USER_EMAIL` / `TEST_USER_PASSWORD` secrets in the `staging` GitHub environment, which is why CI passes while local runs cannot authenticate.

## Why it matters

`npm run test:e2e:staging` is documented in `CLAUDE.md` and `tests/e2e/staging/README.md` as a local command, and it cannot work as written. The practical cost showed up in #550: the "real login succeeds" acceptance criterion had to lean entirely on a CI dispatch, because no local curl or Playwright run could sign in as the real staging user. The direct session-cookie → `/api/v1/books` round trip had to be done with a throwaway account instead.

That matters more now than it did last week. **#556 needs a staging re-verification**, and the three-point test it specifies (an account created on better-auth ≤1.6 must still sign in) is far cheaper to iterate locally than through a 15-minute CI dispatch per attempt.

## There is also a quoting bug in the file

The stored password contains an unescaped `$`, so `set -a; . .env.test` makes bash expand it as a variable — which both corrupts the value and prints a fragment of it in the resulting `unbound variable` error. Anything sourcing this file rather than parsing it will silently get the wrong password. Whatever replaces the credentials should be single-quoted, and `.env.test.example` should show the quoting.

## Acceptance criteria

- [ ] `frontend/tests/e2e/staging/.env.test.example` documents where the live credentials come from and shows a single-quoted value
- [ ] A working local path exists — either refreshed credentials for the documented account, or docs stating plainly that local staging E2E requires credentials pulled from the `staging` environment and how to get them
- [ ] `npm run test:e2e:staging` authenticates locally against staging, or the README says explicitly that it cannot and why
- [ ] The `$`-quoting trap is not reproducible from the example file

## Notes

- `.env.test` is gitignored; the stale values are local-only, not committed. No secret is exposed by this issue.
- Verified during #550 on 2026-08-27 against `dev.autoauthor.app`.

## Related

[...]

PR #598, opened 2026-09-06
fix(e2e): sign in once per worker so staging E2E stops tripping the auth rate limit (#551)

### LINKED ISSUE #551: [P2.22] Staging E2E #83 session canary times out on first login every run (3/3, ~32.4s) and passes only on retry
## What

Every staging E2E run inspected so far reports `1 flaky, 10 passed`. The flaky test is always the same one:

```
regressions.spec.ts:35:7 › Issue #83 regressions › session + user lookup: authenticated routes never return 401
```

It fails its **first** attempt and passes on Playwright's `retry #1`.

## It is not random

Three consecutive runs on 2026-08-27, spanning two different deployed images:

| run | image | first attempt | retry #1 |
|---|---|---|---|
| [33028231204](https://github.com/frankbria/auto-author/actions/runs/33028231204) | `sha-de87eea` (mongodb 6.21.0) | ✘ 32.7s | ✓ 3.9s |
| [33028714230](https://github.com/frankbria/auto-author/actions/runs/33028714230) | `sha-bae440a` (mongodb 7.5.0) | ✘ 32.4s | ✓ 3.7s |
| [33030415162](https://github.com/frankbria/auto-author/actions/runs/33030415162) | `sha-bae440a` | ✘ 32.4s | ✓ — |

3/3, on both drivers, with the first attempt clustering at **32.4–32.7s** and the retry at **3.7–3.9s**. A random flake does not reproduce to a tenth of a second. This is a **consistent first-login timeout**: the first sign-in of a run exceeds the fixture's 30s budget, and every subsequent one takes ~4s.

## Failure

```
TimeoutError: page.waitForURL: Timeout 30000ms exceeded.
  waiting for navigation until "load"
  at fixtures/auth.fixture.ts:52
```

`frontend/tests/e2e/staging/fixtures/auth.fixture.ts:52`:
```ts
await page.waitForURL(/\/dashboard/, { timeout: 30000 });
```

Note this is the *first* spec in the file to sign in; the earlier specs in the run use other paths. So the ~32s is the cost of the **first** authenticated sign-in against staging, and the ~4s retry is the warm path.

## Why it matters

1. **It masks regressions.** `regressions.spec.ts` #83 is the session/401 canary — the spec most likely to catch an auth-path break. A permanently-retried canary cannot distinguish "cold start" from "session handling actually broke": a real regression that made first login fail would look exactly like the status quo and still report `success`.
2. **It devalues the staging E2E gate.** The job exits `success` on `1 flaky`, so the signal is "green" no matter what happens on first attempt.
3. It cost this repo a correction already — [#549](https://github.com/frankbria/auto-author/pull/549) originally recorded the #516 verification as flatly "green" and had to be amended to disclose it.

## Likely cause (to confirm, not assume)

Cold-path latency on the first authenticated request: some combination of Next.js route JIT on a freshly started container, better-auth's first Atlas round trip, and TLS handshake to the hosted cluster. `auth.ts` retries the Mongo connect up to 3× with backoff, which alone could consume the budget. Deploy logs show containers started ~2 min before the first of these runs, which fits — but the third run was 33 min after deploy and still took 32.4s, which does **not** fit a pure container-cold-start theory. That contradiction is the first thing to chase.

## Acceptance criteria

- [ ] Root cause identified with evidence (frontend/backend timings or container logs for a first sign-in), distinguishing container cold start from per-run first-auth cost — specifically explaining why run 33030415162 was still slow 33 min after deploy.
- [ ] First sign-in against staging completes within the fixture budget, **or** the budget is raised with the measured p95 recorded as the justification.
- [ ] Three consecutive staging E2E runs report `0 flaky`.
- [ ] Decide and document whether `E2E Staging (Playwright)` should fail on `flaky`, so a retried canary cannot report green.

## Out of scope

Changing the spec's assertions, or deleting the retry, to make the number go away. The retry is currently the only reason the suite passes; removing it without fixing the latency turns staging E2E permanently red.

frankbria:
Closes #551.

## The issue's hypothesis was wrong, and the contradiction it flagged is the tell

#551 guessed cold-path latency — Next.js route JIT, better-auth's first Atlas round trip, the Mongo connect retry loop — and then noted that run [33030415162](https://github.com/frankbria/auto-author/actions/runs/33030415162) was still slow 33 minutes after deploy, which does not fit. It doesn't fit because container age was never the variable.

**It is better-auth's built-in sign-in rate limit.**

## Evidence

Reproduced today on run [34050880126](https://github.com/frankbria/auto-author/actions/runs/34050880126) (2026-09-06, `1 flaky`, 32.4s — same figure as the three runs in the issue).

**1. The page says so.** The failure artifact's `error-context.md` captures the sign-in form at the moment of the timeout, showing an alert reading `Too many attempts. Please try again later` — which is `src/app/auth/sign-in/page.tsx:59-65` mapping a 429.

**2. The rule.** `better-auth/dist/api/rate-limiter/index.mjs`, `getDefaultSpecialRules()`: any path starting `/sign-in` gets `window: 10` seconds, `max: 3` per IP. Rate limiting is on by default in production and `frontend/src/lib/auth.ts` sets no `rateLimit` override. Crucially the window is not fixed — `decideConsume` resets the counter only when `now - lastRequest >= 10s`, so back-to-back sign-ins never let it reset.

**3. The arithmetic.** The fixture signed in once per test — 11 sign-ins from one CI IP. Replaying the actual login timestamps from that job's log against the rule:

| # | login at | gap | count |
|---|----------|-----|-------|
| 1 | 18:11:43.9 | — | 1 |
| 2 | 18:12:54.8 | 70.9s | reset → 1 |
| 3 | 18:12:57.2 | 2.4s | 2 |
| 4 | 18:13:02.3 | 5.1s | 3 |
| 5 | 18:13:13.7 | 11.4s | reset → 1 |
| 6 | 18:13:20.6 | 6.9s | 2 |
| 7 | 18:13:26.1 | 5.5s | 3 |
| **8** | **18:13:31.6** | **5.5s** | **3 ≥ max → 429** |
| retry | 18:14:04.6 | 38.5s since #7 | reset → 1, passes in 3.7s |

The rule predicts the 429 on exactly the 8th sign-in. That is exactly where it landed. It is `regressions.spec.ts:35` every run not because that spec is special but because it is 8th at the suite's deterministic pacing.

This also settles the cold-start question the issue asked to distinguish: **there is no per-run first-auth cost at all.** The seven sign-ins before the failure took **0.9s, 1.0s, 1.1s, 1.1s, 0.9s, 0.9s, 0.9s** — including sign-in #1, 27 seconds into a fresh run. The 30s fixture budget was never the constraint.

## Changes

- **`fixtures/auth.fixture.ts` — sign in once per worker**, via Playwright's documented worker-scoped `storageState` pattern. 11 sign-ins per run become 1, an order of magnitude under 3-per-10s. `authenticatedPage` keeps its name and semantics, so no spec changes; the `edge-cases.spec.ts` test that calls `clearCookies()` is unaffected (it clears its own context, not the saved state).
- **Surface the sign-in error.** When the dashboard redirect never arrives, the thrown error now carries the sign-in page's own alert text and the current URL, instead of a bare `page.waitForURL: Timeout 30000ms exceeded`. That opaque message is most of why this took a dedicated investigation.
- **`playwright.config.ts` — `failOnFlakyTests: !!process.env.CI`** (AC#4). `retries: 2` stays so a genuinely transient failure still produces its retry trace, but the canary can no longer fail, retry, and report the job green. Rationale documented in the config comment and `tests/e2e/staging/README.md`.
- **`StagingE2eFlakyGate.test.ts`** — a jest guard so the AC#4 decision cannot be silently reverted.

## Acceptance criteria

| Criterion | Status |
|---|---|
| Root cause with evidence, cold start vs per-run first-auth distinguished, 33030415162 explained | Done — above |
| First sign-in within budget, or budget raised with measured p95 | Done — measured 0.9–1.1s over 7 samples; budget unchanged at 30s |
| Three consecutive staging E2E runs report `0 flaky` | Done — **four** consecutive dispatched runs, each `11 passed`, `0 flaky`, **1 sign-in** (was 11). Table in the demo comment. |
| Decide and document fail-on-flaky | Done — enabled under CI, documented in config + README, and **demonstrated**: same flaky spec exits 0 without the gate and 1 with it |

## Known limitations

- **Not verified locally.** The local `.env.test` credentials are stale (#558) — a direct `POST /api/auth/sign-in/email` with them returns 401 — so verification runs through the dispatched staging workflow rather than `npm run test:e2e:staging`.
- **The 3-per-10s limit still applies**, it is just no longer approached. A future spec that signs in per-test would reintroduce this; the README now says so.
- **Possible related exposure, not addressed here:** better-auth's `getIP` returns `null` when `x-forwarded-for` carries more than one hop and no `advanced.ipAddress.trustedProxies` is configured (none is), in which case *every* client shares one rate-limit bucket — i.e. 3 sign-ins per 10s for the whole app, not per user. Whether staging's nginx produces a single-hop XFF was not determined from outside the box. Worth its own issue.
- **Session tokens still reach public artifacts via retry traces.** Found by the post-PR review. `trace: 'on-first-retry'` writes into the uploaded `outputDir`, and trace network records include request headers, so a retried test's `Cookie:` header carries the session token. Pre-existing, not introduced here, and left to #599 because the fix is an artifact-policy decision rather than a code correction. This PR does fix the *new* instance it would have introduced, and sharply reduces the frequency by making retries stop firing. Detail added to #599.
- **`tsconfig.json` excludes `**/e2e/**`**, so `npm run typecheck` does not cover this fixture. Typechecked explicitly instead; clean apart from a pre-existing error in the untouched reporter array.

## Review

Third-party review: opencode (GLM 5.2), two passes — pre-PR and post-PR. Both posted as a PR comment with triage. It caught a Critical this change would otherwise have introduced (a live session token written into the public artifact tree), plus four smaller issues, all fixed.


frankbria:
## Third-party review — opencode (GLM 5.2)

Two passes. **Pre-PR** (on `0212bd9`) found one Critical + one Major, both fixed in `15a39cb`. **Post-PR** (on `5ee4acb`) verified those fixes against the repo and found three Minors, fixed in `91fd228`.

### Pass 1 — pre-PR

> **C1 — Live staging session cookie is uploaded to a public CI artifact.** The fixture wrote storageState (including `__Secure-better-auth.session_token`) to `workerInfo.project.outputDir`, which resolves under `frontend/tests/e2e/staging/test-results/` — uploaded whole `if: always()` from a **public** repo.
>
> **M1 — `STAGING_E2E_WORKERS` deterministically trips the rate limit it fixes.** Workers start concurrently, so W workers = W simultaneous sign-ins; W≥4 hits 3-per-10s instantly.
>
> **M2 — The `waitForURL` catch discards the original error.** A closed context or browser crash gets mislabeled, and the real stack is gone.
>
> Q4 — sharing one session per worker is safe for existing specs; `clearCookies` only touches the calling test's context. The latent trap is **sign-out**, which no spec does today.

Verified C1 independently before fixing: the staging config sets no `outputDir`, so it defaults to `test-results` beside the config, and the workflow uploads exactly that path. Confirmed.

### Pass 2 — post-PR (full text)

> I'll verify this diff against the actual repo. Let me set up a review plan and gather the files.Files match the diff. Now let me read the specs and check the surrounding infrastructure (CI upload paths, Playwright version, tsconfig).Specs read. Now the CI workflow (artifact upload paths — central to fix 1), tsconfig, jest config, and whether `.env.test` is tracked.`.env.test`, `test-results/`, and the HTML report are all gitignored. Now I need the CI workflow to verify what actually gets uploaded as a public artifact.Workflow confirmed: `test-results/` (default outputDir with traces on retry) and the HTML report (which embeds traces) are uploaded `always()` — that matters for the token-leak question. Let me check the installed Playwright version, jest config, and the better-auth rate-limit/session config in the backend.Now verifying Playwright semantics (failOnFlakyTests, worker fixture teardown, trace contents) against current docs.One more docs check — what traces capture (this decides whether the token still leaks via uploaded trace artifacts):Docs confirmed: worker-fixture teardown after `use()` runs when the worker finishes; `failOnFlakyTests` config option exists (≥1.52, installed 1.62.1); trace viewer records request headers. Now let me actually execute the jest guard and the regex checks to verify fix 4 empirically:# Post-PR review — staging E2E session-sharing change
>
> ## Fix verdicts
>
> **(1) Temp-dir storage state — correct and complete for the file itself.**
> - `mkdtempSync` gives a 0700 dir under `/tmp` on `ubuntu-latest` (TMPDIR unset → `/tmp`, outside the workspace). `e2e-staging-tests.yml` uploads only `playwright-report-staging/`, `tests/e2e/staging/test-results/`, and `frontend/test-results/` — nothing sweeps `/tmp`, so the file cannot be collected.
> - Teardown ordering is right: worker-scoped teardown after `use()` runs when the worker finishes; every test context (including retry attempts) reads `storageState` at context *construction*, strictly before teardown. `rmSync(..., {force: true})` is idempotent; `context.close()` in its own `finally` covers login failure. mkdtemp's random suffix makes the `workerIndex` filename collision-free even if a dead worker's index is reused.
> - Only residual: a hard-killed worker (OOM/SIGKILL) skips `rmSync` → stale 0700 file on disk. Ephemeral CI runners destroy the disk; locally it survives until reboot. **Minor.**
>
> **(2) Workers cap — arithmetic verified** (executed): unset/`''`/`'0'`/`'abc'`/`' '` → 1; `'4'`/`'10'` → 3. Note `'0'` and `'abc'` are *improvements* over the old code (old produced `workers: 0` / `NaN`). CI passes `'1'` for schedule/PR (`inputs.staging_e2e_workers || '1'`), dispatch default `'1'`. **Correct.**
>
> **(3) Error chaining — correct.** Node engines `>=24` and CI pins node 24 → `cause` fine (≥16.9). `tsconfig.json:45` excludes `**/e2e/**` from `tsc --noEmit`, so no lib gate applies; even if included, `lib: ["esnext"]` provides `ErrorOptions`. `readSignInError` cannot mask: its only awaitable has `.catch(() => null)` swallowing every rejection (timeout, detached, closed page, crashed browser), and `page.url()` is a sync non-throwing accessor. Costs ≤2s on an already-failed path. **Complete.**
>
> **(4) Jest guard — verified passing** (ran it: 3/3 pass; regexes match the live config). `failOnFlakyTests` is a real option in the lockfile's `@playwright/test@1.62.1` (config-level since ~1.52). **Correct.**
>
> ## Findings
>
> **Major — the token still leaks into public artifacts via retry traces (residual, pre-existing, defeats the stated goal of fix 1).** `trace: 'on-first-retry'` (playwright.config.ts:58) writes `trace.zip` into the default outputDir `tests/e2e/staging/test-results/`, and the HTML report embeds traces — both uploaded `if: always()` (e2e-staging-tests.yml:107-123) from a public repo. Trace network records include request **headers**, so every authenticated request in a retried test carries `Cookie: __Secure-better-auth.session_token=...`. Session lifetime is 7 days (src/lib/auth.ts:60) vs 7-day artifact retention — and this suite's history shows retries *do* fire. Moving the state *file* out of outputDir was necessary but not sufficient; the token itself must stay out of artifacts. Fix: stop uploading the report/traces for staging (or scrub/short-lived staging sessions). If the staging account can touch billing endpoints (backend has them), treat as Critical.
>
> **Minor — cap is one-sided.** `Math.min(3, ...)` doesn't clamp below/at bounds: `'-2'` → `-2`, `'2.5'` → `2.5` (verified). Only reachable via manual dispatch typing; Playwright fails fast at startup, no security impact. Add a `Math.max(1, Math.round(...))` or a numeric workflow_dispatch input with min/max.
>
> **Minor — zero rate-limit headroom at workers=3.** Three workers all bootstrap inside the same 10s window = exactly 3/3 used. If Playwright replaces a crashed worker mid-run, its worker-fixture setup is a 4th sign-in in-window → 429 → that worker's tests fail at setup with the new (good) error message. Default is 1 worker, so this is opt-in only; the fixture comment "order of magnitude under the limit" is true only at 1 worker. Consider capping at 2, or noting the restart hazard in the README's worker bullet.
>
> **Minor — guard regexes are unanchored.** `/failOnFlakyTests:\s*!!process\.env\.CI/` matches `!!process.env.CI && false`, and all three match commented-out lines. Accepted tripwire trade-off (the file says so), but a `$`-ish anchor or JSON-parse would tighten it.
>
> **Minor — local `node_modules` is stale** (`@playwright/test` 1.57.0 vs lockfile/package.json 1.62.1). Irrelevant to CI (`npm ci`), housekeeping only.
>
> ## Shared-session spec audit — no breakage
>
> - No spec signs out (checked all 4 spec files; no sign-out/logout calls anywhere).
> - `edge-cases.spec.ts:69` clears cookies in its **own per-test context** only; later tests build fresh contexts from the state file — the README claim is accurate.
> - better-auth config (src/lib/auth.ts): no per-user concurrent-session limit, no invalidate-on-login; `updateAge: 24h` rotation can't fire mid-run because the worker token is minted at run start (age ≈ 0); `cookieCache: 5min` is per-context.
> - `workers=1` default keeps execution order identical to before; `workers=2–3` enables cross-file parallelism on one account, but that was already reachable pre-cap and titles are `Date.now()`-unique — only dashboard latency risk (~2,700-book account), no logical conflict.
>
> **Bottom line:** all four claimed fixes are real and correctly implemented; the one thing that needs follow-up is the trace-artifact token leak, which is the same class as the bug this PR's fix 1 was closing.

### Triage

| Finding | Action |
|---|---|
| C1 session token in public artifact | **Fixed** in `15a39cb` — per-worker `mkdtemp` outside the workspace, removed on teardown. Verified the reviewer's path claim independently first. |
| M1 uncapped workers | **Fixed** in `15a39cb`, tightened in `91fd228` — clamped to a rounded 1–2. |
| M2 discarded error | **Fixed** in `15a39cb` — chained via `{ cause }`. |
| Major (pass 2) — token still leaks via retry traces + HTML report | **Not fixed here; tracked in #599.** Real and correctly reasoned: `trace: 'on-first-retry'` writes into the uploaded `outputDir` and trace network records include request headers, so a retried test's `Cookie:` header carries the session token. This is pre-existing and is the same exposure class as #599's plaintext password — it needs the artifact-policy decision that issue exists to make, not a widening of this PR. #599 updated with the specifics. |
| Minor — one-sided cap (`-2`, `2.5` passed through) | **Fixed** in `91fd228`. Verified across `undefined/''/0/1/2/3/10/abc/-2/2.5/' '`. |
| Minor — zero headroom at 3 workers | **Fixed** in `91fd228` — ceiling lowered to 2 so a replaced crashed worker has a slot. |
| Minor — unanchored guard regexes | **Fixed** in `91fd228` — anchored to whole lines. Mutation-checked afterwards: `!!process.env.CI && false`, `retries: 0`, and `Math.min(9,` each kill exactly one guard. |
| Minor — stale local `node_modules` (1.57.0 vs lockfile 1.62.1) | Not acted on. CI uses `npm ci`; my earlier local version reading was the stale one, immaterial either way since `failOnFlakyTests` exists in both. |

One correction the review earned: it noted `tsconfig.json` excludes `**/e2e/**`, which means `npm run typecheck` never covered this fixture at all. Typechecked the two files explicitly instead — clean, apart from a pre-existing error in the untouched reporter array.

frankbria:
## Demo — every acceptance criterion, with outcome evidence

### AC#1 — root cause identified with evidence, cold start distinguished from per-run first-auth cost

Three independent lines, all from run [34050880126](https://github.com/frankbria/auto-author/actions/runs/34050880126):

1. **The page states the cause.** The failure artifact's `error-context.md` captures the sign-in form mid-timeout with an alert reading `Too many attempts. Please try again later` — `src/app/auth/sign-in/page.tsx:59-65` mapping a 429.
2. **The rule exists and is on.** `better-auth/dist/api/rate-limiter/index.mjs` → `getDefaultSpecialRules()`: `path.startsWith("/sign-in")` → `{ window: 10, max: 3 }`, per IP. On by default in production; `frontend/src/lib/auth.ts` sets no override.
3. **The arithmetic predicts the exact failure.** `decideConsume` resets only when `now - lastRequest >= 10s`. Replaying the job log's login timestamps under that rule puts the 429 on the 8th sign-in. The 8th sign-in is the one that failed.

**Cold start is ruled out by measurement, not argument.** The seven sign-ins before the failure took 0.9s, 1.0s, 1.1s, 1.1s, 0.9s, 0.9s, 0.9s — the first of those 27 seconds into a fresh run. There is no first-auth penalty to find. That is also why run 33030415162 was still slow 33 min after deploy: container age was never the variable.

### AC#2 — first sign-in within budget, or budget raised with measured p95

Measured p95 over those 7 samples: **1.1s** against a 30s budget. Budget left unchanged; raising it would have been fixing the wrong number.

### AC#3 — three consecutive staging runs report `0 flaky`

Four consecutive dispatched runs on this branch, every one `0 flaky` with **exactly one sign-in per run** (was 11):

| run | commit | sign-ins | result |
|---|---|---|---|
| [34060063159](https://github.com/frankbria/auto-author/actions/runs/34060063159) | `0212bd9` | 1 | `11 passed (3.2m)` |
| [34060573754](https://github.com/frankbria/auto-author/actions/runs/34060573754) | `15a39cb` | 1 | `11 passed (3.1m)` |
| [34060827529](https://github.com/frankbria/auto-author/actions/runs/34060827529) | `5ee4acb` | 1 | `11 passed (3.3m)` |
| [34061098558](https://github.com/frankbria/auto-author/actions/runs/34061098558) | `91fd228` (head) | 1 | `11 passed (3.0m)` |

Baseline for comparison, run 34050880126 on `main`: `1 flaky, 10 passed (3.9m)`, 11 sign-ins. Note `11 passed` vs `10 passed` — the canary now passes on its **first** attempt, which is the actual thing being fixed. The suite is also ~20% faster.

### AC#4 — fail-on-flaky decided and documented

Decided: **yes**, enabled under CI, with retries kept for the trace. Documented in the config comment and `tests/e2e/staging/README.md`.

Not asserted from the flag being present — demonstrated. Same spec, same retry outcome, only `CI` differing, using the exact `failOnFlakyTests: !!process.env.CI` expression from the staging config and a test that fails once then passes:

```
=== A) gate OFF (CI unset) ===
  1 flaky
    flaky.spec.ts:4:5 › passes only on retry, exactly like the #83 canary did
>>> EXIT=0          <-- the #551 status quo: green despite a failed first attempt

=== B) gate ON (CI=1) ===
  1 flaky
    flaky.spec.ts:4:5 › passes only on retry, exactly like the #83 canary did
>>> EXIT=1          <-- the canary can no longer report green on retry
```

The three jest guards were mutation-checked after staging, not just run: `!!process.env.CI && false`, `retries: 0`, and `Math.min(9,` each kill exactly one guard and nothing else.

### Out of scope, as the issue required

The retry was not deleted and no assertion was weakened. The suite passes on first attempt now because the sign-in succeeds, not because the bar moved.

Issue #550, opened 2026-08-27
[P2.18.1] Verify better-auth 1.7 against staging Atlas — shipped in #543, unexercised on the auth path


frankbria:
## What

#543 bumped **better-auth 1.6.26 → 1.7.1** in `/frontend` (frontend-prod group, merged 2026-08-27). It is now on `main` and **has not been exercised against staging Atlas**.

This is the same class of risk #516 was opened for, and #516 deliberately did *not* cover it: that verification deployed `sha-bae440a` rather than the newer `sha-8c796a9` precisely so that `mongodb` 6.21.0 → 7.5.0 stayed the only auth-path variable. Staging therefore runs `sha-bae440a` today — **without** better-auth 1.7.

## Why it deserves its own pass

better-auth *is* the auth path — it owns sessions, cookies, sign-in, and the twoFactor plugin. And 1.7 already proved to carry a breaking change: `twoFactor.enable()` became a discriminated union (`{ method: 'otp' } | { method: 'totp', totpURI, backupCodes }`), which broke the build on #543 and needed a code fix in `TwoFactorSetup.tsx`.

Notably that arrived in a **minor** bump. Whatever else changed shape in 1.7 was not necessarily caught by typecheck — a runtime behaviour change in session or cookie handling would not be.

CI's E2E covers auth against a **local/ephemeral Mongo**. The residual gap is the same one #516 named: real Atlas, real TLS, real hosted-cluster behaviour, plus this repo's `sameSite="lax"` cookie configuration (#341) against a live domain.

## Acceptance criteria

- [ ] Baseline recorded on the currently deployed `sha-bae440a` (better-auth 1.6.26) — `/health`, sign-in, staging E2E. *(Already green as of 2026-08-27; see `docs/demos/2026-08-26-issue-516-mongodb-7-staging-verification.md`.)*
- [ ] Staging redeployed from a `main` image containing #543 (`sha-8c796a9` or later).
- [ ] Real login succeeds — session cookie issued, `/api/v1/books` returns for that user.
- [ ] `npm run test:e2e:staging` green, **including** `regressions.spec.ts` (#83 session/401 and #54 answer persistence both exercise session handling).
- [ ] 2FA enable flow exercised against staging, since that is the API that demonstrably changed shape. `SecuritySettings.test.tsx` covers it with mocks only.
- [ ] No new auth warnings in the frontend container log on startup or first auth.

## Rollback

[...]

PR #601, opened 2026-09-07
fix(security): stop publishing staging credentials in public CI artifacts (#599)

### LINKED ISSUE #558: [P2.25] Local staging E2E credentials are stale, so `npm run test:e2e:staging` cannot authenticate outside CI

### LINKED ISSUE #599: [P1.1] Staging test password is uploaded in plaintext inside public CI failure artifacts
## What

When a staging E2E test fails inside the auth fixture, Playwright writes an `error-context.md` containing an accessibility snapshot of the page. That snapshot **includes the value of every form input** — including the password field, which the fixture has just filled.

`e2e-staging-tests.yml` then uploads that file in the `staging-test-results` and `staging-failure-artifacts` artifacts, with 7-day retention, **from a public repository**. Actions artifacts on a public repo are downloadable by any GitHub user.

The staging test account's password is therefore recoverable from the artifacts of any run where sign-in failed — which, because of #551, is **every run**. Its email is in the same snapshot.

Discovered while investigating #551; the artifact was downloaded and the credential read in plaintext from it during that work.

## Why it matters

This is a live credential for an account on the live staging environment, published in a place anyone can fetch. It is not a secret-scanning hit, because the value never enters git — it enters an artifact.

Severity is bounded by staging being staging, but the account is real, the environment is internet-reachable, and the same `TEST_USER_PASSWORD` secret is used by the Locust load-smoke job.

## Fix

1. **Rotate the staging test account's password** and update the `TEST_USER_PASSWORD` secret in the `staging` GitHub environment. Treat the current value as burned.
2. **Delete the artifacts that carry it.** Every `staging-test-results` / `staging-failure-artifacts` artifact from a run with a sign-in failure, back to the retention horizon.
3. **Stop uploading the snapshot, or stop putting the value in the DOM.** Options, roughly in order of how much they actually fix:
   - Exclude `error-context.md` from the uploaded artifact paths. Cheapest; keeps screenshots/videos/traces, which have the same exposure through a different medium (the password field renders masked, but the trace records the `fill()` argument).
   - Have the fixture authenticate via `POST /api/auth/sign-in/email` through `page.request` so the password never reaches an input element. Costs the browser-level sign-in-form coverage.
   - Make the whole staging artifact set non-public. GitHub has no per-artifact visibility control, so in practice this means not uploading them from a public repo.
   - The trace path (`trace: 'on-first-retry'`) needs its own answer regardless of which of the above is chosen.
4. Consider whether the staging test account should be scoped so a leak of it is uninteresting (no real data, no shared password, no reuse anywhere).

## Acceptance criteria

- [ ] Staging test credential rotated; old value invalid against `https://dev.autoauthor.app`.
- [ ] Existing artifacts containing the plaintext value deleted.
- [ ] A staging E2E failure inside the auth fixture no longer produces an artifact — snapshot, screenshot, video, or trace — from which the password can be read. Demonstrated on a deliberately failed sign-in.
- [ ] The chosen mitigation is documented in `frontend/tests/e2e/staging/README.md` so it is not undone by the next person who wants richer failure artifacts.

## Related

- #551 — the reason sign-in failed on every run, making this exposure continuous rather than occasional. Fixed in #598, which reduces the frequency but does **not** close this: any future auth-fixture failure re-exposes the value.
- #558 — local `.env.test` credentials are stale, which is the same credential from the other direction.

frankbria:
Closes #599. Closes #558.

## The leak

This repo is **public**, and `e2e-staging-tests.yml` uploads `test-results/` as a CI artifact any GitHub user can download. **Artifacts are not secret-masked — GitHub masks `secrets.*` in job logs only.** Four things carried live credentials out:

| artifact | what it exposes |
|---|---|
| `error-context.md` | its page snapshot renders each input's **value**, so the staging test password appears in plaintext |
| `trace.zip` | records request **headers**, so every authenticated call carries `Cookie: __Secure-better-auth.session_token=…`; also records action arguments, so the password passed to `fill()` is stored verbatim |
| `staging-results.json` | the `json` reporter captures test stdout, and the fixture logged the account email — so `secrets.TEST_USER_EMAIL` went out verbatim. **Artifacts are not secret-masked; only job logs are.** Found by the security review, confirmed against a real local file |
| the HTML report | inlines every failure attachment, reproducing `error-context.md` |

Because #551 made sign-in fail on **every** run, this was continuous rather than occasional. I read the password out of a public artifact myself while investigating that issue — which is how this was found.

## Already done, out of band

The containment half is complete and does not live in this diff:

| step | evidence |
|---|---|
| Password rotated | `change-password` → 200, new value 32 chars (was 15) |
| Prior sessions revoked | rotated with `revokeOtherSessions: true`, so every token in an old artifact is dead |
| Old password invalid | `sign-in/email` with the burned value → **401** |
| New password valid | same endpoint, new value → **200** |
| `staging` env secret | `TEST_USER_PASSWORD` updated |
| Local `.env.test` | updated — this also clears **#558** |
| Artifacts purged | all **63** unexpired staging artifacts deleted; re-queried after → **0 remaining** |

## What this diff changes

Rotation alone would have re-leaked the *new* password on the next auth-fixture failure. This drops everything that carries a credential and keeps the two that do not:

- **`trace: process.env.CI ? 'off' : 'retain-on-failure'`**, with the reason stated at the setting so it does not read as a debugging preference someone can flip back. Off where artifacts are published; on locally, where nothing is, so there is still a real way to diagnose a failure.
- **The fixture no longer logs the account email**, and **`error-context.md`, `trace.zip` and `staging-results.json` are excluded** from the upload. The **HTML report step is removed**, and so is the duplicate `if: failure()` upload step that shipped the same two trees under a second name. The exclusions are deliberately redundant with the config setting: re-enabling one must not be enough to re-open the leak.
- **Jest guards** that fail if `trace` stops being CI-off, or if the npm script starts passing `--trace` (a CLI flag overrides the config outright). Mutation-checked — each kills exactly its own guard.
- **Policy documented** in `frontend/tests/e2e/staging/README.md`, next to the artifacts it deliberately keeps, per this issue's last acceptance criterion.

Screenshots and video are kept: a `type="password"` input renders masked, so neither carries the credential. The README warns that a spec clicking the show-password toggle would end that, since nothing enforces it.

## Also: three real email addresses scrubbed

Found while rotating. On a public repo:

- `docs/demos/2026-08-27-…-staging-verification.md` — two plus-addressed throwaway accounts from the #550 reproduction.
- `backend/app/populate_db_test_data.py` — the maintainer's address hardcoded as the dev seed user. It is a local seed helper, omitted from coverage and imported by nothing.
- `archive/TODO.md` — a status line naming the same address. Touched despite the read-only convention for `archive/`, because removing personal data from a public repo is not the same as rewriting history.

Verified by pattern rather than by those three known lines: no real address remains in any tracked file. What is left is placeholders (`example.com`, `admin@domain.org`, `a@b.com`), project addresses (`support@autoauthor.com`), and npm author metadata in `package-lock.json`.

## Cost, stated plainly

**No trace from a CI run**, and no HTML report or results JSON published. A manual `workflow_dispatch` re-runs the same CI config, so it confirms whether a failure reproduces but does **not** help you diagnose it — an earlier draft of this PR claimed otherwise and was wrong.

The real diagnostic path is running the suite locally, where nothing is uploaded and `trace` is `'retain-on-failure'`. The thrown sign-in error also carries the page's own alert text (added in #598), which covers the common cases without a trace at all.

## The exclusions are demonstrated, not asserted

Ran the real `@actions/glob` 0.7.0 — the library `upload-artifact` uses — over a fixture tree mirroring both include roots, with the exact `path:` block from the workflow. Only `screenshot.png`, `test-failed-1.png` and `video.webm` survive; all three excluded filenames are dropped from **both** roots, confirming one `!**/name` covers both. Mutation-checked: dropping `!**/error-context.md` makes it reappear from both roots. Output in the review comment.

## Also closes #558 — and it nearly bit this very PR

Rotating the credential walked into the trap #558 describes, in a variant it did not name. The generated 32-character password contained a `#`, and `dotenv` — which `playwright.config.ts` uses to read `.env.test` — treats an unquoted `#` as an inline comment. **It parsed down to 4 characters.** Local staging E2E would have failed to authenticate with no hint why, while CI stayed green because it reads env vars rather than the file. I would have shipped that while claiming #558 was incidentally fixed.

Both values are now single-quoted and verified end to end: dotenv parses the full 32 characters, and signing in with the dotenv-parsed values returns 200. `.env.test.example` and the README now say where the live values come from (the `staging` environment secrets, which cannot be read back) and why quoting is mandatory, naming both failure modes. Proven, not asserted: a value containing both `$` and `#` in the documented form survives `set -a; . file` and `dotenv.parse` intact. That closes all four of #558's acceptance criteria.

## Nothing left dangling — every open item is filed

| item | tracked as |
|---|---|
| Prove the exclusions on a real failed staging run; guard the show-password invariant; redact the Locust report | **#604** |
| Scope the test account so a future leak is uninteresting (#599's AC 4); delete the five `+aa550-*` throwaway accounts | **#603** |
| better-auth may rate-limit sign-in **globally** rather than per-IP — `getIP` returns null on a multi-hop `x-forwarded-for` with no `trustedProxies` | **#602** |

Two residuals are accepted rather than filed, and stated so they are not mistaken for oversights: the **scrubbed addresses persist in git history** (throwaway plus-aliases; rewriting public history is disproportionate), and the **jest guards are tripwires, not controls** — they catch an accidental revert, not a determined edit, which is labelled in the test file.

## Known limitations

- **Not verified end-to-end against a real failing staging run.** The glob proof above covers the upload filter precisely, but nobody has deliberately failed a staging test and inspected the published artifact. Tracked as #604.


frankbria:
## Third-party security review — opencode (GLM 5.2)

Run against the first commit (`55e989b`). It found **two Majors, both real**, and both are fixed in the follow-up commit. Full output:

> `.env.test` and all results/report trees are gitignored (good). The auth fixture logs only the email and cookie *name* — but that stdout lands in `staging-results.json`, which is uploaded. Let me check the JSON report's actual contents, the specs, the sign-in form, and verify the upload-artifact exclusion semantics for the pinned version.Two quick final checks — the Locust load-smoke (its HTML report is also uploaded from this workflow) and whether any spec touches the show-password toggle:## Verdict
>
> The diff closes both original vectors (password via `error-context.md`/HTML report, session token via `trace.zip`), and the exclusion patterns are correct for the pinned action. But the same artifact still publishes **`secrets.TEST_USER_EMAIL` in plaintext on every run** via the JSON reporter, and the "replacement" debugging guidance points at mechanisms that produce no traces.
>
> ---
>
> ### 1. Are both vectors closed? Mostly — one public path remains
>
> Closed, verified against the repo:
>
> - **`trace.zip`**: never generated (`trace: 'off'`, playwright.config.ts:69) and excluded from both uploads anyway. Double-closed.
> - **`error-context.md`**: still generated by Playwright core on failure (independent of trace/reporter settings — the exclusion is genuinely load-bearing), but excluded from both uploads, and the report that inlined it is gone.
> - **video/screenshots**: the sign-in form is `type={showPassword ? "text" : "password"}` (src/app/auth/sign-in/page.tsx:181) and nothing in the suite clicks the toggle (verified) — password renders masked. Residual risk: a future spec touching the toggle publishes it; nobody guards that.
> - **`GITHUB_STEP_SUMMARY`**: static strings + `job.status` + worker count only. Safe.
> - **Job logs** (public): `list`/`github` reporters print test stdout — including `console.log('🔐 Logging in as ${email}...')` (fixtures/auth.fixture.ts:54) — but GitHub masks `secrets.*` values in **logs**, so it renders `***`. The fixture logs only the cookie *name*, never the value (confirmed in the real `staging-results.json`: zero hits for password/cookie values).
>
> **Remaining leak — Major**: the `json` reporter's `outputFile: 'test-results/staging-results.json'` (playwright.config.ts:50) sits **inside the uploaded tree**, and artifact uploads are **not** secret-masked. The actual local artifact file contains, in plaintext, in `stdout`: `"🔐 Logging in as frank.bria+test-auto-author@pm.me..."` — i.e. `secrets.TEST_USER_EMAIL`, published on every run (the step is `if: always()`). It's not a credential, but it's a GitHub Secret published verbatim, it's the exact `frank.bria+*@pm.me` pattern this same diff scrubs out of the docs, and it identifies the live staging account. Fix: redact the log line, or exclude `!**/staging-results.json`, or move `outputFile` out of the uploaded tree.
>
> Also in-workflow: the load-smoke job uploads `load-smoke-report.html`, and on sign-in ≥400 Locust embeds `response.text[:300]` (backend/tests/load/staging_smoke.py:51). better-auth error bodies don't echo the password, so no live leak — but a framework 422 echoing request fields would land in a public artifact. Hardening only.
>
> ### 2. Exclusion patterns — correct and effective
>
> Verified against the README at the pinned SHA: the action's own "Upload using Multiple Paths and Exclusions" example is exactly this shape — multiple include roots plus one `!path/**/*.tmp`. Matching is set-wise (upload iff matches an include AND no `!` pattern); `!` patterns root at the workspace cwd like the includes, and `**/` matches at any depth, so **one `!**/error-context.md` covers both `frontend/tests/e2e/staging/test-results/` and `frontend/test-results/`**. The leading `!` is safe because `path: |` is a literal block scalar (the README's quoting caveat applies to wildcard-leading flow scalars / YAML aliases). Caveat worth knowing: the exclusion is **filename-based** — if a Playwright rename ever changes `error-context.md`, it's missed; `trace: 'off'` remains the real control, which is exactly how the comment frames it. Correct design.
>
> ### 3. HTML reporter — no failure, cleanly orphaned
>
> Still configured (playwright.config.ts:48) and still written to `playwright-report-staging/` on the runner. Nothing fails: no remaining step references the removed artifact (the summary was updated; there's no download/merge-reports step), html doesn't auto-open under CI, and the folder sits outside both uploaded trees and `.gitignore`d. Cost is dead runner I/O only.
>
> ### 4. Jest guard — a tripwire, not a control
>
> Meaningful against the accidental revert: a config-only flip goes red until someone edits a test literally named "keeps traces off so artifacts cannot carry credentials" — visible and explainable in review. Trivially defeated intentionally: a later spread (`trace: 'off', ...overrides` still satisfies the anchored regex), a CLI `--trace on` in the npm script (the guard reads only the config file), or reformatting (double quotes / dropped trailing comma = false positive). Cheap hardening: assert the workflow's run step has no `--trace`, and count `trace:` occurrences (catches `{trace:'on'}` hidden in a spread).
>
> ### 5. Other findings
>
> - **Vacuous debugging story — Major**: README says local runs keep "traces available" — false, the config is `trace: 'off'` unconditionally and the guard *forbids* an env-conditional. Both README and the step summary say "reproduce with a manual workflow_dispatch" — but dispatch runs the same config; its only inputs are workers and load-smoke. **No path anywhere produces a trace anymore.** Fine as a security decision; then the claims should say that instead of pointing at diagnostics that don't exist.
> - **Email scrubs**: scrubbing a commit scrubs nothing from git history — the addresses stay retrievable in a public repo forever (throwaway aliases, already accepted; worth stating in #599). But the scrub's goal is undercut by the `staging-results.json` stdout leak above, which republishes a `frank.bria+*@pm.me` alias every run.
> - **Stale comments**: playwright.config.ts:32 ("retries stay on so … still yields a trace") and the jest test name at StagingE2eFlakyGate.test.ts:25 ("keeps retries so the retry trace is still captured") now reference traces that no longer exist.
> - `populate_db_test_data.py`: fine — `example.invalid` is non-routable; the password belongs to a user that can't exist.
> - The `if: failure()` upload step is fully subsumed by the `if: always()` one (identical trees, different artifact name). Pre-existing, but this diff edits both.
>
> ---
>
> ### Ranked
>
> | Severity | Finding |
> |---|---|
> | Critical | none remaining — both credential vectors closed; password already rotated |
> | Major | `staging-results.json` uploaded with `secrets.TEST_USER_EMAIL` in plaintext `stdout` on every run (auth.fixture.ts:54 + config:50) |
> | Major | Debugging replacement is fiction: local traces are off, `workflow_dispatch` adds no diagnostics |
> | Minor | Future show-password-toggle spec would publish the password in video/screenshots (unguarded) |
> | Minor | Jest guard defeatable via spread/CLI/formatting; no `--trace` guard on the run command |
> | Minor | Locust embeds `response.text[:300]` into the uploaded HTML on auth failure (staging_smoke.py:51) |
> | Minor | Exclusions are filename-based; stale "retry trace" comments; duplicated upload steps; scrubbed emails persist in git history |

### Triage

| Finding | Action |
|---|---|
| **Major — `staging-results.json` publishes `secrets.TEST_USER_EMAIL` every run** | **Fixed.** Verified against a real local `staging-results.json` before acting — it does contain `Logging in as frank.bria+test-auto-author@pm.me`. GitHub masks secrets in logs, not artifacts. Fixed at both ends: the fixture no longer logs the address, and `!**/staging-results.json` is excluded from the upload. The reviewer's point that this undercut the doc scrub in the same commit is exactly right. |
| **Major — the debugging replacement is fiction** | **Fixed.** My README claimed local runs "keep traces available" while `trace` was `'off'` unconditionally, and pointed at `workflow_dispatch`, which runs the same config. Now `process.env.CI ? 'off' : 'retain-on-failure'` — local runs genuinely produce a trace, and nothing is uploaded locally so it costs nothing. `retain-on-failure` rather than `on-first-retry` because local `retries` is 0. README and step summary corrected. |
| Minor — show-password toggle would publish the password in screenshot/video | **Documented, not enforced.** Added an explicit warning in the README next to the artifacts that are still uploaded. |
| Minor — guard defeatable via CLI `--trace` | **Fixed.** New guard asserts `test:e2e:staging` passes no `--trace`. The spread/formatting evasions remain possible; the guard is a tripwire against accidental revert, not a control against a determined edit, and it is labelled as such. |
| Minor — duplicated `if: failure()` upload step | **Removed.** It uploaded the identical two trees under a second name, fully subsumed by the `always()` step, and was a second place to forget an exclusion. |
| Minor — stale "retry trace" comments | **Fixed** in the config and the jest test name. |
| Minor — Locust embeds `response.text[:300]` into an uploaded HTML | **Not fixed.** The reviewer confirms no live leak (better-auth error bodies do not echo the password); it is hardening against a hypothetical framework 422. Out of scope here — noted on #599. |
| Minor — scrubbed emails persist in git history | **Accepted and stated.** They are throwaway plus-aliases; rewriting public history is disproportionate. Recorded on #599. |
| Minor — exclusions are filename-based | **Accepted by design.** If Playwright renames `error-context.md` the exclusion misses it, which is precisely why `trace: 'off'` under CI is the primary control and the exclusions are the second layer. The reviewer reached the same conclusion. |

### The exclusions are now demonstrated, not asserted

I originally shipped these as "correct per glob semantics", which is the same
unverified-gate mistake I recorded a lesson about earlier today. So I ran the real
`@actions/glob` 0.7.0 — the library `upload-artifact` uses — over a fixture tree
mirroring both include roots, with the exact `path:` block from the workflow:

```
UPLOADED:
  frontend/test-results/other-test/screenshot.png
  frontend/tests/e2e/staging/test-results/some-test/test-failed-1.png
  frontend/tests/e2e/staging/test-results/some-test/video.webm

LEAKY FILES STILL UPLOADED: NONE
```

All three excluded filenames are dropped from **both** include roots, confirming
one `!**/name` covers both. Mutation-checked — dropping `!**/error-context.md`
makes it reappear from both roots:

```
LEAKY FILES STILL UPLOADED: [
  'frontend/test-results/other-test/error-context.md',
  'frontend/tests/e2e/staging/test-results/some-test/error-context.md'
]
```

PR #605, opened 2026-09-07
fix(ci): skip bot PRs in GLM Review, and re-pin off the July callee

### LINKED ISSUE #519: [P3.26] glm-review runs on Dependabot PRs and always fails for want of a withheld secret, leaving every bot PR permanently red
## Problem

The `review / review` check **fails on every Dependabot PR**, and has for as long as the workflow has existed. All nine PRs merged in the 2026-08-25 sweep that had frontend changes showed it red (#507, #508, #509, #510, #511), as did #405, #437, #438 earlier.

It is never a finding. The job cannot authenticate, because GitHub withholds secrets from `pull_request` runs triggered by Dependabot, so `ZHIPU_API_KEY` arrives empty and the review exits non-zero.

## Why the existing guard doesn't catch it

`.github/workflows/glm-review.yml:19-23` guards on head repo and change size:

```yaml
if: |
  github.event.pull_request.head.repo.full_name == github.repository &&
  (github.event.pull_request.changed_files >= 5 ||
   github.event.pull_request.additions >= 20 ||
   github.event.pull_request.deletions >= 20)
```

Dependabot branches live **in the repo**, not in a fork, so `head.repo.full_name == github.repository` is true and the same-repo guard passes. A lockfile bump trivially clears the 20-addition threshold (#509 had 433 added lines). So the job runs, and then fails for want of a secret it was never going to receive.

## Why it is worth fixing rather than tolerating

A check that is always red is a check nobody reads. It sits next to the four genuinely required checks (`Frontend Tests`, `Backend Tests`, `E2E Tests (Playwright)`, `Security Audit`) and makes `mergeStateStatus` report `UNSTABLE` instead of `CLEAN` on bot PRs — which means any automation gating on "is this PR clean" has to special-case it, and any human triaging a batch has to remember that this particular red is the acceptable one. That is exactly the habit that lets a real failure through.

It also has no upside: an LLM review of a machine-generated lockfile diff adds nothing that the supply-chain triage doesn't already cover.

## Proposed fix

Add an actor guard to the existing `if:`:

```yaml
if: |
  github.actor != 'dependabot[bot]' &&
  github.event.pull_request.head.repo.full_name == github.repository &&
  (github.event.pull_request.changed_files >= 5 ||
   github.event.pull_request.additions >= 20 ||
   github.event.pull_request.deletions >= 20)
```

The job then reports **skipped** rather than failed, and bot PRs reach `CLEAN`. Keep every existing clause — the same-repo guard closes the pwn-request class from #270/#191 and must not be weakened while adding this.

## Acceptance criteria

- [ ] The next Dependabot PR shows `review / review` as **skipped**, not failed, and `mergeStateStatus` is `CLEAN`.
- [ ] A normal human PR of >= 5 files still triggers the review and posts its comment — verified on a real PR, not just by reading the condition.
- [ ] Fork-PR behaviour unchanged (still guarded).

## Related

Noted as known noise in the triage on #507/#508/#511. #517 (P2.19) covers the grouping/auto-merge side of the same Dependabot pipeline.

### LINKED ISSUE #547: [P3.26.1] glm-review is cancelled at ~20m on large diffs and posts an empty review that reads as completed

frankbria:
Closes #519. Closes #547.

12 of the last 20 GLM Review runs here failed, most in about 8 seconds with:

```
Action failed with error: Workflow initiated by non-human actor: dependabot (type: Bot).
Add bot to allowed_bots list or use '*' to allow all bots.
```

This is cause (2) in [glm-review#9](https://github.com/frankbria/glm-review/issues/9). `claude-code-action` refuses any run whose actor is not a `User`. This repo gets heavy Dependabot traffic and a lockfile bump clears the size gate easily, so each one painted a red check on a PR that was never reviewed.

## Why skip rather than `allowed_bots`

The issue suggested passing `allowed_bots` as a one-liner. It would clear the guard — `isAllowedBot` lowercases and strips a trailing `[bot]`, so `dependabot` matches — but it would not produce a review.

GitHub runs Dependabot-triggered `pull_request` workflows with a **read-only token** and the **separate Dependabot secret store**. So `ZHIPU_API_KEY` would arrive empty, and the summary comment would be unpostable even if it didn't. It trades an 8-second bot-guard failure for a credential failure a few minutes in.

It also isn't worth wanting: a bug-hunting reviewer scoped to logic errors, races and injection has little to say about a dependency bump — which is the kind of trivia the size gate already exists to filter. Skipping reports these grey and honest.

Testing `github.event.pull_request.user.type == 'User'` covers every bot rather than naming Dependabot.

## Re-pin while here

`b877d15a` capped runs with a 20-minute **job** timeout, which cancels every remaining step — `if: always()` included — so a killed review uploaded no transcript and left its "Reviewing…" stub standing, indistinguishable from a review that ran and found nothing. `77dfa9c` moves the clock onto the review step, makes the turn cap the limit that normally binds, uploads the transcript, and rewrites the stub into an explicit "did not complete" notice.

`issues: write` and `id-token: write` go with it. They were **not** optional on the old pin — GitHub fails a run whose caller grants less than the called job requests — but the new callee declares neither. Sibling repo `narrative-modeling-app` is currently at 20/20 `startup_failure` for exactly this reason in reverse, so the comment spells out that the pin and the permissions block move together.

## Unchanged

The same-repo head guard from #270 is untouched and still evaluated first — fork PRs still never reach a run that binds `ZHIPU_API_KEY`. The new bot condition is ANDed after it, so it can only narrow what runs, never widen it. The pinning policy is preserved: still an immutable SHA, not `@main`.

## Verification

`actionlint` passes on the changed file. This PR is human-authored, so its own review runs normally — which exercises the new condition on the true branch, together with the new pin and the reduced permissions block, since `pull_request` builds the workflow from the PR's own branch. Run [34084069923](https://github.com/frankbria/auto-author/actions/runs/34084069923): **success in 8m28s**, real review posted.

Full acceptance-criteria mapping for both issues is in the demo comment; the cross-family review and its triage are in the review comment.

## Added during review (`e047373`)

Cross-family review (opencode/GLM) found **no Critical and no live Major**, and one real latent gap:

`claude-code-action` keys off `github.actor`, which is the PR *author* on `opened` but the **pusher** on `synchronize`. Gating on `pull_request.user.type` alone would therefore still go red on a human-authored PR that a bot pushed to — an autofix job, a rebase app, `github-actions[bot]` — and the size gate cannot save it, because `changed_files` and `additions` are cumulative PR totals, so a one-line bot push to an already-qualifying PR re-fires the trigger.

`sender.type == 'User'` is the account the action actually resolves, so both are now tested. It is a strict narrowing, so the #270 fork guard is untouched and still fails closed. The gap was latent, not live: no workflow in this repo requests `contents: write` or pushes to a PR today. The in-file comment now says what the guard covers rather than implying it mirrors the actor rule.

The changelog entry, which the branch was missing, is added in the same commit.

## Not done here — filed as #607

- **Neither fix has been observed live.** The bot-skip effect is demonstrated by evaluating the `if:` against real payloads (8 of 8 Dependabot PRs flip red → skip), but that re-implements the expression rather than letting Actions evaluate it — `act` is not installed here, so the #191/#270 truth-table precedent could not be followed. #519's own AC needs a Dependabot PR clearing the size gate; the two open now (#586, #587) are below it.
- **#547's fourth AC** — confirm legibility on a large-diff PR — is unexercised by a one-file diff.
- **Draft PRs** burn a full review, and a draft→ready PR is never reviewed at all (no `ready_for_review` trigger).
- **Two notes recorded, no action needed today**: GLM Review is not a required check (skipped would never satisfy classic branch protection if it were), and review transcripts are public 14-day artifacts containing the diff and model turns.

https://claude.ai/code/session_01UdSq9dMfiLL12RHcmcPER7

---
_Generated by [Claude Code](https://claude.ai/code/session_01UdSq9dMfiLL12RHcmcPER7)_


frankbria:
## Cross-family review — opencode (GLM 5.2)

This PR entered outside the Phase 7 pipeline, so an independent cross-family review was run on the full diff. **No Critical, no live Major.** One real latent gap, fixed in `e047373` before merge. Full output:

> The working tree has the post-change file on branch `fix/glm-review-skip-bots` — and the `on:` block is `pull_request` (not `pull_request_target`), types `[opened, synchronize]`. That matters for Q2. Now checking sibling workflows for bot-pushers, the Dependabot config, branch protection, and spot-checking the callee at both pins:Callee confirmed at both pins — b877d15a declares `issues: write` + `id-token: write` with a 20m job cap; 77dfa9c declares only `contents: read`/`pull-requests: write` with 35m step/40m job, transcript upload, and the failure-notice step. Also noted: "GLM Review" is **not** in required status checks, and the concurrency group lives at the callee's *top level*. Now checking for bot-pushers in this repo's workflows, Dependabot config, actual run overlap, and the action's actor check:Empirical results: no bot-pushers exist in any auto-author workflow; required checks don't include GLM Review; the cancelled-run chains on branches #599/#551 prove the callee's top-level concurrency group (`cancel-in-progress`) is actively firing; and pre-change Dependabot runs failed in 12–15s (33999079116, 33999083715) — the exact symptom this change targets. One last check — the exact actor validation in claude-code-action:All evidence is in. Here's the assessment.
>
> ## 1. Security — no weakening
>
> A conjunction can only shrink the passing set: `{same-repo ∧ User ∧ size} ⊆ {same-repo ∧ size}`. No PR that previously couldn't reach `ZHIPU_API_KEY` can now — structurally impossible. Precedence is safe (the size disjunction stays parenthesized; the block scalar folds to `A && B && (C || D)`), and `user.type` is server-assigned account metadata, not attacker-choosable text. A null `user` degrades to `null == 'User'` → false → skip: fail-closed, same as the deleted-fork case the existing guard documents. Context note: the trigger is `pull_request` on a public repo (glm-review.yml:3-4), so fork PRs never received secrets anyway — the same-repo check is defense-in-depth against trigger drift, and this change doesn't touch it.
>
> ## 2. The `user.type` vs actor gap — real, but latent here
>
> Verified in claude-code-action (`src/github/validation/actor.ts`): it resolves **`github.actor`** via the Users API and hard-fails unless `type == "User"` (glm-review passes no `allowed_bots`; the 12–15s failures in runs 33999079116/33999083715 match the API roundtrip). On `synchronize`, `github.actor` is the **pusher**, not the author — so a human PR plus a bot push (autofix CI, rebase app, `github-actions[bot]`) passes your gate and dies with the same red. The size gate won't save you: `changed_files`/`additions` are cumulative PR totals, so a one-line bot push to an already-gated PR re-fires the trigger.
>
> **But**: I checked every workflow in this repo — none has `contents: write`, git-auto-commit, or any push-to-PR step — and Dependabot only pushes its own PRs (author = Bot, covered). So the gap fires zero times today; it is one autofix-workflow away. Cheap exact fix: add `github.event.sender.type == 'User' &&` (sender = pusher on `synchronize`, opener on `opened` — the same account the action resolves). Keep the author conjunct too if you also want Dependabot PRs skipped when a human amends them (author-gating skips those runs even though they'd work — harmless, but it means the comment's actor-based justification at glm-review.yml:20-27 doesn't strictly match the author-based gate).
>
> ## 3. Other `type` values — nothing wrongly excluded
>
> Public personal repo ⇒ PR authors are only ever `User` or `Bot`. Organizations can't author PRs (no credentials); `Mannequin` is Enterprise-import-only; EMU accounts surface as `User` in REST payloads; `ghost` (deleted author) is `User`. Any exotic value that did appear fails toward *skip* (missed review), never toward secret access. No correction needed.
>
> ## 4. Permission drop — mandatory, not just safe; regression is loud
>
> Verified both pins: b877d15a's called job declared `issues: write` + `id-token: write`; 77dfa9c declares exactly `contents: read` + `pull-requests: write` and passes an explicit `github_token`, so no OIDC path. Since GitHub enforces caller-grant ≥ callee-declared at startup, the caller block is now a **hard ceiling**: a future compromised pin physically cannot mint OIDC tokens or write issues under this caller — that's the actual security win. Re-pinning backward to b877d15a without restoring the grants makes every run a `startup_failure` — fail-closed, noisy, zero silent escalation; the in-file warning (glm-review.yml:59-60) is the right mitigation. The new pin has already run green on its own PR (run 34084069923, ~8.5 min), so the dogfood check exists.
>
> ## 5. Ranked findings
>
> **Critical:** none. **Major:** none live.
>
> **Minor:**
> 1. **Actor/author mismatch (Q2)** — latent gap + the comment overclaims coverage. Add the `sender.type` conjunct, or amend the comment to say "bot-authored PRs" rather than implying it mirrors the action's actor rule.
> 2. **Draft PRs** — `opened` fires on drafts; a sizeable draft burns a 35m/60-turn review, and with no `ready_for_review` type, a draft→ready PR is never reviewed at all. Add `!github.event.pull_request.draft` + `ready_for_review`, or accept.
> 3. **Required-checks future** — GLM Review is not in `required_status_checks` today (verified: Frontend/Backend/E2E/Security Audit only), so skips can't strand merges. If it's ever added, classic branch protection doesn't count skipped as satisfied — Dependabot PRs would sit "Expected" forever. Keep in mind, no action now.
> 4. **Transcript artifacts** — `upload_transcript` defaults true ⇒ 14-day artifacts in a public repo containing the diff, opened files, and model turns. No secrets reach the prompt; fine while the repo is public. `upload_transcript: false` is the knob.
> 5. **Non-issue, verified** — concurrency comes from the callee's top-level group and demonstrably fires (the cancelled-run chains on #599/#551 are `cancel-in-progress` cancelling superseded reviews). Just know that property lives in glm-review, not this file.
>
> Solid change overall: ship it. Want me to open the `sender.type` + draft-PR follow-up as a prioritized issue?

### Triage

| Finding | Action |
|---|---|
| **Minor 1 — actor/author mismatch; the comment overclaims** | **Fixed in `e047373`.** Verified the claim independently: the action resolves `github.actor`, which is the pusher on `synchronize`, not the author. `sender.type == 'User'` added — it is the account the action actually resolves — and the comment now states what the guard covers instead of implying it mirrors the actor rule. Strict narrowing, so the #270 fork guard is untouched. |
| Minor 2 — draft PRs burn a review; no `ready_for_review` type | **Filed as #607.** Genuinely out of this PR's scope (it is about bots and the pin), and it changes when reviews happen, which deserves its own decision. |
| Minor 3 — GLM Review is not a required check today | **No action, confirmed.** Independently checked: branch protection requires only Frontend Tests, Backend Tests, E2E Tests (Playwright), Security Audit. Recorded in #607 so that if it is ever promoted, the "skipped never satisfies classic branch protection" trap is already written down. |
| Minor 4 — transcript artifacts are public for 14 days | **Noted in #607.** Same category as #599, so worth stating explicitly: the transcript holds the diff and model turns, no secrets, on an already-public repo. `upload_transcript: false` is the knob if that changes. |
| Minor 5 — concurrency lives in the callee, not this file | **No action, and it corrected me.** See below. |

### One correction the review earned

I had attributed the recent `cancelled` GLM runs on my own feature branches to #547's timeout. Measuring them says otherwise:

```
179s  305s  337s  373s  413s  560s  595s  842s   <- cancel-in-progress, superseded by my own pushes
1220s 1237s 1240s                                <- ~20m20-40s: the #547 signature
711s                                             <- "Claude encountered an error", neither
```

Most were my own pushes superseding each other via the callee's top-level concurrency group. Only the three clustered at 20m20–40s match #547's documented 20m24s / 20m27s. #547's evidence stands, on those three — but the larger number I would have cited was wrong.

frankbria:
## Demo — acceptance criteria mapped to outcome evidence

This PR closes two issues. Both are workflow-config changes, so the evidence is run conclusions and payload evaluation rather than a UI walkthrough.

### #519 — "The next Dependabot PR shows `review / review` as skipped, not failed, and `mergeStateStatus` is CLEAN"

**Baseline confirmed.** The failure is on the record for the Dependabot branches still open today:

```
dependabot/npm_and_yarn/frontend/minimatch-3.1.5      33999069478  failure
                                                       33995655901  failure
dependabot/.../frontend-prod-7257b05964                33999083715  failure
```

**Effect, evaluated against real payloads.** All eight Dependabot PRs the issue cites flip from *ran → red* to *skips*, while a human PR is unaffected:

| PR | author | type | files | + | − | OLD `if:` | NEW `if:` | effect |
|---|---|---|---|---|---|---|---|---|
| 605 | frankbria | User | 1 | 38 | 5 | True | True | runs |
| 507 | dependabot[bot] | Bot | 2 | 33 | 30 | True | **False** | was red → skips |
| 508 | dependabot[bot] | Bot | 2 | 112 | 128 | True | **False** | was red → skips |
| 509 | dependabot[bot] | Bot | 2 | 431 | 71 | True | **False** | was red → skips |
| 510 | dependabot[bot] | Bot | 2 | 67 | 68 | True | **False** | was red → skips |
| 511 | dependabot[bot] | Bot | 2 | 43 | 10 | True | **False** | was red → skips |
| 405 | dependabot[bot] | Bot | 2 | 178 | 76 | True | **False** | was red → skips |
| 437 | dependabot[bot] | Bot | 2 | 220 | 75 | True | **False** | was red → skips |
| 438 | dependabot[bot] | Bot | 2 | 10248 | 7242 | True | **False** | was red → skips |

**8 of 8.** Plus the two cases added in `e047373`:

| case | OLD | NEW | outcome |
|---|---|---|---|
| human PR, **bot** push (autofix / rebase app) | True | **False** | was red → skips |
| fork PR (#270 pwn-request guard) | False | False | skips, unchanged |

**The true branch is confirmed by GitHub, not by replay.** This PR is human-authored, so its own `review / review` had to run — and it did: run [34084069923](https://github.com/frankbria/auto-author/actions/runs/34084069923), **success in 8m28s**, posting a real review ("no defects found"). Because `pull_request` builds the workflow from the PR's own branch, that run already exercised the new condition, the new pin, and the reduced permissions block together.

### #547 — four criteria

| Criterion | Evidence |
|---|---|
| Set an explicit `timeout-minutes` so the bound is the repo's, not an opaque internal one | ✅ Callee at the pinned SHA: `timeout-minutes: 40` on the job, **35 on the review step**, with `--max-turns 60` as the limit that normally binds. Read from `frankbria/glm-review` at `77dfa9c`, not taken from the PR description. |
| A cancelled/timed-out review must not post a comment that reads as completed | ✅ The callee's `Report an incomplete review` step (`if: always() && steps.review.outcome != 'success'`) rewrites the tracking stub into **"⚠️ GLM review did not complete"**, deriving the reason from the result subtype (`error_max_turns`, `error_during_execution`, killed-before-result, …). |
| Decide what a non-completing review does to the check | ✅ Stays **red** — the review step has no `continue-on-error`, so the job fails — and is now distinguishable from a finding by the notice above. That is the option the issue called defensible. |
| Confirm on a large-diff PR that the new behaviour is legible | ❌ **Not done.** This PR's diff is one file; it cannot exercise the timeout path. Filed as #607. |

Also verified independently of the description: the pin `77dfa9c` is **identical to** `glm-review` `main` (immutable-pin policy intact), the callee declares exactly `contents: read` + `pull-requests: write` (so dropping `issues: write` / `id-token: write` is correct, and GitHub's caller-grant ≥ callee-declared rule is satisfied), the transcript uploads with `if: always()`, and `actionlint` exits 0 on the changed file.

### Honest limits

- **The bot-skip branch has not been observed live.** The table above evaluates the `if:` expression against real payloads, but it is a re-implementation in Python, not GitHub evaluating it. The repo's precedent for this (#191, #270) was an `act -n` truth table; `act` is not installed on this machine — only its config remains. The precedence subtlety that #270 called load-bearing is preserved: the size disjunction stays parenthesized and the new terms are ANDed, which can only narrow. Live confirmation needs one Dependabot PR that clears the size gate — tracked in #607.
- The two Dependabot PRs open right now (#586, #587) are **below** the size gate (9 and 5 additions), so they already skip for an unrelated reason and cannot demonstrate this.

PR #608, opened 2026-09-07
fix(auth): bucket sign-in rate limits per client, not globally (#602)

### LINKED ISSUE #602: [P2.23] better-auth may rate-limit sign-in globally rather than per-IP: getIP returns null on a multi-hop x-forwarded-for with no trustedProxies
## What

better-auth's rate limiter buckets by client IP. `@better-auth/core/utils/ip` `getIP()` resolves that IP from `x-forwarded-for` (the only header in `DEFAULT_IP_HEADERS`), and `getIPFromHeader()` has this branch:

```js
if (trustedProxies.length > 0) { /* walk the chain from the right */ }
if (forwardedIps.length !== 1) return null;   // <-- more than one hop, no trustedProxies
```

`frontend/src/lib/auth.ts` configures no `advanced.ipAddress.trustedProxies` and no `ipAddressHeaders`. So **if staging's nginx emits an `x-forwarded-for` with more than one hop, `getIP()` returns `null`** and every request falls into `NO_TRUSTED_IP_KEY` — a single shared per-path bucket.

The rate limiter then applies its `/sign-in*` special rule — **window 10s, max 3** (`getDefaultSpecialRules()`) — to *all users at once* rather than per client.

## Why it matters

If the bucket really is global, three sign-in attempts from anywhere in the world within ten seconds lock out every other user for the rest of the window, and better-auth's window only resets after 10s of *silence*, so sustained traffic keeps it closed indefinitely. That is a trivially reachable denial of service on the login path, and it would look like "login is flaky" rather than like a limit.

better-auth itself logs a warning when this happens:

> Rate limiting could not determine a client IP and is falling back to a single shared per-path bucket. Ensure your runtime forwards a trusted client IP header, then set `advanced.ipAddress.ipAddressHeaders` or `advanced.ipAddress.trustedProxies` so the address can be resolved.

## Status: unconfirmed, and that is the first task

This was noticed while fixing #551, where the same rule was tripping the E2E suite from a single CI IP. Per-IP bucketing is fully consistent with what #551 observed, so **this may well be a non-issue** — a single-hop `x-forwarded-for` resolves fine. It could not be determined from outside the box which shape staging's nginx produces.

## Acceptance criteria

- [ ] Determine what `x-forwarded-for` actually reaches the Next.js container on staging — one hop or several. `docker logs` on the frontend container will show better-auth's warning above if resolution is failing.
- [ ] If resolution fails: configure `advanced.ipAddress.trustedProxies` (or `ipAddressHeaders`) in `frontend/src/lib/auth.ts` so the client IP resolves, and confirm the warning stops.
- [ ] Demonstrate the outcome either way: three rapid sign-ins from one address must **not** rate-limit a sign-in from a different address.
- [ ] If the deployment can ever put more than one proxy in front of the app, `trustedProxies` needs to be set regardless — record the decision.

## Related

- #551 — where this was found; the same 3-per-10s rule, hitting CI from a single IP.
- #598 — the fix for #551, which reduced the E2E suite to one sign-in per worker.

frankbria:
Closes #602.

## What was actually wrong

The issue filed this as *unconfirmed*, with "find out what reaches the container" as task one. It is real — but the trigger is one step off from the guess, and that difference is the whole fix.

better-auth's `/sign-in*` rule is **3 per 10s**, and its window reopens only after 10s of *silence*. When `getIP()` cannot resolve a client, every such request falls into a single `no-trusted-ip|<path>` bucket. The default reads `x-forwarded-for` only, and without `trustedProxies` a chain longer than one hop is unresolvable (the leftmost entry is attacker-controlled).

The trigger is that nginx builds the header with `$proxy_add_x_forwarded_for`, which **appends** rather than sets:

| Client sends | Container sees | `getIP()` |
|---|---|---|
| *(nothing)* | `<client-ip>` | resolves |
| `X-Forwarded-For: 203.0.113.7` | `203.0.113.7, <client-ip>` | **`null`** → shared bucket |

So ordinary browser traffic was fine — which is why the container, up since 2026-08-28, has never logged better-auth's warning. But **any** request carrying its own `X-Forwarded-For` lands in the shared bucket: reachable on demand by anyone who adds the header, and arriving unbidden for users behind a corporate proxy, VPN or carrier proxy.

**Confirmed on staging** (5 spoofed sign-ins, then a *different* claimed address, then a clean one):

```
spoof 203.0.113.7  x5 -> 401, 401, 401, 429, 429
spoof 198.51.100.9    -> 429   <-- different claimed client, same bucket
no XFF header         -> 401   <-- own bucket, unaffected
```

## The fix

`advanced.ipAddress.ipAddressHeaders: ["x-real-ip", "x-forwarded-for"]`.

nginx sets `X-Real-IP` with `proxy_set_header` from `$remote_addr`, which **overwrites** any client value — always one hop, unspoofable. The container is bound to `127.0.0.1:3002`, so nothing reaches it without that rewrite. `x-forwarded-for` stays second so a deployment with no nginx keeps today's behaviour.

**`trustedProxies` was rejected on the merits, not skipped.** nginx is the *last* hop and never appears in the forwarded chain, so any address listed there would never be matched — it would work only as a side effect of flipping the resolver into chain-walking mode. Configuration that works for a reason other than the one it states is worse than none.

## Outcome evidence

Real better-auth handler, real nginx header shape, attacker and victim on different addresses — config before and after:

```
### BEFORE — better-auth default
attacker 192.0.2.10 x5 -> 401, 401, 401, 429, 429
victim   192.0.2.20    -> 429     ==> LOCKED OUT by the attacker

### AFTER  — ipAddressHeaders: ["x-real-ip", "x-forwarded-for"]
attacker 192.0.2.10 x5 -> 401, 401, 401, 429, 429
victim   192.0.2.20    -> 401     ==> unaffected
```

The attacker still burns their own three per ten seconds — the rule doing its job. What changes is that the victim is no longer in it.

Full walkthrough, including the nginx config read off the box and a runnable repro: [`docs/demos/2026-09-07-issue-602-signin-rate-limit-per-client.md`](docs/demos/2026-09-07-issue-602-signin-rate-limit-per-client.md).

## A second, quieter finding

Enabling the guard test exposed that `jest.config.cjs`'s `transformIgnorePatterns` entry has **never done anything**. `transformIgnorePatterns` is OR'd across entries and next/jest *prepends* its own `/node_modules/...` entry, so a custom "don't ignore X" pattern can never un-ignore a package — the `better-auth` mocks were carrying that load. next's entry is patched instead, and the patch throws a named error if next/jest changes that string, rather than degrading into a cryptic ESM parse failure.

## Known limitations

- **A second proxy in front of nginx** (CDN, load balancer) makes `X-Real-IP` *that proxy's* address and re-collapses every user into one bucket. AC4 asked for this decision on the record: it is documented next to the nginx step in `docs/STAGING-DEPLOYMENT.md` with both exits — have the outermost proxy set `X-Real-IP`, or move to `trustedProxies` listing the chain. Not pre-emptively configured, because there is no second proxy today and a `trustedProxies` list of addresses that never appear in the chain is exactly the decoration rejected above.
- **The backend's own limiter is untouched, deliberately.** `backend/app/api/dependencies.py` keys per authenticated user and falls back to `request.client.host` (the proxy, behind nginx) — but that fallback is only reached if an authenticated caller has no id at all, which the dependency's contract prevents. Dead branch, not a second instance.
- **Staging verification is pre-merge only for the broken state.** The fixed state is demonstrated against the real handler in-process; re-running the staging probe after deploy is a post-merge step.

## Verification

- `frontend/src/lib/__tests__/auth-ip.test.ts` drives the **real** `getIP()` — a config-shape assertion would not have caught this. Mutation-checked: reverting to `['x-forwarded-for']` fails 3 of 5.
- Full frontend suite: **135 suites, 2298 passed**, 5 skipped. Coverage gate passed.
- `npm run typecheck` clean; `npm run lint` 0 errors (443 pre-existing warnings).
- Third-party review: `codex review --base main` — no findings. opencode/GLM produced zero bytes in 12 minutes and was abandoned per its known silent-hang failure mode.


frankbria:
## Third-party review (post-PR pass)

Diff unchanged since the pre-PR run (`HEAD == REVIEWED_SHA` for the code; only docs landed after), so this posts that run rather than re-running the reviewer.

**Reviewer:** `codex review --base main` (cross-family). **Findings: none.**

> The changes correctly configure better-auth IP resolution around the documented nginx behavior and add focused regression coverage. I did not identify any discrete introduced bug that would break existing functionality.

It was given the deployment context explicitly (single nginx hop, `X-Real-IP $remote_addr`, `X-Forwarded-For $proxy_add_x_forwarded_for`, container bound to loopback) and asked four pointed questions: whether reading `x-real-ip` first has a spoofing path, whether the `x-forwarded-for` fallback reintroduces a weakness, whether the `jest.config.cjs` patch is correct and non-fragile, and what could regress silently. It read the vendored `@better-auth/core` IP resolver and the `tsc --listFiles` output for the new test before answering.

**opencode/GLM was attempted first and abandoned**: zero bytes of output in 12 minutes on a 198-line diff, which matches its known silent-hang failure mode. One attempt, then the documented fallback.

PR #609, opened 2026-09-08
feat(dashboard): page through the book list instead of dumping 100 cards

### LINKED ISSUE #493: [P3.22] Dashboard renders a 100-book page with no pagination UI; first paint exceeds 15s on a large library
Surfaced while triaging #488.

## What happens

`GET /api/v1/books/` returns up to **100** books (`limit: int = Query(100, ge=1, le=100)`), and the dashboard renders every one as a card. There is **no pagination control in the UI**, so a user past 100 books simply cannot reach the rest.

Measured on staging, where the E2E account holds **2,702 books**:

- 100-book payload: **~195 KB** (~1.9 KB per book)
- 100 cards rendered in one list
- First paint exceeded **15 s** late in a suite run — an E2E spec waiting 15 s for its book to appear captured a snapshot with **zero cards rendered**

That spec has been moved to 20 s to match its siblings (#488), which treats the symptom. The underlying slowness is this issue.

## Why it is worth fixing

- **The remaining 2,602 books are unreachable.** `skip` exists on the endpoint but nothing in the UI drives it.
- A ~200 KB payload plus 100 card renders is a poor first-paint budget for the app's primary landing screen.
- Ordering only recently became correct (#491 added `sort(updated_at, -1)`; before that the page showed the *oldest* 100 forever). With ordering fixed, the missing pagination is now the binding limit.

## Notes for whoever picks this up

- `owner_updated_idx (owner_id, updated_at DESC)` already exists and backs the current sort — `explain()` shows `LIMIT → FETCH → IXSCAN`. Keep any paging change on that index; adding an `_id` tiebreaker measured as a blocking in-memory `SORT` stage.
- Cursor-based paging (`updated_at` + `_id` as a cursor) avoids the skip/limit drift problem on a list that changes while you page through it, but note the index caveat above.
- The list endpoint already projects out per-chapter `content`, so the payload is metadata only — the size is card count, not fat rows.

## Priority

P3.22 — degraded UX and an unreachable tail, not a correctness or security problem, and it only bites accounts with large libraries. The staging E2E account is the extreme case; real users are unlikely to be near 100 yet.

frankbria:
Closes #493.

## The problem

`GET /api/v1/books/` caps at 100 rows and the dashboard rendered every one of them with no pagination control. Past 100 books the tail was simply unreachable — `skip` existed on the endpoint but nothing in the UI drove it — and first paint meant ~195 KB and 100 card renders on the app's primary landing screen.

## The change

Offset paging, 24 books per page.

**No change to the Mongo query.** #493's own notes measured an `_id` tiebreaker as a blocking in-memory `SORT` stage, which is what cursor paging would need. The existing `.sort("updated_at", -1).skip().limit()` shape keeps riding `owner_updated_idx` (`LIMIT → FETCH → IXSCAN`).

**No total count, and no response envelope.** The client over-fetches one row (`limit = PAGE_SIZE + 1`) and derives `hasMore` from whether it came back. That avoids a second `count_documents` per page and avoids breaking `List[BookResponse]` — four existing API tests assert the bare-array shape.

Frontend:
- `bookClient.getUserBooks(opts?)` takes optional `{ skip, limit }`, compared against `undefined` rather than truthiness so `skip=0` still serializes. Omitting both leaves the URL byte-identical for existing callers.
- A labelled `<nav>` with Previous/Next and a "Page N" live region, rendered only when there is more than one page — invisible for normal libraries.
- The full-page loading skeleton is gated on "nothing has ever loaded", so the pager stays mounted while a later page is in flight.
- Delete keeps its optimistic splice on a single-page library, and additionally refetches once the list is paged (the row shifting up from the next page was never fetched). Emptying a later page steps back rather than stranding the user on a blank page showing the onboarding empty state.

Backend (two small changes, both consequences of the UI now driving these inputs):
- `skip` capped at 100k, matching the documented rationale already on the admin users route — Mongo walks every skipped document.
- Rate limit raised 20/min → 60/min. A dashboard visit used to cost exactly one GET; a 21-page library now costs 21 just to browse forward.

## Review

Cross-family review by GLM (opencode) on the branch diff, pre-PR. Two Major findings, both verified against the code and fixed in `706ea77`:

1. **Concurrent fetches were last-write-wins.** Two deletes in quick succession are enough — the per-card delete buttons are not gated while loading, so the first delete's refetch can land after the second's and restore a pre-delete snapshot. Fixed with a monotonic request token.
2. **Per-page GETs collided with the 20/60s budget**, and the 429 surfaced as the full-screen error branch replacing the grid *and* the pager. Fixed by raising the budget and by rolling back to the last good page with a toast instead of evicting the view.

Two Suggestions also fixed: the post-delete branch read stale closure state across the DELETE round-trip, and the pager used real `disabled`, which dropped keyboard focus to `<body>` on every page change (now `aria-disabled` + an onClick guard).

## Tests

23 new dashboard tests plus 3 client tests and a backend boundary test, all mutation-checked — 19 mutations across four rounds, every one caught except a single defensive line documented at the source. Demo verification is in a PR comment: all four acceptance criteria with outcome evidence from the real stack.

One note on process: the first mutation round was invalid. A stale `.git/index.lock` made every `git checkout` revert a silent no-op, so mutations accumulated and four "caught" results were contaminated. Re-run clean against committed work.

## Known limitations

- **Offset paging drifts.** A book updated while you are on page 3 can shift rows across a page boundary. Cursor paging is the fix and is explicitly deferred for the index reason above.
- **No "Page N of M".** Deliberate — it needs a total count.
- **A failed step-back leaves a false empty state.** If the refetch after emptying a later page fails (a 429, say), the screen lands on page 0 with a spliced-empty list and renders the onboarding empty state for a library that is not empty. The toast fires, but there is no in-place retry until refocus or reload.
- **Same-timestamp ties are unordered.** Pre-existing (the sort is `updated_at` alone, deliberately — an `_id` tiebreaker measured as a blocking in-memory `SORT`), but pagination makes it observable: a row can momentarily appear on two pages.
- **The pager is not covered by an axe test in CI.** It was scanned with real `axe-core` in a browser during demo verification (0 violations, 8 rules passed), but `ComponentAccessibilityAudit.test.tsx` has a Navigation Components block that would be the natural permanent home.

https://claude.ai/code/session_01KrEC9H3RKkpXzWaddjKMJf


frankbria:
## Third-party review — GLM (opencode), three passes

Cross-family independent review, on the record. Reviewer: GLM via opencode (`zai/glm-5.2`), read-only (no `--auto`). Three passes were needed because the first two each found real defects; the third is a convergence check against `0e52c31`.

| Pass | Diff reviewed | Findings | Outcome |
|---|---|---|---|
| 1 (pre-PR) | branch @ `e9cb124` | 2 Major, 3 Suggestion | all fixed in `706ea77` |
| 2 (post-PR) | PR @ `ba84966` | 3 (2 Medium, 1 Low) | all fixed in `f6ca183` |
| 3 (convergence) | PR @ `0e52c31` | none blocking | clear to merge |

### Pass 1 — 2 Major, 3 Suggestion (all fixed)

1. **Major, concurrent fetches were last-write-wins.** Paging turned one-fetch-per-visit into many; two deletes in quick succession suffice, since the per-card delete buttons are not gated while loading. Fixed with a monotonic request token.
2. **Major, per-page GETs collided with the 20/60s rate limit**, and the 429 surfaced as the full-screen error branch replacing the grid *and* the pager. Fixed by raising the budget to 60/min and rolling back to the last good page with a toast.
3. Suggestion, post-delete branch read stale closure state across the DELETE round-trip. Fixed with a ref.
4. Suggestion (a11y), real `disabled` on the pager dropped keyboard focus to `<body>` on every page change. Fixed with `aria-disabled` + an onClick guard.
5. Suggestion, missing tests for the new failure/concurrency paths. Added.

### Pass 2 — 3 findings (all fixed)

1. **Medium**, `pageStateRef` was synced only from a passive effect, which flushes on a macrotask. Two DELETE continuations in the same microtask batch both read the pre-splice snapshot, so the step-back never fired. The reviewer also correctly identified that the existing test could not catch this, because its `act()` flush forced an ordering production does not guarantee. Ref is now written synchronously at both mutation sites.
2. **Medium**, routing 404 through the new `hasLoadedOnce` branch turned "this user has no books" into a recurring error, toasted on every window refocus and unretryable. 404 now coerces to the empty state on every load and clamps to page 0.
3. **Low**, the rollback could resurrect a page a delete had just evacuated. Clamped with `Math.min`.

### Pass 3 — convergence

> **No blocking defects — clear to merge.**

The reviewer independently re-ran the gates (47/47 backend route tests, 23/23 pagination tests, 128/128 bookClient tests, `tsc --noEmit` clean) and verified rate-limit buckets are keyed per user per path, so the 20→60 raise is scoped to this route only.

Two non-blocking nits it raised, added to Known Limitations:

- If the step-back refetch after emptying a later page fails, the screen lands on page 0 with a spliced-empty list and renders the onboarding empty state for a library that is not empty. The toast fires, but there is no in-place retry until refocus or reload.
- The pre-existing `updated_at`-only sort makes ordering nondeterministic among same-timestamp ties, so a row can momentarily appear on two pages. Pre-existing, but pagination makes it observable.

### Mutation testing

19 mutations across four rounds, every one caught, except one deliberate exclusion documented at the line (a defensive synchronous ref write that the passive effect also covers, observable only in an ordering not practical to stage in a unit test).

One process note: the first mutation round was invalid and had to be redone. A stale `.git/index.lock` made every `git checkout` revert a silent no-op, so mutations accumulated across runs and four "caught" results were contaminated.

frankbria:
## Demo verification — every acceptance criterion, with outcome evidence

Run against the **real stack**: FastAPI on `127.0.0.1:8000` against a local MongoDB (`auto_author_demo493`), Next.js dev on `:3002`, driven through a real Chromium via `agent-browser`. Seeded a 120-book library — the shape #493 describes. The Atlas cluster in the local `.env` was never touched.

---

### AC1 — the tail past 100 books is reachable from the UI ✅

| | old behaviour (`limit=100`, no pager) | new behaviour |
|---|---|---|
| cards in one list | 100 | 24 |
| oldest book reachable | **Demo Book 021** | **Demo Book 001** |
| unreachable tail | **20 books** | none |

Paged forward in the browser to page 5 and read the DOM back:

```
{"page":"Previous Page 5 Next","count":24,"first":"Demo Book 024","last":"Demo Book 001",
 "btns":["Previous page=false","Next page=true"]}
```

Books 001–020 were previously unreachable — the old 100-row page stopped at Book 021. Next correctly reports `aria-disabled=true` on the final page.

### AC2 — first paint no longer renders 100 cards / ~195 KB ✅

Measured response sizes from the running API:

| request | rows | bytes |
|---|---|---|
| `?limit=100` (what the dashboard used to send) | 100 | **38,443** |
| `?skip=0&limit=25` (what it sends now) | 25 | **9,643** |

**75% smaller**, and 24 card renders instead of 100. (Absolute numbers are lower than the issue's ~195 KB because these seeded books have empty TOCs; the ratio is what this change controls.)

### AC3 — paging keeps the `owner_updated_idx` query shape, no in-memory SORT ✅

`explain("executionStats")` on the exact paged shape (`skip=96, limit=25`):

```
stages: LIMIT -> PROJECTION_DEFAULT -> FETCH -> SKIP -> IXSCAN
hasBlockingSort: false
docsExamined: 24   keysExamined: 120   returned: 24
```

No `SORT` stage — the index still serves the ordering, exactly as #493 required. `docsExamined: 24` confirms only the page is fetched. `keysExamined: 120` is the offset cost (Mongo walks skipped index keys), which is precisely why this PR also caps `skip` at 100k.

### AC4 — WCAG 2.1 AA on the new control ✅

Real `axe-core` in the browser, `wcag2a/wcag2aa/wcag21a/wcag21aa`:

```
scope: pagination nav only
violations: 0
passes: aria-allowed-attr, aria-conditional-attr, aria-prohibited-attr,
        aria-valid-attr-value, aria-valid-attr, button-name,
        color-contrast, nested-interactive
```

And the specific fix from review pass 1 — focus surviving a page change — verified in a real browser rather than jsdom:

```
{"focusBeforeClick":"Previous page","focusImmediatelyAfter":"Previous page"}
{"afterLoad":"Previous page","page":"Previous Page 4 Next"}
```

Focus stays on the activated button across the fetch. With real `disabled` it would have dropped to `<body>`.

---

### Bonus — the step-back branch, end to end

The highest-risk logic in the PR, demonstrated in the real app. Reseeded to 25 books so page 2 holds exactly one:

1. Page 2 shows a single card, `Demo Book 001`, Next `aria-disabled=true`.
2. Deleted it through the real type-to-confirm modal.
3. Result:

```
{"pager":"HIDDEN (single page now)","cards":24,
 "first":"Demo Book 025","last":"Demo Book 002","emptyState":false}
```

Stepped back to page 1, the pager correctly removed itself (24 books is one page), and `emptyState:false` — the user was **not** stranded on a blank page showing "you haven't created any books yet". That stranding is the failure this branch exists to prevent.

---

### One pre-existing defect found while running axe

The full-page scan returned exactly one violation, and it is **not** from this PR:

```
color-contrast (serious) — <a class="text-primary font-medium" href="/dashboard">Dashboard</a>
contrast 2.85:1 (#4f46e5 on #171717), needs 4.5:1
```

The header nav link fails WCAG 2.1 AA on the dark theme. Pre-existing, unrelated to pagination, filed separately rather than folded into this PR.

frankbria:
## Triaged the CI bot's two findings — both real, both fixed in `30df6a2`

Verified each against the code before acting, per the usual rule. Both were genuine, and the first is a regression I introduced myself in `f6ca183`.

**1. Failed *Previous* left `page` behind the content (minor).** The `Math.min` clamp I added to stop a failed step-back refetch from resurrecting an emptied page was directional, and direction is the wrong discriminator. On a failed back-navigation it is a silent no-op: `page` ends up one behind the books on screen, so the pager reads "Page 2" over page-3 content and can disable Previous while later-page books are displayed.

Replaced the clamp with the honest rule — roll back to whichever page the visible books actually belong to — and handled the step-back case where it happens, by moving `loadedPageRef` along with the step-back. Both cases are now correct and the clamp is gone entirely.

**2. The 404 branch did not clear `hasMore` (minor).** A 404 arriving on a page-0 fetch reset `projects` and `page` but left `hasMore` true, so the pager stayed mounted and enabled behind the onboarding empty state, and Next fired another doomed fetch. Now clears `hasMore` and the last-good marker too.

Two tests added, both mutation-checked — including one that fails if the directional clamp comes back:

```
[baseline]                        25 passed
[M20-directional-clamp-returns]   1 failed  ✓ caught
[M21-404-leaves-hasMore]          1 failed  ✓ caught
[M22-stepback-leaves-marker]      1 failed  ✓ caught
```

That brings this PR to 22 mutations across five rounds. Worth noting the bot earned its keep here: three manual GLM passes read the rollback logic and none caught the directional-clamp regression.

Issue #584, opened 2026-09-05
[P3.31] 30 react-hooks v7 violations warned rather than fixed during the ESLint 9 migration


frankbria:
`eslint-plugin-react-hooks` 7 (arrived with `eslint-config-next@16`) adds three rules that are **errors by default**. They flag 30 pre-existing call sites, so `frontend/eslint.config.mjs` downgrades all three to `warn` to keep that PR an ESLint upgrade rather than a React refactor.

| Rule | Sites |
|---|---|
| `react-hooks/set-state-in-effect` | 26 |
| `react-hooks/refs` | 2 |
| `react-hooks/immutability` | 2 |

Full list: `cd frontend && npx eslint . -f json`, filter for those three rule ids.

Known offenders include `src/app/auth/reset-password/page.tsx`, `src/app/dashboard/books/[bookId]/page.tsx`, `src/app/dashboard/books/[bookId]/summary/page.tsx`, `src/components/BookCreationWizard.tsx`, `src/hooks/useSwipeGesture.ts`, `src/components/chapters/questions/QuestionContainer.tsx`, `src/components/export/ExportOptionsModal.tsx`.

`set-state-in-effect` is the one worth reading closely: calling `setState` in an effect usually means a render-derived value that should be computed during render or folded into the state that drives it, and each instance is a wasted extra render pass.

## Condition to close

Fix the call sites and delete the three `"warn"` lines from the `ALL_FILES` block in `frontend/eslint.config.mjs` so the rules go back to their default `error`. Partial progress is fine — the rules only flip back once the count reaches zero.

Sits alongside the existing "Downgrade to warnings during ESLint migration" block in the same file, which is older debt of the same shape.

frankbria:
## Measured 2026-09-12 — the count has grown, and the three rules are not equally hard

Re-ran the sweep (`npx eslint . -f json`, filtered to the three rule ids) rather than trusting the issue body:

| rule | issue said | actual | change |
|---|---|---|---|
| `react-hooks/set-state-in-effect` | 26 | **26** | — |
| `react-hooks/refs` | 2 | **4** | +2 |
| `react-hooks/immutability` | 2 | **2** | — |
| | 30 | **32** | |

Across **20 files**, concentrated in `summary/page.tsx` (4), `dashboard/page.tsx` (3) and `ChapterEditor.tsx` (3).

## The useful finding: this splits into three jobs, not one

[...]

PR #615, opened 2026-09-08
fix(a11y): lift text-primary to indigo-400 on dark surfaces (#610)

### LINKED ISSUE #610: [P2.27] Header nav link fails WCAG 2.1 AA contrast on the dark theme (2.85:1)
Found while running `axe-core` in a real browser during demo verification for #493. It is the **only** violation on the dashboard, and it is pre-existing — unrelated to that PR.

## The finding

```
color-contrast (serious)
<a class="text-primary font-medium transition-colors" href="/dashboard">Dashboard</a>
Element has insufficient color contrast of 2.85
  foreground: #4f46e5   background: #171717   16px, normal weight
Expected contrast ratio of 4.5:1
```

`text-primary` (indigo `#4f46e5`) on the dark surface `#171717` reaches **2.85:1** against a required **4.5:1** for normal-weight body text.

## Why it is worth fixing

- `CLAUDE.md` states **WCAG 2.1 AA for user-facing UI** as a project standard, so this is a stated-standard violation, not a preference.
- It is the *active* nav link — the element indicating where the user currently is — on a persistent site header, so it is on every authenticated page.
- Same indigo is used elsewhere (`bg-indigo-600` buttons pass because they are white-on-indigo; this is indigo-on-near-black, the inverted case). Worth checking whether any other `text-primary`/`text-indigo-*` on a dark surface has the same problem.

## Reproducing

```bash
# with the app running
cp frontend/node_modules/axe-core/axe.min.js frontend/public/axe-demo.js
agent-browser open http://localhost:3002/dashboard
agent-browser eval "const s=document.createElement('script');s.src='/axe-demo.js';document.head.appendChild(s);new Promise(r=>s.onload=()=>r('ok'))"
agent-browser eval "axe.run(document,{runOnly:{type:'tag',values:['wcag2a','wcag2aa','wcag21a','wcag21aa']}}).then(r=>JSON.stringify(r.violations))"
rm frontend/public/axe-demo.js
```

## Fix direction

Lighten the token used for `text-primary` on dark surfaces until it clears 4.5:1 — indigo-400 (`#818cf8`) on `#171717` is ~7:1 and is already the shade used for the `Book02Icon` heading accent right below it, so the page would also get more internally consistent. Verify with axe rather than by eye.

## Acceptance criteria

- [ ] `axe-core` reports 0 `color-contrast` violations on `/dashboard` at wcag21aa.
- [ ] Any other `text-primary`-on-dark usage audited in the same pass.
- [ ] A regression test added to `frontend/src/__tests__/accessibility/ComponentAccessibilityAudit.test.tsx` (its `4.1 Navigation Components` block is currently an `it.skip` placeholder).

frankbria:
Closes #610.

## The bug

`tailwind.config.js` declares `primary` a **theme-fixed** brand colour:

```js
primary: { DEFAULT: "rgb(79, 70, 229)", foreground: "white" },   // indigo-600
```

The `--primary` CSS variable in `globals.css` does *not* drive it — that variable only
appears in the `@theme inline` block, which Tailwind v3 ignores (the config's own comment
already says so). So `text-primary` is indigo-600 in **both** themes, while the sticky
header sits on `bg-card` = `oklch(0.205 0 0)` = `#171717` in dark. That is **2.85:1**,
under the 4.5:1 WCAG 2.1 AA floor for normal-weight text — on the *active* nav link, i.e.
on every authenticated page.

## The fix

A `.dark`-scoped override of the **text role only**, in `globals.css`:

```css
.dark .text-primary,
.dark .hover\:text-primary:hover { color: rgb(129, 140, 248); }   /* indigo-400 */
```

**Why not re-theme the token.** `primary` also drives `bg-primary`, whose pair is
`text-primary-foreground: white`. White on indigo-400 is **2.98:1** — re-theming would
trade one AA failure for another (the "Create New Book" button, every default `Button`,
the mobile active nav pill). Light mode is untouched at 6.28:1.

**Why the hover selector is named separately.** Tailwind emits its own
`.hover\:text-primary:hover` at equal specificity and *later* in the stylesheet, so
`.dark .text-primary` alone loses to it — the inactive header links would keep failing
on hover. Verified live: hovering Settings resolves to `rgb(129, 140, 248)`.

Base and `hover:` are the only variants of this colour in the codebase — swept for
`text-primary/<opacity>` and for `focus:`/`group-`/`data-`/`aria-` variants; the
`focus:`/`selection:` hits are all `text-primary-foreground`, which is unaffected
(`.text-primary` is a distinct class token). `codex review` swept for the same
independently and agreed.

## Demo — real axe-core in Chrome, not jsdom

Local stack, `defaultTheme="dark"`, 30 seeded books so the pagination nav actually mounts.

| Acceptance criterion | Evidence |
|---|---|
| 0 `color-contrast` violations on `/dashboard` at wcag21aa | **0 violations**, 22 rules passed, **153** elements passed `color-contrast` |
| The reported element specifically | `<a class="text-primary font-medium…" href="/dashboard">` → **6.01:1**, fg `#818cf8`, bg `#171717` (was 2.85:1) |
| Hover state | Hovering `/dashboard/settings` → computed `rgb(129, 140, 248)` |
| Other `text-primary`-on-dark audited | `/dashboard/help` **0 violations** (6.01:1 on card), `/terms` **0** (6.63:1 on background), `/privacy` **0** (6.63:1) |
| Light theme unchanged | **0 violations**; nav link still `#4f46e5` at **6.28:1** on white |
| Pagination nav (#493 follow-up, per the issue comment) | scoped scan: **0 violations**, 8 rules passed |

Measured ratios match the values computed in the test to two decimals (6.01 / 6.63 / 6.28),
and the pre-fix 2.85 in the issue reproduces exactly — the oklch→sRGB conversion in the
guard is calibrated against a real browser.

## Guards

**`theme/dark-primary-text-contrast.test.ts`** (new) recomputes the ratio from the values
actually in `globals.css`, so it fails both if the override colour is weakened *and* if the
dark surface tokens are re-tuned out from under it. jsdom carries no stylesheet and no
layout, so a `jest-axe` `color-contrast` scan there is vacuous — this is the real guard,
and the test says so.

**`ComponentAccessibilityAudit.test.tsx` §4.1** — the `it.skip` placeholder is filled with a
structural axe scan of the header nav and the pagination nav in their real composition
(`<DashboardLayout><Dashboard /></DashboardLayout>`). Per the issue comment, the #493
pagination nav had only ever been scanned by hand once; it is now in CI.

Mutation-checked (each reverted and re-asserted clean afterwards):

| Mutation | Caught by |
|---|---|
| indigo-400 → indigo-500 (4.01:1 on card) | 3 contrast assertions fail |
| `--card` dark re-tuned to `oklch(0.55 0 0)` | 1 contrast assertion fails |
| drop `aria-label="Main navigation"` | §4.1 scan fails |
| pagination `aria-label` → `"Books"` | §4.1 scan fails |
| remove the override / the hover selector | RED before the fix |

## Gates

- `npm test` — 137 suites, **2332 passed**, 0 failures
- `npm run typecheck` — clean
- `npm run lint` — **0 errors** (446 warnings, all pre-existing, tracked in #584)
- pre-commit — secrets, lint, typecheck, unit tests, **coverage ≥85%** all passed
- Third-party review — `codex review --base main`: **no findings**. (opencode/GLM produced
  zero output after 10 min — the known silent hang; codex is the documented fallback.)

## Known limitations

- The override targets Tailwind's generated class names. That is deliberate — it fixes every
  current *and future* `text-primary` on dark in one place, where per-call-site
  `dark:text-indigo-400` (17 occurrences across 9 files) would let the next one reintroduce
  the bug. The trade is a specificity dependency, which the hover test pins explicitly.
- Not a general contrast audit: only the `primary` text role is covered. `ChapterTab` still
  hardcodes `text-gray-700` on `bg-muted`, which is its own dark-theme problem outside this
  issue's scope — filed separately.
- The `@theme inline` block in `globals.css` is dead under v3 and makes Next log
  `Unknown at rule: @theme` on every compile. Pre-existing and unrelated; noted on #513
  (Tailwind v4 migration), which is what would actually resolve it.


frankbria:
## Third-party review (on the record)

Diff unchanged since the pre-PR pass (`c45faff`), so this is that run's output rather than a re-run.

**Primary — opencode / GLM:** produced **zero output** after 10 minutes and was killed (exit 143). This is the documented silent-hang failure mode, not a clean review; falling back per the repo's review policy.

**Fallback — `codex review --base main`:** **no findings.**

> The changes add a scoped dark-mode `text-primary` contrast override and targeted regression tests. I did not identify any discrete introduced bug that would break existing behavior or tests.

It independently swept for other Tailwind variants of this colour — `text-primary/<opacity>`, and `focus:`/`group-`/`aria-`/`data-`/`active:`/`visited:` variants — and found none beyond the base and `hover:` selectors the override already names. That matches the sweep in the PR description.

Internal review, mutation checks, and the real-browser axe demo are in the PR body.

Issue #589, opened 2026-09-05
[P2.26] nwsapi 2.2.26+ makes ChapterTab keyboard tests time out; pinned to 2.2.25 via overrides


frankbria:
`frontend/package.json` pins `nwsapi` to `2.2.25` in its `overrides` block. That pin is load-bearing — without it the frontend test suite does not pass.

## What breaks

`nwsapi` is jsdom's CSS selector engine, reached transitively as `jest-environment-jsdom → jsdom → nwsapi` on a `^2.2.x` range, so **nothing in this repo pinned it** and any lockfile regeneration floated it forward.

At `2.2.26` and `2.2.27`, `src/components/chapters/__tests__/ChapterTab.keyboard.test.tsx` goes from **23 passed in 1.1s** to **16 failed, 7 passed in ~220s**, every failure being `Exceeded timeout of 5000 ms`. A single `userEvent` keypress against `ChapterTab` takes **12–15 seconds**.

Profiling `Element.prototype.matches` during one keypress:

```
matches(':fullscreen')   n = 29,999,829
matches(':modal')        n = 30,581
```

Each `:modal` evaluation triggers roughly a thousand nested `:fullscreen` evaluations. `2.2.26` rewrote the compiled-resolver cache from a `Map` to a hand-rolled ES5 LRU (and renamed `clr` → `clear`); the selector cache itself still behaves — a micro-benchmark of repeated `matches()` shows no regression — so the fault is in the `:modal` / `:fullscreen` resolution path, not the cache lookup.

## Isolated, not inferred

Bisected on `main`'s **exact** dependency tree with only `nwsapi` changed:

| nwsapi | Result |
|---|---|
| 2.2.24 (main today) | 23 passed, 1.1s |
| 2.2.25 | 23 passed, 1.0s |
| **2.2.26** | **16 failed, ~219s** |
| **2.2.27** (latest) | **16 failed, ~219s** |

This first surfaced as a `Frontend Tests` failure on Dependabot PR #588, which made it look like one of that PR's 10 updates. It is not — `@testing-library/user-event`, `@testing-library/react`, `@testing-library/jest-dom` 7, `jest`, `jest-environment-jsdom` and `jest-axe` were each cleared individually. #588's only role was regenerating the lockfile, which floated `nwsapi` forward. Verified the other way too: #588's exact dependency set **plus** this pin gives 133 suites / 2288 passing in 39s.

## Condition to close

[...]

PR #617, opened 2026-09-08
fix(security): patch js-yaml and sharp to unblock the Security Audit gate (#616)

### LINKED ISSUE #616: [P1.1] Security Audit is red on main: new js-yaml and sharp advisories block every merge
`main` is currently red on the **required** `Security Audit` check, so **no PR in the repo can merge**. Two high-severity advisories were published against packages already in the tree; nothing in the repo changed to cause it.

Reproduced against pristine `origin/main` (`5a314c1`) with zero local changes:

```
New advisories not in the baseline (2):
  [npm] GHSA-2883-xcg3-v3hh  js-yaml  severity=high  fix=yes
  [npm] GHSA-rgj7-g3m4-5g8c  sharp    severity=high  fix=yes

FAIL — fix these, or add each ID to the baseline with a reason if it is accepted debt.
```

Surfaced while running #610's PR through CI (#615), which is blocked behind it.

## The advisories

| ID | Package | Vulnerable | Patched | Note |
|---|---|---|---|---|
| GHSA-2883-xcg3-v3hh | `js-yaml` | `>=4.0.0 <4.3.2` **and** `>=3.0.0 <3.15.2` | 4.3.2 / 3.15.2 | `maxTotalMergeKeys` does not limit CPU use for empty merge sources |
| GHSA-rgj7-g3m4-5g8c | `sharp` | `<0.35.4` | 0.35.4 | CVSS 8.9 — bundled libheif (GHSA-g89c-p67h-r497, GHSA-2jg2-4ch7-h545) |

Both are transitive and both have fixes, so neither belongs in the baseline — the baseline is for advisories with no upgrade path.

- `js-yaml` ← `eslint → @eslint/eslintrc` (v4 line) and `ts-jest → @jest/transform → babel-plugin-istanbul → @istanbuljs/load-nyc-config` (v3 line)
- `sharp` ← `next@16.3.3`

## The trap

`frontend/package.json` **already** overrides `js-yaml`, pinned at the now-vulnerable versions from an earlier advisory:

```json
"js-yaml@^3": "3.15.1",
"js-yaml@^4": "4.3.1",
```

So this is not "float the transitive forward" — the existing pins have to be moved, or the fix silently does nothing. The advisory also spans **both** major lines; bumping only `@^4` leaves the gate red on the v3 copy under `ts-jest`.

## Fix

Bump the two `js-yaml` pins to `3.15.2` / `4.3.2` and add `"sharp": "^0.35.4"` to `overrides`.

## Acceptance criteria

- [ ] `scripts/audit_gate.py` reports `PASS — no new advisories`
- [ ] `security-baseline.json` unchanged (these are fixed, not accepted)
- [ ] The `nwsapi` `2.2.25` pin survives the lockfile regeneration — floating it is #589, worth an explicit check rather than an assumption
- [ ] Frontend suite, lint, and typecheck green

frankbria:
Closes #616. **Unblocks every PR in the repo** — `Security Audit` is a required check and `main` is currently red on it.

## Not caused by anything in the repo

Two high-severity advisories were published against packages already in the tree. Reproduced against pristine `origin/main` (`5a314c1`) with zero local changes:

```
New advisories not in the baseline (2):
  [npm] GHSA-2883-xcg3-v3hh  js-yaml  severity=high  fix=yes
  [npm] GHSA-rgj7-g3m4-5g8c  sharp    severity=high  fix=yes
```

Found while running #610 through CI (#615), which is blocked behind this.

## The trap this walks into

`overrides` **already** pinned `js-yaml` — at exactly the two versions the new advisory names:

```json
"js-yaml@^3": "3.15.1",     // vulnerable range is >=3.0.0 <3.15.2
"js-yaml@^4": "4.3.1",      // vulnerable range is >=4.0.0 <4.3.2
```

So this is not "let the transitive float forward" — the pins have to move or the fix silently does nothing. And the advisory spans **both** major lines: bumping only `@^4` still leaves the gate red on the v3 copy under `ts-jest → babel-plugin-istanbul → @istanbuljs/load-nyc-config`. (The GitHub advisory API returns only the first vulnerability entry, which shows the v4 range alone — the v3 range is visible in `npm audit`'s `via`.)

`sharp` (CVSS 8.9, bundled libheif) comes in under `next@16.3.3` and needs its own override.

Both have upstream fixes, so neither belongs in `security-baseline.json` — that file is for advisories with **no** upgrade path. It is unchanged.

## Blast radius

The lockfile delta is exactly what it should be — 30 version moves, 2 added, 2 removed:

- `sharp` 0.35.3 → 0.35.4, plus its 13 platform binaries and 9 `libvips` packages
- `js-yaml` 4.3.1 → 4.3.2 and 3.15.1 → 3.15.2
- an `@emnapi` hoisting shuffle (`@emnapi/core`, `@emnapi/wasi-threads` moved under `@unrs/resolver-binding-wasm32-wasi`)

Nothing else moved. In particular **the `nwsapi` 2.2.25 pin survived the regeneration** — checked explicitly rather than assumed, because a floated `nwsapi` is #589 and its symptom is a 200× test slowdown, not an obvious failure:

```
└── nwsapi@2.2.25 overridden
ChapterTab.keyboard.test.tsx — 23 passed, 0.958s     (floated: 16 failed, ~220s)
```

## Verification

| Check | Result |
|---|---|
| `scripts/audit_gate.py` | **PASS — no new advisories** (4 known, already in the baseline) |
| `security-baseline.json` | unchanged |
| `npm test` | 136 suites, **2326 passed**, 0 failures |
| `npm run typecheck` | clean |
| `npm run lint` | **0 errors** (446 warnings, all pre-existing, tracked in #584) |
| pre-commit | secrets, lint, typecheck, tests, coverage ≥85% all passed |

## Known limitations

- Frontend only. `pip-audit` reported nothing new; the backend is untouched.
- `sharp` is pinned by override rather than by a `next` upgrade. When `next` ships a release that depends on `sharp >= 0.35.4` itself, the override becomes redundant and should be dropped — same lifecycle as the other pins here.
- This does not shorten the 91-entry baseline; it only keeps new advisories from accumulating into it.


frankbria:
## Third-party review (on the record)

**Primary — opencode / GLM:** skipped. It produced zero output for ~10 min on this session's earlier review and had to be killed (exit 143) — the documented silent-hang mode. Falling back per repo policy.

**Fallback — `codex review --base main`:** one **[P1] blocking** finding, raised and **verified as not reproducing**.

> **[P1] Restore missing @emnapi peer entries in lockfile** — `frontend/package-lock.json`
> With this lockfile, a clean frontend install fails because `@napi-rs/wasm-runtime` still declares `@emnapi/core` and `@emnapi/runtime` as peers, but the top-level lockfile entries that satisfied them were removed. Running `npm ci --dry-run --ignore-scripts --no-audit --no-fund --offline` now errors with `Missing: @emnapi/runtime@ from lock file`.

### Verification — does not reproduce

The failure is an artefact of the reviewer's own `--offline` flag, not of the lockfile. The `@emnapi/*` entries are optional peers of `@napi-rs/wasm-runtime`, itself an `optional: true`, `dev: true` wasm32-wasi binding; `--offline` cannot resolve an optional peer that is absent from the local cache, and reports it as a lockfile gap.

Same command without `--offline`:

```
$ npm ci --dry-run --ignore-scripts --no-audit --no-fund
up to date in 657ms          # exit 0
```

And a genuine clean-room install — empty directory, only `package.json` + `package-lock.json`, no `node_modules`, which is what CI does:

```
$ npm ci --ignore-scripts --no-audit --no-fund
added 1006 packages in 25s   # exit 0
```

No change made. The `Frontend Tests` check on this PR runs `npm ci` from a clean checkout and is the independent confirmation.

No other findings.

frankbria:
## Correction — the [P1] finding above was real, and my rebuttal was wrong

`codex` was right. I dismissed it on a verification that could not detect the problem. CI then failed exactly as it predicted:

```
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json are in sync.
npm error Missing: @emnapi/runtime@1.11.3 from lock file
npm error Missing: @emnapi/core@1.11.3 from lock file
```

### Why my check missed it

Two flaws, either one sufficient:

1. **Wrong npm.** This machine has npm **11.6.2**; CI runs Node 24.20.0 with npm **11.19.0**. The two resolve `@napi-rs/wasm-runtime`'s optional `@emnapi/*` peers differently — 11.6.2 drops the top-level entries and is happy, 11.19.0 requires them. The lockfile was generated by the wrong resolver.
2. **`npm ci --dry-run` against a populated `node_modules` passes regardless.** It reported `up to date in 657ms`, which I read as a pass. It is not a lockfile-consistency check.

My clean-room `npm ci` was closer but still used the local npm, so it also passed. The finding was reproducible the whole time; I just never ran the command that could see it.

### Fix

Lockfile regenerated with `npx npm@11.19.0 install`. `@emnapi/core` and `@emnapi/runtime` are back at the top level, and the delta vs `main` is now **31 version moves, 2 additions, 0 removals** — the previous one removed 2 entries, which was the bug.

### Verification that actually discriminates

| Lockfile | `npm ci` under npm 11.19.0, empty dir |
|---|---|
| previous commit (`634f1b1`) | **FAIL** — `Missing: @emnapi/runtime@1.11.3` |
| this commit (`d462202`) | **PASS** — 1004 packages, exit 0 |

Re-verified after the regeneration: audit gate `PASS — no new advisories`; `security-baseline.json` untouched; `nwsapi` still pinned at 2.2.25 with `ChapterTab.keyboard` 23/23 in 0.85s (the #589 tripwire); 136 suites / 2326 passed; lint 0 errors; typecheck clean.

frankbria:
GLM review: **no defects found**. Its one observation — that the regeneration dropped the `libc` key from 16 platform packages, which my PR body's "nothing else moved" glossed over — is correct, and I verified it rather than taking it on faith:

```
entries that lost `libc`: 16
  6 × @parcel/watcher-linux-*   (glibc/musl)
 10 × @unrs/resolver-binding-linux-*  (gnu/musl)

entries where os/cpu gate changed or dev/optional lost: NONE
entries that GAINED a libc restriction: NONE
```

Every one is `dev: true, optional: true` with its `os`/`cpu` gate byte-identical, and the glibc/musl split is already encoded in each package name. Dropping a `libc` filter can only widen where an optional package is eligible, never exclude one a host needs — and no entry gained a filter, which is the direction that *could* have broken a musl builder. npm-version normalization between 11.6.2 and 11.19.0, no failure scenario.

Merging: all four required checks green (Frontend Tests, Backend Tests, E2E Tests (Playwright), Security Audit), and `Security Audit` passing here is the whole point of the PR.

Issue #331, opened 2026-07-23
[P0.2] Core authoring pages hardcode near-white text — invisible in the shipped light theme


frankbria:
**Phase:** P0.2 — launch blocker (core flow unusable in a shipped state). No dependencies.

**Problem:** Book detail, summary, and TOC-wizard pages hardcode `text-gray-100` (and `bg-gray-800`/`border-gray-700`) headings/body directly on the theme-aware `bg-background`. The app ships a working light/dark toggle (Settings #64), so light mode is a reachable state and these elements drop to ~1.1:1 contrast (invisible). The #206 fix already migrated BookCard/EmptyBookState to tokens; these pages were missed.

**Evidence:**
- `frontend/src/app/dashboard/books/[bookId]/page.tsx:428` (book title `text-gray-100`), `:453,469,498` (stepper)
- `frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:157`
- `frontend/src/components/toc/TocGenerationWizard.tsx:313`

**Impact:** A user who selects the Light theme cannot see the book title, summary heading, or TOC heading — the core authoring flow is unusable. WCAG 1.4.3 failure.

**Acceptance criteria:**
- [ ] Replace hardcoded `text-gray-*`/`bg-gray-*`/`border-gray-*` on these pages with the token vocabulary (`text-foreground`, `text-muted-foreground`, `bg-card`, `border-border`).
- [ ] Add a light-theme render assertion (or contrast check) so the regression can't recur silently.

PR #619, opened 2026-09-09
fix(a11y): theme-token the chapter tab strip — inactive title was 1.46:1 on dark (#618)

### LINKED ISSUE #618: [P1.2] Inactive chapter tab title is 1.47:1 on the dark theme — hardcoded text-gray-700 outside the #331 guard's path list
Found while auditing `text-primary` usages for #610. Out of that issue's scope (it is a different token), but strictly worse than the failure #610 was opened for.

## The finding

`frontend/src/components/chapters/ChapterTab.tsx:89` styles the **inactive** chapter tab title with a hardcoded, theme-independent gray:

```tsx
isActive ? "text-primary" : "text-gray-700 hover:text-primary"
```

The tab's own surface a few lines up is theme-aware (`bg-muted` inactive, `bg-background` active), so in the dark theme — which is `defaultTheme` — gray-700 lands on a near-black surface:

| foreground | surface | ratio | required |
|---|---|---|---|
| `text-gray-700` `#374151` | `bg-muted` dark `#262626` | **1.47:1** | 4.5:1 |
| `text-gray-700` `#374151` | `bg-background` dark `#0a0a0a` | **1.92:1** | 4.5:1 |

1.47:1 is effectively invisible — it is the same failure mode as #331 (`~1.1:1`, "the core authoring flow was unusable"), and this is the chapter tab strip in that same authoring flow. For comparison, the nav link in #610 was 2.85:1.

Line 113 of the same file carries a whole hardcoded light-mode cluster on the delete button — `border-gray-300 bg-gray-100 text-gray-700 focus:bg-gray-200 focus:text-gray-900` — which should be checked in the same pass.

## Why the existing guard missed it

`src/__tests__/theme/core-authoring-pages-tokens.test.ts` (#331) bans gray literals, but only across three explicitly listed paths:

```
src/app/dashboard/books/[bookId]/page.tsx
src/app/dashboard/books/[bookId]/summary/page.tsx
src/components/toc/TocGenerationWizard.tsx
```

`ChapterTab.tsx` is not in that list, so the guard is green and the bug ships. The gate exists and does not cover the file.

## Caveat on the numbers

The ratios above are computed from the token values in `globals.css` (`--muted`/`--background` under `.dark`), not measured in a browser. The same conversion reproduced axe-core's real-browser numbers exactly at four points during #610 (2.85, 6.01, 6.63, 6.28), so the arithmetic is trustworthy; what is assumed is that `bg-muted`/`bg-background` are the actual painted surfaces behind the label. Worth one axe run to confirm before fixing.

## Suggested fix

Use the theme tokens the rest of the component already uses — `text-muted-foreground` for the inactive label (that is what the header nav uses for exactly this state, and it measures 6.93:1 on dark), `text-foreground` on hover — rather than picking a different gray literal.

## Acceptance criteria

- [ ] Confirm with real `axe-core` in a browser, dark theme, on a book page with chapter tabs
- [ ] No `*-gray-\d+` literals left in `ChapterTab.tsx` (line 89 and the line 113 button cluster)
- [ ] `axe-core` reports 0 `color-contrast` violations on the chapter-tab strip in **both** themes
- [ ] Add `src/components/chapters/ChapterTab.tsx` to the `PAGES` list in `core-authoring-pages-tokens.test.ts`, and audit what else that list should be covering — a hand-maintained path list is the reason this shipped

frankbria:
Closes #618.

## The bug

`ChapterTab.tsx` styled the **inactive** chapter tab title with a hardcoded `text-gray-700`. The tab's own surface is theme-aware (`bg-muted` inactive, `bg-background` active), so on the dark theme — which is `defaultTheme` — the label all but disappeared.

Two more hardcoded clusters in the same file went with it: the status dot's `border-gray-400 bg-white`, and the close button's `border-gray-300 bg-gray-100 text-gray-700 focus:bg-gray-200 focus:text-gray-900`.

## Measured, in a real browser

axe-core 4.13.0 injected into Chrome against a seeded book, scoped to the chapter-tab strip. The issue's predicted `#262626` surface was exactly right.

| | before | after |
|---|---|---|
| inactive title, dark (`#262626`) | **1.46:1** ❌ | `#fafafa` → **14.49:1** ✅ |
| inactive title, light (`#f5f5f5`) | 4.74:1 ❌ | `#0a0a0a` → **18.15:1** ✅ |
| active title, dark | 6.63:1 ✅ | 6.63:1 ✅ |
| active title, light | 6.28:1 ✅ | 6.28:1 ✅ |

`color-contrast` violations on the strip: **0 dark, 0 light, 0 incomplete.** The "before" row is not arithmetic — I set the old `#374151` back on the live nodes and re-ran axe, which reported 3 violations at 1.46:1, then restored and got 0.

## Two things the browser caught that the issue's suggested fix would not have

1. **`text-muted-foreground` fails AA on the light theme.** It is what #618 recommended and it is fine on dark (5.85:1), but `--muted-foreground` `#737373` on `--muted` `#f5f5f5` is **4.34:1** — under the floor. The AC asks for 0 violations in *both* themes, so the inactive title is `text-foreground`. `hover:text-primary` returns as the hover affordance, safe on dark since #610.
2. **The draft status dot was invisible.** `statusConfig[DRAFT].color` was `bg-muted` — the same token as the inactive tab's own surface. The `bg-white` in the base class list never masked it: `twMerge` drops it for `config.color` in every branch, verified with a scratch assertion. Now `bg-muted-foreground`.

Found by `codex review`, which flagged the dot as a regression this PR introduced. It isn't — `bg-white` was already dead before the change — but the underlying bug it pointed at is real and pre-existing, so it's fixed here.

## Why the gate was green while the bug shipped

`core-authoring-pages-tokens.test.ts` (#331) bans gray literals across a **hand-maintained** three-path `PAGES` list, and `ChapterTab.tsx` was not on it. Adding it fixes this file; mutation-checked (put `text-gray-700` back → the guard fails; revert → passes).

## The `PAGES` audit (AC 4)

The list is the root cause, so I measured it rather than guessing. Walking imports transitively from the four core authoring routes plus the TOC wizard and the dashboard layout: **80 files reachable, 14 still carrying 272 gray literals** — including every child step of `TocGenerationWizard`, which is itself on the list.

| literals | file |
|---|---|
| 62 | `app/dashboard/books/[bookId]/export/page.tsx` |
| 44 | `components/toc/TocReview.tsx` |
| 34 | `components/toc/ClarifyingQuestions.tsx` |
| 25 | `app/dashboard/books/[bookId]/edit-toc/page.tsx` |
| 24 | `components/toc/NotReadyMessage.tsx` |
| 20 | `components/BookMetadataForm.tsx` |
| 16 | `components/toc/TocGenerating.tsx` |
| 14 | `components/navigation/ChapterBreadcrumb.tsx` |
| 13 | `components/toc/ErrorDisplay.tsx` |
| 8 | `components/toc/ReadinessChecker.tsx` |
| 6 | `components/errors/ErrorNotification.tsx` |
| 3 | `components/ui/ChapterStatusIndicator.tsx` |
| 2 | `components/ui/avatar.tsx` |
| 1 | `components/export/ExportOptionsModal.tsx` |

That is a separate, much larger piece of work than a P1 contrast fix, so it is filed rather than bundled — see #620.

## Known limitations

- The audit is a **static** import walk; it would miss a component reached only through a dynamic import.
- The 4.34:1 `--muted-foreground` on `--muted` pairing in light mode is a **token-level** problem, not a ChapterTab one. This PR routes around it; wherever else that pair carries body text, it is still failing. Captured in the follow-up.
- The demo seeded chapters through `table_of_contents`, which does not carry per-chapter status, so all four tabs rendered as DRAFT. That exercises the fixed dot but leaves the other three status colours (`bg-blue-500` / `green` / `purple`, all theme-fixed brand accents) browser-unverified.
- Reviewed by `codex review`; `opencode` was tried first and failed with an upstream `UnknownError` (`err_bf136cbe`).


frankbria:
## Third-party review (post-PR)

**Reviewer:** `codex review --base main`, run against this PR's final diff.
`opencode` (GLM) was the primary and was tried first; it failed upstream with `UnknownError` / `err_bf136cbe` and produced no output, so this is the documented fallback.

**Verdict — no findings:**

> The changes are narrowly scoped to replacing hardcoded gray utilities with theme tokens and making the draft status dot visible against the inactive tab surface. I did not identify any introduced correctness, accessibility, or test-breaking issues in the diff.

### The pre-PR pass (same reviewer, earlier diff) did raise one

> **[P3] Preserve visible draft status indicator** — `ChapterTab.tsx:85`
> For inactive draft chapters, the parent tab already uses `bg-muted`, and `config.color` for `ChapterStatus.DRAFT` is also `bg-muted`, so removing the contrasting `bg-white` fill makes the 2x2 status dot blend into the tab surface.

Half right, and worth writing down which half.

**Not a regression.** `bg-white` never rendered. `cn` is clsx + `tailwind-merge`, and `config.color` is merged *after* the base list, so `twMerge` resolved `bg-white` away in all four status branches before this PR existed. Verified rather than reasoned:

```ts
expect(cn("... border border-gray-400 bg-white", "bg-muted")).toBe("... border border-gray-400 bg-muted");
expect(cn("... border border-gray-400 bg-white", "bg-blue-500")).not.toContain("bg-white");
```

**But the bug it points at is real and pre-existing** — `statusConfig[DRAFT].color` was `bg-muted`, the inactive tab's own surface, so the default-state dot was invisible against its own tab. Fixed here as `bg-muted-foreground`, with a test that fails if any status colour ever matches the tab surface again. Confirmed in the browser: all four dots now render distinct from their surface.

Both passes are on the record per the review invariants.

PR #622, opened 2026-09-09
fix(a11y): sweep the whole src tree for gray literals, ledger the debt (#620)

### LINKED ISSUE #620: [P2.28] The #331 gray-literal guard is a hand-maintained path list: 14 files / 272 literals on the core authoring flow are unguarded
Split out of #618, where the AC asked to "audit what else that list should be covering — a hand-maintained path list is the reason this shipped".

## The measurement

`frontend/src/__tests__/theme/core-authoring-pages-tokens.test.ts` (#331) bans hardcoded `*-gray-\d+` literals, but only across an explicit `PAGES` array. #618 shipped a 1.46:1 label because `ChapterTab.tsx` was simply not on it.

Walking imports transitively from the four core authoring routes plus `TocGenerationWizard` and the dashboard layout: **80 files reachable, 14 still carrying 272 gray literals.**

| literals | file |
|---|---|
| 62 | `app/dashboard/books/[bookId]/export/page.tsx` |
| 44 | `components/toc/TocReview.tsx` |
| 34 | `components/toc/ClarifyingQuestions.tsx` |
| 25 | `app/dashboard/books/[bookId]/edit-toc/page.tsx` |
| 24 | `components/toc/NotReadyMessage.tsx` |
| 20 | `components/BookMetadataForm.tsx` |
| 16 | `components/toc/TocGenerating.tsx` |
| 14 | `components/navigation/ChapterBreadcrumb.tsx` |
| 13 | `components/toc/ErrorDisplay.tsx` |
| 8 | `components/toc/ReadinessChecker.tsx` |
| 6 | `components/errors/ErrorNotification.tsx` |
| 3 | `components/ui/ChapterStatusIndicator.tsx` |
| 2 | `components/ui/avatar.tsx` |
| 1 | `components/export/ExportOptionsModal.tsx` |

The sharpest signal: **every child step of `TocGenerationWizard` is unguarded even though the wizard itself is on the list** — 139 literals across `TocReview`, `ClarifyingQuestions`, `NotReadyMessage`, `TocGenerating`, `ErrorDisplay`, `ReadinessChecker`. The guard covers the shell and none of the screens the user actually reads.

Each of these is a #331/#610/#618 waiting to happen. Not every literal is a contrast failure — but nothing currently tells us which are.

## Second finding: a token pair, not a file

Also from #618: `--muted-foreground` `#737373` on `--muted` `#f5f5f5` measures **4.34:1** in the light theme, under the 4.5:1 AA floor. That is real axe-core output, not arithmetic. #618 routed around it by using `text-foreground` for the tab title, but the pair itself is unfixed — anywhere `text-muted-foreground` carries normal-weight body text on a `bg-muted` surface, it is failing today. Worth a sweep of `text-muted-foreground` on `bg-muted`/`bg-secondary`/`bg-accent` (all `oklch(0.269)` in dark, `#f5f5f5` in light).

## Suggested approach

The hand-maintained list is the defect; replace it with something that cannot silently omit a file.

1. Make the guard sweep a **directory** (or the transitive import set) rather than a literal array.
2. That fails on 14 files immediately, so ledger the known-unfixed ones the way `security-baseline.json` already does for advisories in this repo — a file with a reason per entry, so adding a new gray literal fails CI while the existing backlog is explicit and countable.
3. Burn the ledger down, highest-traffic first (the TOC wizard screens and the export page).
4. Separately, decide whether `--muted-foreground` should be darkened in light mode or whether `text-muted-foreground` should stop being used for body text on muted surfaces.

Steps 1–2 are the ones that stop the bleeding; 3 can be incremental.

## Acceptance criteria

- [ ] The guard fails when a gray literal is added to any file on the core authoring flow, without anyone editing a path list
- [ ] Existing violations are ledgered with reasons, not silently excluded
- [ ] Mutation-checked: add a gray literal to a previously-unguarded file (e.g. `TocReview.tsx`) → the guard fails
- [ ] `--muted-foreground` on `--muted` in light mode either clears 4.5:1 or is documented as not-for-body-text
- [ ] The static-walk caveat is addressed: dynamic imports are either covered or explicitly out of scope

## Caveat

The walk above is **static**; a component reached only through a dynamic import would not appear. The counts are a floor, not a ceiling.

frankbria:
Closes #620.

## The defect

The #331 guard banned hardcoded `*-gray-N` Tailwind literals across a **hand-maintained four-entry `PAGES` array**. That list is why #618 shipped a 1.46:1 label — `ChapterTab.tsx` was simply not on it — and why **every child step of `TocGenerationWizard` was unguarded while the wizard shell itself was guarded**. The guard covered the shell and none of the screens the user actually reads.

A list you have to remember to edit is not a guard.

## What changed

**1. The sweep is now the whole tree.** Every shipped `.ts`/`.tsx` under `frontend/src/` (182 files), checked against `frontend/gray-literal-baseline.json` — the same ledger shape `security-baseline.json` uses for advisories. **34 files / 464 literals / 174 rows**, each file with a reason. Adding a literal anywhere fails; burning one down does not.

Whole-tree rather than the transitive import set is deliberate: **an import walk is exactly the mechanism that can silently omit a file** — the issue's own "static walk misses dynamic imports" caveat — which is the defect being fixed. Covering everything retires the caveat by construction instead of documenting around it.

Test sources are explicitly out of scope: they aren't shipped UI, several legitimately assert on gray class names, and this guard's own doc comment carries `*-gray-N` literals, so scanning tests would self-match.

**2. Counts are per distinct literal, not one total per file.** Pre-PR review (codex, below) found the loophole in the first cut: a single total lets a change drop one `text-gray-500` and add one `text-gray-300` for a net-zero count — precisely the wrong-shade edit that shipped #618.

**3. The second #618 finding is fixed, not documented away.** Light `--muted-foreground` `#737373` on `--muted` `#f5f5f5` is **4.34:1** in real axe-core output, under the AA floor — and `text-muted-foreground` is used on 65 components. `oklch(0.556)` → `oklch(0.54)` (`#6e6e6e`, **4.68:1**), light theme only; every other surface the token lands on only improves. The new guard checks it against `background`/`card`/`muted`/`secondary`/`accent` in **both** themes and pins the maths to axe's observed 4.34 figure.

**4. Shared contrast helpers.** The browser-validated oklch→sRGB conversion moved out of the #610 guard into `src/__tests__/theme/helpers/contrast.ts`.

## Verification

Full suite: **139 suites, 2345 passed, 0 failed**, coverage gates green. Lint 0 errors.

Six mutations, each reverted with `git diff --quiet HEAD --` asserted:

| # | Mutation | Result |
|---|---|---|
| 1 | Gray literal added to `TocReview.tsx` (ledgered; **unguarded before this PR**) | ✅ fails — 47 > 44 |
| 2 | Gray literal added to `ChapterTab.tsx` (zero literals, unledgered) | ✅ fails — listed as unledgered |
| 3 | Brand-new `MutationProbe.tsx` nobody registered anywhere | ✅ fails — listed as unledgered |
| 4 | `--muted-foreground` reverted to `oklch(0.556)` | ✅ fails — 4.349 < 4.5 on muted/secondary/accent |
| 5 | `SRC_ROOT` narrowed to `src/types` (guard-the-guard) | ✅ fails the vacuity assertion |
| 6 | **Net-zero swap**: one `text-gray-400` → `text-gray-200` in `TocReview.tsx` | ✅ fails — file total still 44 |

Mutation 6 is the codex finding; it passed before the per-literal change and fails after.

## Third-party review

`opencode` (GLM) returned an upstream server error (`err_f1fe58b5`) on its one attempt, so this fell back to `codex review --base main` per the documented path. Codex raised one **[P2]**: the count-only swap loophole. **Fixed and mutation-checked** in `d04a15d` (mutation 6). No other findings.

## Known limitations

- **`npm run typecheck` does not cover these files.** `frontend/tsconfig.json` excludes `**/__tests__/**`, so `tsc` never sees any test file in this repo — pre-existing and repo-wide, not introduced here. These four files were typechecked directly with an explicit strict `tsc` invocation (exit 0). Un-excluding tests repo-wide is out of scope for #620.
- **The ledger is 464 literals of real debt.** This PR stops the bleeding and makes the backlog explicit and countable; it does not burn it down. The TOC-wizard screens and the export page are the highest-traffic entries.
- **Out of AC, measured anyway** (the issue's follow-up comment asked for it when this flow is swept): the theme-fixed status dots against the 3:1 non-text floor of WCAG 1.4.11. `bg-blue-500` and `bg-purple-500` pass on every surface in both themes; the DRAFT `bg-muted-foreground` dot passes in both. **`bg-green-500` (COMPLETED) fails in light mode — 2.28:1 on `bg-background`, 2.09:1 on `bg-muted`**, and the dot's `border-border` is 1.09:1 so it provides no rescue. Filed as a follow-up rather than widened into this PR.


frankbria:
## Third-party review — post-PR pass

**Reviewer**: `codex review --base main` (fallback; `opencode`/GLM returned an upstream server error `err_f1fe58b5` on its one attempt, per the documented one-attempt-then-fallback path).

**Run at**: `708f3a7` — re-run rather than replayed, because the diff changed after the pre-PR pass.

**Verdict**: no findings.

> The changes add a broader gray-literal regression guard, shared contrast helpers, and a light-theme muted foreground adjustment without introducing a clear functional regression. I did not find any discrete, actionable bugs in the diff.

### Pre-PR pass (for the record)

The earlier pass at `2757802` raised one **[P2]**:

> **Detect gray literal swaps in ledgered files** — for files already in `gray-literal-baseline.json`, this only compares the total number of gray literals to the recorded count. If a change removes one existing gray literal and adds a new inaccessible one elsewhere in the same ledgered file, `actual` stays equal to `ledgered` and CI passes, which contradicts the guard's stated goal.

Verified as real by construction and **fixed in `d04a15d`** — the ledger now counts per distinct literal. Mutation 6 in the PR body is that exact scenario: swapping one `text-gray-400` for `text-gray-200` in `TocReview.tsx` keeps the file at 44 literals and now fails the guard. It passed before the fix.

frankbria:
## Demo verification — outcome evidence per acceptance criterion

Every criterion is mapped to an observed outcome, not a "it rendered / exited 0".

### AC1 — the guard fails when a gray literal is added to any file on the core authoring flow, without anyone editing a path list

Six mutations against the committed tree, each reverted with `git diff --quiet HEAD --` asserted afterwards (the repo's stale-`index.lock` failure mode makes an unasserted revert worthless):

| # | Mutation | Observed |
|---|---|---|
| 1 | Gray literal added to `TocReview.tsx` — ledgered, and **not covered by the old `PAGES` list at all** | ❌ `has no ledgered file above its recorded count` → `actual: 47, ledgered: 44` |
| 2 | Gray literal added to `ChapterTab.tsx` — zero literals today, so unledgered | ❌ `has no gray literals outside the ledger` → `+ "src/components/chapters/ChapterTab.tsx"` |
| 3 | Brand-new `src/components/MutationProbe.tsx`, registered nowhere | ❌ `+ "src/components/MutationProbe.tsx"` |
| 5 | `SRC_ROOT` narrowed to `src/types` (can the guard pass vacuously?) | ❌ `sweeps the whole src tree, so no file can be silently omitted` |
| 6 | **Net-zero swap**: one `text-gray-400` → `text-gray-200` in `TocReview.tsx`; file total stays **44** | ❌ `literal: "text-gray-200", actual: 1, ledgered: 0` |

Mutation 3 is the one that matters most for this AC: a file nobody added to any list is caught, because there is no list.

### AC2 — existing violations are ledgered with reasons, not silently excluded

`frontend/gray-literal-baseline.json`: **34 files, 464 literals, 174 per-shade rows**, every file carrying a prose reason. A fourth test fails if any entry has an empty reason, an empty literal map, or names a file that no longer exists — so the ledger can't rot into an ignore-list.

### AC3 — mutation-checked: a gray literal added to a previously-unguarded file → the guard fails

Mutation 1 above, on the exact file the issue names (`TocReview.tsx`). ✅

### AC4 — `--muted-foreground` on `--muted` in light mode clears 4.5:1

**Real axe-core 4.13.0 in headless Chromium against the running dev server**, not arithmetic. The probe is built from real app classes (`bg-muted` + `text-muted-foreground`) so the compiled stylesheet supplies both colours; nothing is a literal. Theme was pinned via `localStorage` in an init script and the rendered `<html>` class asserted, because next-themes persists across reloads and will otherwise run both halves of a comparison in the same theme.

```
=== BEFORE — --muted-foreground: oklch(0.556) (pre-PR) ===
theme class    : "scroll-smooth light" | stored: light
violations     : 1 | incomplete: 0
  -> #737373 on #f5f5f5 = 4.34:1 (needs 4.5:1)

=== AFTER — --muted-foreground: oklch(0.54) (this PR) ===
theme class    : "scroll-smooth light" | stored: light
violations     : 0 | incomplete: 0
  -> #6f6f6f on #f5f5f5 = 4.6:1 (needs 4.5:1)
```

The before number reproduces the **4.34:1** the issue reported, from axe itself — so the fix is what flipped it, verified in both directions. (Computed style came back as `lab(46.64 0 0)`, which is why the ratio is axe's and not hand-rolled off `getComputedStyle`.)

Dark mode is unaffected by construction — the change is inside `:root` only — and `muted-foreground-contrast.test.ts` covers dark arithmetically (5.86:1 on `--muted`). Mutation 4: reverting the token to `oklch(0.556)` fails that suite on `--muted`, `--secondary` and `--accent`. ✅

### AC5 — the static-walk caveat is addressed

Addressed by **removing the walk**. The guard globs the tree rather than following imports, so dynamic imports are covered rather than excluded — there is no import graph to miss anything. Test sources are the one documented exclusion, with the self-match reason stated in the file.

### Gates

- Full frontend suite: **139 suites / 2345 passed / 0 failed**, coverage thresholds green.
- Lint: 0 errors (446 pre-existing warnings, unchanged).
- CI: `Frontend Tests` ✅ 2m55s · `Backend Tests` ✅ 2m58s · `Security Audit` ✅ · `Build backend image` ✅ · GitGuardian ✅.
- Typecheck caveat is in the PR body: `tsconfig` excludes `**/__tests__/**` repo-wide, so these four files were checked with a direct strict `tsc` invocation (exit 0) rather than credited to a `npm run typecheck` that never sees them.

Issue #623, opened 2026-09-09
[P2.29] The COMPLETED chapter status dot is 2.09:1 in light mode — bg-green-500 fails WCAG 1.4.11


frankbria:
Split out of #620, where the theme-fixed status-dot colours were measured while sweeping the core authoring flow. The #620 follow-up comment asked for exactly this and noted axe-core cannot see it.

## The measurement

`ChapterTab.tsx` and `ChapterStatusIndicator.tsx` encode chapter status as a 2×2 `rounded-full` dot whose colour is the only at-a-glance signal (the tab's `aria-label` is `Open chapter {title}` — no status; the label lives in the tooltip). That makes it a non-text UI component under **WCAG 2.1 1.4.11**, whose floor is **3:1**, not 4.5:1. `axe-core`'s `color-contrast` rule does not check non-text elements at all, so a clean axe run says nothing here — these are Tailwind defaults, computed against the real surface tokens:

| dot | light `bg-background` | light `bg-muted` | dark `bg-background` | dark `bg-muted` |
|---|---|---|---|---|
| `bg-blue-500` (IN_PROGRESS) | 3.68 ✅ | 3.37 ✅ | 5.38 ✅ | 4.11 ✅ |
| `bg-purple-500` (PUBLISHED) | 3.96 ✅ | 3.63 ✅ | 5.00 ✅ | 3.82 ✅ |
| `bg-muted-foreground` (DRAFT, fixed in #618) | 5.10 ✅ | 4.68 ✅ | 7.66 ✅ | 5.86 ✅ |
| **`bg-green-500` (COMPLETED)** | **2.28 ❌** | **2.09 ❌** | 8.69 ✅ | 6.64 ✅ |

`green-500` is `#22c55e`. It is the single failure, and it fails in the **light** theme on both surfaces the dot is ever painted on.

The dot also carries `border border-border`, which does not rescue it: light `--border` `oklch(0.922)` = `#ebebeb` against `bg-muted` `#f5f5f5` is **1.09:1**, i.e. the border is itself invisible in light mode.

`bg-blue-500` on light `bg-muted` at 3.37 and `bg-purple-500` on dark `bg-muted` at 3.82 are passes, but thin ones — worth not regressing.

## Suggested approach

The other three status colours are theme-fixed brand accents that happen to clear the bar; `green-500` does not. Options, cheapest first:

[...]

PR #633, opened 2026-09-10
fix(a11y): ChapterTab spinner ring is 2.33:1 in light mode with a theme-fixed white core (#629)

### LINKED ISSUE #629: [P2.31] The ChapterTab loading spinner is 2.33:1 in light mode and its core is a hardcoded bg-white
Found while sweeping `ChapterTab`'s indicator row for #623 (PR #628). #623 fixed every dot on that row; the spinner is the one indicator left, and it needs a redesign rather than a colour swap, so it was split out instead of widening that PR.

## The measurement

`ChapterTab.tsx` renders the per-chapter loading indicator as:

```tsx
<div className="w-3 h-3 border-2 border-blue-400 border-t-transparent rounded-full animate-spin bg-white" />
```

Two problems, both non-text UI components under **WCAG 2.1 1.4.11** (3:1 floor, which `axe-core`'s `color-contrast` rule does not check):

| part | light `--background` | light `--muted` | dark `--background` | dark `--muted` |
|---|---|---|---|---|
| `border-blue-400` ring | 2.54 ❌ | **2.33 ❌** | 7.79 ✅ | 5.95 ✅ |
| `bg-white` core | 1.00 ❌ | 1.06 ❌ | 18.9 ✅ | 14.6 ✅ |

The ring is the spinner's only shape, and it fails in the light theme on both surfaces the tab is ever painted on. The `bg-white` core is theme-fixed: in light mode it is the surface colour, so the spinner reads as a thin ring rather than a disc; in dark mode it is a bright white dot that the rest of the tab strip does not have.

Measured arithmetically from the `globals.css` oklch tokens using `src/__tests__/theme/helpers/contrast.ts`, not from `getComputedStyle` — see the #620 note on why.

## Why it is not a one-word fix

`blue-600` would clear light (3.56 / 3.26) but drop dark-muted to 4.25 — fine. The harder part is `bg-white`, which has no dark counterpart and is doing layout work (it hides whatever is behind the ring). The right shape is probably `bg-background` plus a `dark:` ring variant, or dropping the fill entirely and letting the ring spin over the tab surface — which changes how the indicator looks, so it wants a look at the rendered component rather than arithmetic alone.

The repo already has a `LoadingStateManager` / `ProgressIndicator` pair (`docs/references/component-documentation.md`); worth checking whether this hand-rolled spinner should just use one of them.

## Acceptance criteria

- [ ] The spinner ring clears 3:1 against both `--background` and `--muted`, in both themes
- [ ] The core is theme-aware — no hardcoded `bg-white`
- [ ] Covered by the #623 guard (`src/__tests__/theme/status-dot-contrast.test.ts`), extending its `DOT_SOURCES` table, with the per-file match count updated so it cannot go vacuous
- [ ] Mutation-checked: restoring `border-blue-400` fails the guard
- [ ] Verified visually in the running app in both themes, since the fill change alters how the indicator reads

frankbria:
Closes #629.

## The failure

`ChapterTab.tsx` rendered the per-chapter loading indicator as a hand-rolled div:

```tsx
<div className="w-3 h-3 border-2 border-blue-400 border-t-transparent rounded-full animate-spin bg-white" />
```

Two WCAG 2.1 **1.4.11** failures (3:1 floor for non-text UI components — a rule
`axe-core`'s `color-contrast` check does not evaluate at all):

| part | light `--background` | light `--muted` | dark `--background` | dark `--muted` |
|---|---|---|---|---|
| `border-blue-400` ring | 2.54 ❌ | **2.33 ❌** | 7.79 ✅ | 5.95 ✅ |
| `bg-white` core | 1.00 ❌ | 1.06 ❌ | 18.9 ✅ | 14.6 ✅ |

The ring is the spinner's only shape and it failed on both surfaces the tab is
ever painted on in light mode. The `bg-white` core had no dark counterpart: the
surface colour in light (so the spinner read as a thin ring rather than a disc),
a bright white dot in dark that nothing else on the tab strip has.

## The fix

Replaced with the icon spinner already used elsewhere in the app (see
`ChapterEditor`'s save indicator):

```tsx
<HugeiconsIcon icon={Loading03Icon} size={12} className={cn('animate-spin', LOADING_SPINNER)} />
```

`Loading03Icon` has no fill, so the theme-blind core is **gone** rather than
re-coloured — that was the part the issue flagged as needing a redesign, not a
colour swap. The stroke keeps the indicator row's blue "working" signal, per
theme: `text-blue-600 dark:text-blue-400`.

Measured after the change, from the `globals.css` oklch tokens:

| | light `--background` | light `--muted` | dark `--background` | dark `--muted` |
|---|---|---|---|---|
| spinner stroke | **5.17** ✅ | **4.74** ✅ | **7.79** ✅ | **5.95** ✅ |

> The issue predicted 3.56 / 3.26 for the light pair. Those are Tailwind **v4**'s
> `blue-600`; this repo is still on v3 (see #513), where it is comfortably higher.
> Noted in a code comment so the next reader does not think the guard drifted.

## Guard

Extends the #623 guard, `src/__tests__/theme/status-dot-contrast.test.ts`:

- `dotColor` → `indicatorColor`, resolving `bg-` **or** `text-` **or** `border-`
  utilities. Colour reaches these indicators as a fill, an icon stroke or a ring
  depending on the indicator; whichever it is, it is the shape the user sees.
  No existing `DOT_SOURCES` entry captures a `text-`/`border-` utility, so this
  does not change how any of them resolve — case count went 57 → 62, entirely
  from the new entry.
- New `DOT_SOURCES` entry for `LOADING_SPINNER`, with its own `count: 1` vacuity
  assertion.
- Pins the **2.33** light-`--muted` figure for the old `border-blue-400`, the way
  the file already pins 2.09 for the old `bg-green-500`.

**Mutation-checked**, both directions:

| mutation | result |
|---|---|
| `LOADING_SPINNER = 'border-blue-400'` | ❌ 2 failed — `light theme is at least 3:1 on --muted` |
| rename the const (guard goes vacuous) | ❌ vacuity assertion: `count 1` → `count 0` |

Each revert asserted clean with `git diff --quiet HEAD --` before the next.

## Verification

- `pre-commit`: lint, typecheck, full frontend unit suite, coverage ≥85% — all green.
- Guard: 62 passed.
- Visual check in the running app, both themes — see the demo comment below.

## Review record

- **opencode (GLM)** — attempted, failed with a server-side `UnknownError`
  (`err_4edac537`). Not retried, per the repo's one-attempt-then-fall-back note.
- **`codex review --base main`** — ran clean: *"I found no discrete correctness,
  accessibility, or maintainability issues introduced by this diff."*

## Known limitations

- `MobileChapterTabs` has no per-chapter loading indicator at all, so there is
  nothing to bring along; the mobile tab strip is unaffected by this change.
- Other hand-rolled `animate-spin` rings elsewhere in the app are **not** touched
  here. Measured and filed as **#635**: `border-gray-400` and `border-blue-400`
  fail identically (2.33 on light `--muted`) at two more sites; the four
  `border-indigo-500` full-page loaders clear the floor everywhere.
- The dark-theme demo surfaced an unrelated, more severe defect on the same
  component — the errored tab's `bg-red-50` has no dark variant, putting its
  label at **1.05:1**. Different failure class (1.4.3 text, not 1.4.11 non-text)
  and a different line, so filed as **#634** rather than folded in.


frankbria:
## Demo — Phase 11 verification

Rendered the **real `ChapterTab`** in the running Next dev server (temporary
route, removed afterwards; working tree confirmed clean), four rows — loading,
unsaved, completed, errored — on both `bg-background` and `bg-muted`, screenshot
in each theme. Theme was asserted from `document.documentElement.className`
between shots (`light` → `dark`), not assumed, because `next-themes` persists to
`localStorage` and a reload can silently keep the previous theme.

| criterion (#629) | evidence | outcome |
|---|---|---|
| Ring clears 3:1 on `--background` **and** `--muted`, both themes | guard `status-dot-contrast.test.ts`, 62 passed | ✅ 5.17 / 4.74 light, 7.79 / 5.95 dark |
| Core is theme-aware — no hardcoded `bg-white` | `grep bg-white src/components/chapters/ChapterTab.tsx` → no match; `Loading03Icon` has no fill at all | ✅ removed, not recoloured |
| Covered by the #623 guard, `DOT_SOURCES` extended, per-file count updated | new `LOADING_SPINNER` entry, `count: 1`; case count 57 → 62 | ✅ |
| Mutation-checked: restoring `border-blue-400` fails the guard | `LOADING_SPINNER = 'border-blue-400'` → **2 failed**, incl. `light theme is at least 3:1 on --muted`. Renaming the const → vacuity assertion trips, `count 1` → `count 0` | ✅ both directions |
| Verified visually in the running app, both themes | screenshots below | ✅ |

**Light** — the spinner is a legible blue mark on both surfaces; it no longer
reads as a hairline ring around a surface-coloured disc.
**Dark** — the spinner is legible on both surfaces, and the bright white core
that the rest of the tab strip did not have is gone.

### Found while demoing — filed, not fixed here

The dark screenshot shows the **errored** row painting `bg-red-50` with no dark
variant (`ChapterTab.tsx:94`), so its label is near-white on near-white and
effectively unreadable. That is a different failure class from #629 (text
contrast, 1.4.3, not non-text 1.4.11) and a different line, so it is filed
separately rather than folded in here.

PR #631, opened 2026-09-10
fix(a11y): burn 282 gray literals to tokens — 29 measured contrast violations to 0 (#624)

frankbria:
Closes #624.

Real axe-core in Chromium across **5 routes and all 6 TOC-wizard steps, in both themes, before and after**. **29 `color-contrast` violations → 0.** Ledger: **34 files / 464 literals → 22 / 182**, with **12 entries deleted, not recounted**.

## What the measurement said, versus what the counts said

The issue's central question was which ledgered literals are real failures and which are merely theme-blind. The counts pointed one way; the browser pointed another.

The **six TOC-wizard steps — 139 literals, the largest single block — produced almost no axe violations.** They paint self-consistent dark cards (`text-gray-300` on `bg-gray-900`), so each fg/bg *pair* stays legible even sitting on a white page. They were theme-blind, not low-contrast. The severe failures were in files with much smaller counts:

| where | theme | measured | element |
|---|---|---|---|
| export, edit-toc | light | **1.1:1** | `<h1 class="text-gray-100">` — the page title, invisible |
| edit-toc | light | **1.2:1** ×5 | "Tips" `<li class="text-gray-400">` on `bg-gray-800/50` |
| edit-toc, review | light | **1.94:1** ×8 | `ChapterStatusIndicator` label on the gray-800 TOC cards |
| book-detail | light | 2.53:1 ×2 | `ChapterBreadcrumb` |
| export | light | 2.53:1 | subtitle `text-gray-400` |

## The regression the measurement existed to catch

Converting the surfaces moved **every accent that had been sitting on a dark box onto a light one**. A blind sweep would have traded old failures for new ones — precisely what the issue warned about. The post-conversion scan caught **seven violations that did not exist before the fix**:

| new failure | measured | fix |
|---|---|---|
| `text-green-400` (ClarifyingQuestions, TocReview) | 1.59, 1.74 | `text-green-700 dark:text-green-400` |
| `text-purple-400` (TocReview stat) | 2.64 | `text-purple-700 dark:text-purple-400` |
| `text-red-400` (NotReadyMessage, ClarifyingQuestions) | 2.76 | `text-red-700 dark:text-red-400` |
| `text-gray-400` ×3 in the GENERATING step | 2.32 | from `LoadingStateManager`/`ProgressIndicator` — **components the wizard renders but that were not on the target list** |

No dark value moved; each accent gained a light-mode partner.

[...]

Issue #625, opened 2026-09-09
[P3.38] npm run typecheck covers no test file in the repo — tsconfig excludes **/__tests__/**, and SWC strips types without checking


frankbria:
Found while shipping #620, and it is the same shape as the finding recorded in the repo's `gate-must-cover-the-file` lesson (which caught `**/e2e/**` and `**/playwright*.ts`). The exclusion list is broader than that lesson recorded.

## The gap

`frontend/tsconfig.json`:

```json
"exclude": [
  "node_modules",
  "**/*.test.ts",
  "**/*.test.tsx",
  "**/__tests__/**",
  "**/jest.setup.ts",
  "**/playwright*.ts",
  "**/e2e/**"
]
```

So `npm run typecheck` (`tsc --noEmit`) sees **zero test files** — not the 170-odd files under `src/__tests__/`, `__mocks__/`, or any `*.test.tsx` anywhere. Confirmed with `npx tsc --noEmit --listFiles | grep <path>`: no match for any file in the #620 diff.

Nothing else covers them either. Jest runs through `next/jest`, which transpiles with SWC and **strips types without checking them**. A test file can reference a prop that no longer exists, pass the wrong argument type to a helper, or `await` a non-promise, and every gate in the repo stays green.

This matters more than usual here because several tests are *guards* — `theme/core-authoring-pages-tokens.test.ts`, `theme/muted-foreground-contrast.test.ts`, `theme/dark-primary-text-contrast.test.ts` are the entire enforcement for #331/#610/#618/#620, and they parse CSS and JSON by hand. A type error in a guard is a guard that may not be guarding.

Pre-existing and repo-wide; #620 worked around it by typechecking its four files with an explicit one-off `tsc` invocation (exit 0), which is not a gate anyone will remember to re-run.

## Suggested approach

Do not just delete the exclusions — that will surface an unknown number of pre-existing errors across ~170 files and block the next PR that touches any of them. Measure first:

1. Run `tsc --noEmit` with the test exclusions removed and count/classify the errors. That number decides everything below.
2. If small: fix them and drop the exclusions outright.
3. If large: add a second `tsconfig.test.json` that extends the base, includes tests, and is wired to a separate `npm run typecheck:tests` script — then ledger the known-failing files (the way `security-baseline.json` and, since #620, `gray-literal-baseline.json` do) so new test code is checked while the backlog stays explicit and countable.
4. Whichever path, the check has to run in the `Frontend Tests` CI job, not just locally.

[...]

Issue #624, opened 2026-09-09
[P2.30] Burn down the 464-literal gray ledger, measuring which are real contrast failures first


frankbria:
Split out of #620, which stopped the bleeding but explicitly deferred the burn-down ("steps 1–2 are the ones that stop the bleeding; 3 can be incremental").

## Where it stands

`frontend/gray-literal-baseline.json` now ledgers **34 files / 464 literals / 174 per-shade rows**, each with a reason. The guard fails on any new literal — including a same-total shade swap — so the debt can only shrink. It has not shrunk yet.

The concentration is on the core authoring flow, and #620's framing still holds: *"Each of these is a #331/#610/#618 waiting to happen. Not every literal is a contrast failure — but nothing currently tells us which are."* That is the actual open risk: the ledger counts the debt, it does not tell you which entries are shipping an unreadable element **right now**.

Highest-traffic first, from the ledger:

| literals | file |
|---|---|
| 62 | `app/dashboard/books/[bookId]/export/page.tsx` |
| 44 | `components/toc/TocReview.tsx` |
| 40 | `components/BookCreationWizard.tsx` |
| 34 | `components/toc/ClarifyingQuestions.tsx` |
| 31 | `components/toc/TocSidebar.tsx` |
| 25 | `app/dashboard/books/[bookId]/edit-toc/page.tsx` |
| 24 | `components/toc/NotReadyMessage.tsx` |

The six TOC-wizard child steps account for **139** on their own.

## Suggested approach

Measurement before mass edits — a blind `text-gray-400` → `text-muted-foreground` sweep is how #618's *first* attempted fix would have shipped a new light-mode failure.

1. Render each high-traffic screen in **both** themes and run real axe-core against it, the way #618 and #620 did. That produces a ranked list of which literals are actually failing, versus which are merely theme-blind.
2. Fix the measured failures first, then convert the rest to tokens file by file, shrinking the ledger entry each time and deleting it at zero.
3. Prefer `text-foreground` / `text-muted-foreground` / `bg-muted` / `border-border`. Note `--muted-foreground` on `--muted` is 4.68:1 since #620 — fine for body text, but it has no headroom, so do not stack it on a lighter surface variant without re-measuring.

## Acceptance criteria

- [ ] Every high-traffic screen above has a real axe-core result in both themes, recorded, with any `color-contrast` violation fixed
- [ ] The ledger shrinks — entries deleted, not recounted — with the six TOC-wizard steps and the export page at zero
[...]

PR #636, opened 2026-09-10
fix(a11y): burn 17 theme-blind dark translucent overlays, guard the class (#632)

### LINKED ISSUE #632: [P2.32] Theme-blind dark translucent overlays: 19 sites where a light accent sits on a bg-<hue>-900/NN authored for a dark page
Split out of #624 (PR #631), where converting the TOC wizard and export page from `*-gray-N` literals to theme tokens turned their surfaces light and stranded every accent that had been painted on top of them. Eleven such sites were found and fixed inside that PR's file set. This issue covers the rest of the tree and the guard that should stop the class recurring.

## The pattern

A **dark translucent overlay with no `dark:` prefix** — `bg-<hue>-{700,800,900}/NN` — is by construction authored for a dark page. Composited over a light surface it turns to mud, and the light `text-<hue>-{300,400}` glyph or label sitting on it goes with it. Measured examples from #624, all in the browser or from the `globals.css` tokens:

| site | before | after |
|---|---|---|
| `bg-yellow-900/20` badge + `text-yellow-400` icon | **1.02:1** | 4.76 |
| `bg-green-900/30` badge + `text-green-400` icon | **1.05:1** | 4.79 |
| `bg-amber-900/20` + `text-amber-300` warning text | **1.05:1** | pass |
| `bg-red-900/20` badge + `text-red-400` icon | **1.76:1** | 5.91 |

The fix shape is always the same: `bg-<hue>-50 dark:bg-<hue>-900/NN`, `text-<hue>-700 dark:text-<hue>-400`, `border-<hue>-300 dark:border-<hue>-700`.

## What is left

Nineteen unprefixed overlays across seven files that #624 did not convert:

| count | file |
|---|---|
| 5 | `components/errors/ErrorNotification.tsx` |
| 3 | `components/examples/ai-error-handling-example.tsx` |
| 2 | `components/chapters/ChapterTabs.tsx` |
| 1 | `components/BookMetadataForm.tsx` |
| 1 | `app/page.tsx` |
| 1 | `app/dashboard/page.tsx` |
| 1 | `app/dashboard/books/[bookId]/summary/page.tsx` |

(Remaining count is by unprefixed-overlay occurrences; several of these files also still carry gray literals in `gray-literal-baseline.json`, so the two burn-downs overlap and are worth doing together per file.)

## Why a guard, and why axe will not be it

**Six of the eight failures #624 fixed were icon badges, and axe reported zero on every one.** Its `color-contrast` rule does not evaluate non-text elements at all — these are WCAG 1.4.11 at 3:1 — so a green axe run is not evidence here, exactly as #623 found for the status dots. The guard has to be static.

Proposed shape, mirroring `gray-literal-baseline.json` and the #623 status-dot guard:

- Sweep every shipped `.ts`/`.tsx` under `src/` for `(^|[^:])bg-<hue>-(700|800|900)/\d+`.
- Ledger the known 19 in a baseline file with a reason each, so adding a new one fails while burning one down does not.
- Include the **stale-row assertion** added in #624: a ledgered value the file no longer contains must be deleted, so the ledger cannot pre-authorise.
- Mutation-check it: add an unprefixed overlay to an unledgered file, and to a ledgered one.

## Acceptance criteria

- [ ] Guard exists, sweeps the whole `src/` tree, and is mutation-checked against both an unledgered and a ledgered file
- [ ] All 19 sites converted to `dark:`-prefixed pairs, each measured in both themes (arithmetically for the non-text ones — axe cannot see them)
- [ ] The baseline file empties and is deleted, or every remaining row carries a reason
- [ ] No `dark:`-only fix that regresses light, and no light-only fix that regresses dark

frankbria:
Closes #632.

An unprefixed `bg-<hue>-{700,800,900}/NN` is by construction authored for a dark
page. Composited over a light surface it turns to mud and takes the light accent
on top with it — the failure #624 (PR #631) hit when it converted the TOC wizard
and export page onto theme tokens.

**Six of the eight failures #624 fixed were icon badges, and axe reported zero on
every one** — its `color-contrast` rule does not evaluate non-text elements at
all. So the guard here is static, not an axe run.

## The count was wrong three ways, and the sweep is how we know

The issue's title says nineteen sites. Its own file table sums to **fourteen**.
The sweep finds **eighteen**: the table missed `ui/toaster.tsx` entirely, because
its four overlays carry a `group-[.toast]:` variant prefix that a hand-written
grep for an *unprefixed* class cannot match. That is the #620 lesson again — the
sweep finds what the list forgets.

Seventeen burn down here. The four in `ui/toaster.tsx` are ledgered, and
deliberately: that component hardcodes `theme="dark"` on the Sonner root and
paints `bg-gray-900` in **both** app themes, so its severity tints composite over
a surface that is pinned dark, not over an unknown page. Adding `dark:` there
would delete the tint in light mode rather than fix it. The real fix is making
the toast follow the app theme — a behaviour change, which
`gray-literal-baseline.json` already defers for the same file.

## Measured, in a real browser

A temporary route rendering all five `ErrorNotification` severities, captured in
Chromium with real axe-core, both themes:

| | light | dark |
|---|---|---|
| **before** | **32** `color-contrast` violations | — |
| **after** | **0** | **0** |

Among the 32: `#ffffff` heading on `#e5d2d2` at **1.45:1**, and `#d1d5db` detail
text on `#e5d2d2` at **1.01:1** — literally invisible. Only AAA
(`color-contrast-enhanced`, 7:1) fires after, which is not this repo's bar.

Axe's own composited figures also validate the new `compositeOver()` helper
against a browser rather than against itself: axe reported background `#210e0e`
and 6.69 where the helper computes 6.70, and `#e5d2d2` matches its arithmetic
over white exactly.

## Two figures in the issue do not hold, and are corrected here

1. **The four "before" ratios reproduce over `--muted`, not `--background`.** Over
   white the same pairs give 1.10 / 1.91 / 1.03 / 1.02 instead of 1.02 / 1.76 /
   1.05 / 1.05. A contrast figure without the surface it was taken against is not
   a measurement, so the guard records the surface with every one. (Per the #629
   lesson about copying a ratio out of an issue body.)
2. **The proposed `border-<hue>-300 dark:border-<hue>-700` pair does not clear
   3:1** — 1.74:1 on its own card in light, 2.14–2.86 in dark. It is not held to
   it, and should not be: every one of these surfaces is an alert carrying
   `role="alert"`, an icon, and literal text naming the condition, so colour is
   never the sole indicator and WCAG 1.4.11 does not attach to the border. It is
   held to a visibility floor instead, documented as a design drift guard rather
   than a WCAG claim. Holding it to 3:1 would force a garish `-600` outline on
   every alert in the app for no accessibility gain.

## `ErrorNotification` needed a pass, not a swap

Its heading was `text-white` and its body `text-gray-300/400/500` on what is now
a light card. Converted to `text-foreground` / `text-muted-foreground` (18.1 and
4.59 light, 17.7 and 7.17 dark), which empties and deletes its
`gray-literal-baseline.json` row. Two dark-only `className` overrides on shadcn
`Button`s were **removed** rather than re-themed — `outline` and `ghost` are
already fully token-based, and the overrides were the only reason the Retry
button rendered white-on-white in light mode. Severity now derives from a single
`hue` value so the card, its stripe and its icon cannot drift apart again.

## Mutation-checked seven ways

All RED, tree verified clean after each revert:

| mutation | result |
|---|---|
| unledgered file gains an overlay | RED |
| ledgered file gains a *new* overlay | RED |
| ledgered file exceeds its recorded count | RED |
| a burned-down ledger row left stale (pre-authorising) | RED |
| the `DARK_TRANSLUCENT` pattern itself broken | RED |
| a fixed site regressed to its pre-#632 shape | RED |
| a ledger row losing its reason | RED |

The pattern-broken case matters most: once the tree is clean the sweep passes
whether or not the regex still works, so a fixture-based anti-vacuity test pins
it independently.

`ErrorNotification.test.tsx`'s severity assertions were **tightened**, not
adjusted to pass: they read `[class*="bg-red-900"]`, which matched the broken
`bg-red-900/20` and the fixed `dark:bg-red-900/20` equally and so could not tell
the bug from the fix.

## Known limitations

- **The stranded-accent half is not guarded.** A light `text-<hue>-{300,400}`
  with no `dark:` prefix is the other side of this bug, and 27 such sites remain
  tree-wide. Out of scope here; follow-up filed.
- **Net is 700/800/900 only.** A 400/500/600 translucent (`bg-amber-500/10`,
  `bg-indigo-600/10`) composites to a pale tint rather than mud.
- **White/black translucents are outside the net.** The three `bg-black/50` modal
  scrims are legitimately theme-fixed.
- 446 pre-existing lint warnings (0 errors) are untouched — #584.

## Verification

- 141 suites / 2450 tests pass; coverage gates green
- `npm run lint` 0 errors, `npm run typecheck` clean
- New test files typechecked with `tsc` directly, since `tsconfig` excludes
  `**/__tests__/**` (#625)
- Third-party review on the record: opencode (GLM) attempted and hit its known
  silent hang on a larger diff (zero bytes, 10 min); `codex review --base main`
  fallback returned **no findings**, having independently re-derived the border
  contrast figures


frankbria:
## Third-party review (pre-PR, Phase 7c-1)

**opencode (GLM-4.6) — primary: no output.** Produced zero bytes in 10 minutes on
this 765-line diff and was killed. This is the repo's known opencode failure mode
on larger diffs, not a clean result — recording it so the run does not read as if
GLM passed the diff.

**`codex review --base main` — fallback: no findings.**

> The changes are focused on replacing theme-blind dark translucent overlays with
> light/dark-aware classes and adding regression coverage. I did not identify any
> discrete, introduced issue that would break existing behavior or tests.

Worth noting what it did on the way there: it read `ui/toaster.tsx` in full
(the ledgered file) and independently re-derived the dark-mode border ratios for
all four hues, getting 2.16–2.85 for `border-<hue>-700` on the composited card.
That is the same sub-3:1 result this PR reports, reached without being told —
which is the corroboration that matters for the "the issue's border idiom does
not clear 1.4.11" claim in the description.

frankbria:
## Review-bot status (not a clean bill of health)

`review / review` did **not** complete on either run — killed before the agent
wrote a result, zero checklist items done. Its own comment says the right thing:
that is not the same as "no defects found".

Both GLM paths failed today, on the same backend:

| reviewer | result |
|---|---|
| `opencode` (GLM-4.6), pre-PR | zero bytes in 10 min, killed |
| `review / review` workflow (GLM) | step killed, no verdict, twice |
| **`codex review --base main`, pre-PR** | **completed, no findings** |

That pairing is #607's signature, and it is why the third-party gate here rests
on codex rather than on GLM. Recording it as a data point for #607.

Merging on the four required checks (Frontend Tests, Backend Tests, E2E Tests,
Security Audit — all green on `0529054`), the codex review, and the browser demo
in the description. `review / review` is not a required check.

Issue #522, opened 2026-08-25
[P3.27] The pip audit gate scans production-only, so advisories in test/dev extras never fail CI — GHSA-6w46-j5rx-g56g is open and unbaselined


frankbria:
Surfaced while verifying #512 against the live Dependabot alert list.

## The gap

`.github/workflows/tests.yml` audits the backend like this:

```bash
uv export --no-dev --no-emit-project --format requirements-txt > /tmp/reqs.txt
uvx pip-audit -r /tmp/reqs.txt --format=json
```

`--no-dev` and the absence of `--all-extras` mean the audited surface is the
**production** dependency set only. Anything reachable solely through the `test` or
`load` extras is invisible to `scripts/audit_gate.py`, so it can never fail CI and
never lands in `security-baseline.json` either — it is not deferred with a reason,
it is simply unseen.

Note the deliberate asymmetry with the neighbouring sync gate, which exports
`--all-extras`: the two steps scan different sets.

## Concrete instance

```
$ gh api "repos/frankbria/auto-author/dependabot/alerts?state=open"
medium  pip  pytest  backend/uv.lock          direct
medium  pip  pytest  backend/requirements.txt unknown
```

`GHSA-6w46-j5rx-g56g` — "pytest has vulnerable tmpdir handling", affects `< 9.0.3`.
`pyproject.toml` pins `pytest==8.4.1`. `Security Audit` is green, because pytest is
in the `test` extra.

(The duplicate row is a separate artifact of the export being counted as a manifest
in the dependency graph — see #512, which stops Dependabot *updating* it but does not
change the graph.)

## Decide, then implement

The gap may be intentional — a test-only advisory does not ship. But it is currently
implicit, and the effect is that GitHub's alert list and CI disagree with no record of
why. Two honest options:

1. **Audit `--all-extras`** and baseline whatever that surfaces, so dev-tool
   advisories are an explicit ledger entry rather than a blind spot. Costs one
   baseline sweep.
2. **Keep production-only** and say so in the step's comment and in
   `docs/references/quality-standards.md`, so nobody reads a green `Security Audit`
   as "no known advisories anywhere".

Either way, decide about `GHSA-6w46-j5rx-g56g`: pytest `8.4.1 -> >=9.0.3` is a major
bump across the whole suite, so it wants its own PR, not a drive-by.

## Acceptance criteria

- [ ] The audited surface is either widened to the extras or documented as
      production-only in both the workflow comment and quality-standards.md.
[...]

PR #651, opened 2026-09-12
chore(security): prune 42 resolved advisories from the audit baseline (#647)

### LINKED ISSUE #647: [P2.36] Remove the stale picomatch baseline line after #648, and prove the audit gate is not vacuous
## Problem

`security-baseline.json` ledgers the high-severity picomatch advisory **with a fix already available upstream**, so the `Security Audit` gate stays green while two vulnerable copies sit in the non-dev dependency tree. No Dependabot PR is filing against them.

The baseline entry is self-incriminating:

```json
"npm": { "GHSA-c2c7-rcm5-vvqj": "picomatch; fix: yes" }
```

`fix: yes` means the advisory is patchable now. Per `CLAUDE.md`, the baseline exists to ledger *pre-existing* advisories so new ones fail CI — not to hold patchable ones indefinitely.

## Current state

Four open Dependabot alerts (#112–#115), all `scope: runtime`, all `frontend/package-lock.json`:

| Alert | Severity | GHSA | Vulnerable range | Patched |
|---|---|---|---|---|
| #112 | high | `GHSA-c2c7-rcm5-vvqj` (ReDoS via extglob quantifiers) | `>= 4.0.0, < 4.0.4` | 4.0.4 |
| #113 | high | `GHSA-c2c7-rcm5-vvqj` | `< 2.3.2` | 2.3.2 |
| #114 | moderate | `GHSA-3v7f-55p6-f55p` (method injection in POSIX character classes) | `>= 4.0.0, < 4.0.4` | 4.0.4 |
| #115 | moderate | `GHSA-3v7f-55p6-f55p` | `< 2.3.2` | 2.3.2 |

`npm audit` in `frontend/` reports `high: 2, total: 2` — and the gate passes anyway because the high GHSA is baselined.

Of the eight installed picomatch copies, **two are still vulnerable and both are in the non-dev tree**:

```
node_modules/picomatch                             2.3.1   dev=false   <- vulnerable (<2.3.2)
node_modules/tinyglobby/node_modules/picomatch      4.0.3   dev=false   <- vulnerable (>=4.0.0,<4.0.4)
node_modules/@rollup/plugin-commonjs/.../picomatch  4.0.5   dev=false   ok
node_modules/@rollup/pluginutils/.../picomatch      4.0.5   dev=false   ok
node_modules/tinyglobby/... (see above)
node_modules/@parcel/watcher/.../picomatch          4.0.7   dev=true    ok
node_modules/jest-haste-map/.../picomatch           4.0.7   dev=true    ok
node_modules/jest-message-util/.../picomatch        4.0.7   dev=true    ok
node_modules/jest-util/.../picomatch                4.0.5   dev=true    ok
```

Both vulnerable copies are deduped roots, not leaves — so an `overrides` entry can reach them.

## Why this matters, and the honest caveat

Dependabot's `scope: runtime` means "reached through a `dependencies` edge", **not** "ships to the browser". Both vulnerable copies most likely enter through build/watch tooling (`tinyglobby`, and the 2.3.1 root via micromatch-family consumers), in which case the glob patterns are ours and not attacker-controlled, and real exploitability is low.

That caveat is the thing to *verify*, not assume — and it does not justify leaving a patchable high-severity advisory baselined either way. Fixing it is cheap; proving it harmless is the expensive path.

## Acceptance criteria

- [ ] Establish which consumers pull `picomatch@2.3.1` and `picomatch@4.0.3` (`npm ls picomatch --all` in `frontend/`), and record whether either reaches the shipped bundle or is build/test-time only.
- [ ] Raise both to patched versions — `>=2.3.2` and `>=4.0.4`. Prefer a real upstream bump; fall back to a narrow `overrides` entry in `frontend/package.json` if the parents pin. Note that #589 already uses `overrides` for `nwsapi`, so the mechanism is established here.
- [ ] `cd frontend && npm audit` reports **0 high**, and no vulnerable picomatch copy remains (`node_modules/picomatch` >= 2.3.2, `tinyglobby`'s >= 4.0.4).
- [ ] **Remove** the `GHSA-c2c7-rcm5-vvqj` line from `security-baseline.json` rather than leaving a fixed advisory ledgered — a stale baseline entry silently absorbs the *next* advisory on the same package.
- [ ] Verify the alerts close: alerts #112–#115 move to `fixed`.
- [ ] Regenerate the lockfile with CI's exact npm (11.19.0) and clean-room `npm ci` — local npm drops peers that CI keeps, and `npm ci --dry-run` passes regardless.
- [ ] Mutation-check the audit gate: with the baseline line removed and the bump reverted, `scripts/audit_gate.py` must **fail**. A gate that passes both ways is proving nothing.
- [ ] Frontend Tests, Backend Tests, E2E Playwright and Security Audit green.

## Notes

Surfaced while triaging the 2026-09-11 Dependabot batch (#586, #643, #644, #645, #646 — all merged/queued as Merge-now). #586 closed the one *dev*-scope minimatch alert; these four picomatch alerts were the residue, and unlike minimatch nothing upstream is filing for them.

Related: #522 (pip audit gate scans production-only — same class of gap on the Python side), #589 (existing `overrides` precedent).

frankbria:
Closes #647.

## What

Prunes **42 of 43** entries from `security-baseline.json`. All 41 npm entries and one pypi entry ledger advisories the tree no longer carries. Only `PYSEC-2026-2132` (click, fix: 8.3.3) is still live.

## Why a stale entry is worse than no entry

The baseline is an allowlist: `scripts/audit_gate.py` fails CI on any advisory **not** listed. An entry for an advisory that is already fixed suppresses nothing today — which is exactly why it is easy to leave in place, and exactly what makes it a trap. It stays armed to absorb the **next** occurrence of that advisory ID instead of failing the build.

`GHSA-c2c7-rcm5-vvqj` (picomatch) is the case that surfaced this. #648 patched every vulnerable copy in the tree; the baseline entry stayed behind. Had picomatch regressed — a lockfile revert, a resolution change, a transitive pulling an old copy back — CI would have gone green on a live high-severity ReDoS advisory.

## Evidence, measured not assumed

`npm audit` in `frontend/` (same invocation CI uses — from the lockfile, no `--omit=dev`, 1102 deps resolved):

```
{'info': 0, 'low': 0, 'moderate': 0, 'high': 0, 'critical': 0, 'total': 0}
```

`pip-audit` on the `--no-dev` export, as CI runs it: one live advisory, `PYSEC-2026-2132` (click). `PYSEC-2026-2270` (python-dotenv) is fixed.

The gate itself had been reporting all 42 under its own **"Resolved — prune these from the baseline"** heading. This PR acts on that output; it adds no new machinery. Notably, that reporting path exists because a #342 test anticipated precisely this day — it pins that `evaluate()` must take scanned ecosystems explicitly, since inferring them "would strand 53 dead entries on the happy day an ecosystem goes fully clean."

## Mutation verification

The point of this PR is that the gate still has teeth afterwards, so both directions were run rather than asserted:

| | Lockfile | Baseline | Gate | Meaning |
|---|---|---|---|---|
| **A** | picomatch reverted to 2.3.1 / 4.0.3 | **pruned (this PR)** | **exit 1**, names `GHSA-c2c7-rcm5-vvqj  picomatch  severity=high` | the gate catches the regression |
| **B** | same vulnerable lockfile | **old (main)** | **exit 0**, "no new advisories" | the stale entry *was* suppressing a live high advisory |

B is the one that matters: it is a working demonstration of the trap, not a hypothesis about it. Same tree, same advisory, opposite verdicts — the only variable is the line this PR deletes.

The mutation was reverted with `git checkout HEAD --` and the revert asserted with `git diff --quiet HEAD --` (a stale `.git/index.lock` silently no-ops checkouts and lets mutations accumulate unnoticed; checked for one first — none present). Working tree verified clean afterwards.

## Also answered: does the gate see moderate advisories?

#647 flagged that `GHSA-3v7f-55p6-f55p` (moderate, picomatch) was never baselined yet the gate passed while it was live — raising whether the gate silently skips moderates.

It does, **by design**: `--min-severity` defaults to `high` and the workflow does not override it. That is documented in the script and deliberate, not a hole — so no follow-up issue. Worth knowing explicitly, though: **a moderate advisory will never fail this gate**, whatever the baseline says.

## Not in scope

- **click `PYSEC-2026-2132` is retained and still live** — a fix exists (8.3.3). Bumping it is a backend dependency change with its own test surface, deliberately not widened into a baseline-hygiene PR.
- `pip-audit` runs `--no-dev`, so advisories in test/dev extras never reach the gate. Pre-existing, tracked in #522.

## Verification

- `python scripts/audit_gate.py --pip-audit … --npm-audit … --baseline security-baseline.json` → `PASS — no new advisories (1 known, already in the baseline)`, and the "Resolved" section is now empty
- `uvx --with pytest --with pyyaml pytest scripts/ -q` → **77 passed, 4 skipped** (the gate's own 29-test suite runs inside the audit job, so the guard is itself guarded)
- Both mutations above
- Required checks on CI

### DIFF
--- a/security-baseline.json
+++ b/security-baseline.json
@@ -1,50 +1,7 @@
 {
-  "_comment": "Known-and-accepted advisories as of #342. CI fails on anything NOT listed here, so this file is the repo's dependency-debt ledger, not an ignore-list to grow casually. Shrink it: each entry names its package and fix version. Regenerate wholesale with: python scripts/audit_gate.py --pip-audit <json> --npm-audit <json> --baseline security-baseline.json --write-baseline",
+  "_comment": "Known-and-accepted advisories. CI fails on anything NOT listed here, so this file is the repo's dependency-debt ledger, not an ignore-list to grow casually. Shrink it: each entry names its package and fix version. Started at 91 entries (#342); pruned to 1 in #647 once the tree caught up — a fixed advisory left ledgered here silently absorbs the NEXT advisory on the same package, so prune whatever the gate reports under 'Resolved'. Regenerate wholesale with: python scripts/audit_gate.py --pip-audit <json> --npm-audit <json> --baseline security-baseline.json --write-baseline",
   "pypi": {
-    "PYSEC-2026-2132": "click; fix: 8.3.3",
-    "PYSEC-2026-2270": "python-dotenv; fix: 1.2.2"
+    "PYSEC-2026-2132": "click; fix: 8.3.3"
   },
-  "npm": {
-    "GHSA-23c5-xmqv-rm74": "minimatch; fix: yes",
-    "GHSA-25h7-pfq9-p65f": "flatted; fix: yes",
-    "GHSA-267c-6grr-h53f": "next; fix: yes",
-    "GHSA-26hh-7cqf-hhc6": "next; fix: yes",
-    "GHSA-2w6w-674q-4c4q": "handlebars; fix: yes",
-    "GHSA-36qx-fr4f-26g5": "next; fix: yes",
-    "GHSA-392p-2q2v-4372": "better-auth; fix: yes",
-    "GHSA-3mfm-83xf-c92r": "handlebars; fix: yes",
-    "GHSA-3ppc-4f35-3m26": "minimatch; fix: yes",
-    "GHSA-492v-c6pp-mqqv": "next; fix: yes",
-    "GHSA-6g55-p6wh-862q": "postcss; fix: yes",
-    "GHSA-737v-mqg7-c878": "defu; fix: yes",
-    "GHSA-7r86-cg39-jmmj": "minimatch; fix: yes",
-    "GHSA-7w99-5wm4-3g79": "better-auth; fix: yes",
-    "GHSA-86j7-9j95-vpqj": "better-auth; fix: yes",
-    "GHSA-89xv-2m56-2m9x": "next; fix: yes",
-    "GHSA-8cpq-38p9-67gx": "kysely; fix: yes",
-    "GHSA-8h8q-6873-q5fj": "next; fix: yes",
-    "GHSA-96hv-2xvq-fx4p": "ws; fix: yes",
-    "GHSA-9cx6-37pm-9jff": "handlebars; fix: yes",
-    "GHSA-9h47-pqcx-hjr4": "better-auth; fix: yes",
-    "GHSA-c2c7-rcm5-vvqj": "picomatch; fix: yes",
-    "GHSA-c4j6-fc7j-m34r": "next; fix: yes",
-    "GHSA-f88m-g3jw-g9cj": "sharp; fix: yes",
-    "GHSA-fmh4-wcc4-5jm3": "better-auth; fix: yes",
-    "GHSA-g38m-r43w-p2q7": "better-auth; fix: yes",
-    "GHSA-h25m-26qc-wcjf": "next; fix: yes",
-    "GHSA-hmw2-7cc7-3qxx": "form-data; fix: yes",
-    "GHSA-m99w-x7hq-7vfj": "next; fix: yes",
-    "GHSA-mg66-mrh9-m8jx": "next; fix: yes",
-    "GHSA-p6v2-xcpg-h6xw": "better-auth; fix: yes",
-    "GHSA-p9j2-gv94-2wf4": "next; fix: yes",
-    "GHSA-pv5w-4p9q-p3v2": "kysely; fix: yes",
-    "GHSA-pw9m-5jxm-xr6h": "better-auth; fix: yes",
-    "GHSA-q4gf-8mx6-v5v3": "next; fix: yes",
-    "GHSA-qq9h-g4jm-xgf3": "better-auth; fix: yes",
-    "GHSA-r28c-9q8g-f849": "postcss; fix: yes",
-    "GHSA-rf6f-7fwh-wjgh": "flatted; fix: yes",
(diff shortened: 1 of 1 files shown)


frankbria:
## Third-party review

**opencode/GLM (primary): no output.** Ran `opencode run --model zai-coding-plan/glm-4.6` with the diff inline; it produced a zero-byte result and had to be killed — the documented silent-hang failure mode. One attempt, then fell back as usual.

> The diff only prunes resolved entries from the security baseline while retaining the remaining known Click advisory. I did not identify any actionable regression in the changed file.

It read `scripts/audit_gate.py` and traced the baseline's consumers before concluding.

### Limit of that review, stated plainly

codex tried to run `npm audit` itself and **could not** — its sandbox has no DNS (`getaddrinfo EAI_AGAIN registry.npmjs.org`). So its clearance rests on reading the gate and the diff, **not** on independently reproducing the "these 42 advisories are gone" claim. That claim is load-bearing for this PR, and the evidence for it is mine: the audit output and the two mutations in the description, plus CI, which runs both audit tools with network access and will fail this PR if the tree still carries an unbaselined advisory.

In other words: CI is the real check that the prune is safe, and it is exercising exactly the code path in question.

### One thing neither reviewer could confirm

Whether emptying `"npm"` to `{}` behaves identically to removing the key. It does — `load_baseline` is `{eco: data[eco] for eco in ECOSYSTEMS if eco in data}`, so an empty dict is a present-but-empty section, and `evaluate` treats every npm finding as unbaselined. Mutation A demonstrates this end-to-end: with `"npm": {}`, a reintroduced picomatch advisory exits 1 rather than being silently accepted.

Issue #642, opened 2026-09-10
[P3.42] Disabled-for-busy buttons render their live status text at 2.29:1 — disabled:opacity-50 applies to every Button


frankbria:
Found while implementing #635 (PR #641), and deliberately left out of it: it is
not a spinner bug, it applies to every `<Button>` in the app.

## The measurement

`button.tsx:10` puts `disabled:opacity-50` in the shared base string. CSS
`opacity` groups the element, so the background and the label are *each* blended
50% with the page behind the button — they do not composite against each other.
For the default variant (`bg-primary text-primary-foreground` = brand indigo /
white):

| | enabled | disabled |
|---|---|---|
| light, on `--background` or `--card` | 6.29 | **2.29** |
| dark, on `--background` | 6.29 | 3.37 |
| dark, on `--card` | 6.29 | 3.33 |

Measured with `src/__tests__/theme/helpers/contrast.ts` against this repo's
`globals.css` and `tailwind.config.js`. (The first model I used composited the
label against its own tinted background and gave 1.56 — wrong. `opacity` on a
parent flattens the group against what is *behind* the group.)

## Why this is worth a ticket and not a shrug

**It is not a WCAG failure.** SC 1.4.3 exempts "text that is part of an inactive
user interface component", and 1.4.11 exempts inactive components too. A greyed
button that is simply unavailable is fine at 2.29:1 — that is the whole point of
the affordance.

The problem is that this codebase uses `disabled` for **busy**, not only for
unavailable. Every one of these buttons sets `disabled={isLoading}` and then
renders live status text inside itself:

- `QuestionGenerator.tsx:213-220` — "Generating Questions..." + a `◌` glyph
- `RegenerateQuestionsDialog.tsx:108-111` — "Regenerating..."
- `BookCreationWizard.tsx:253-260` — "Creating..."
- `DeleteBookModal.tsx:189-197` — "Deleting..."
- `ai-error-handling-example.tsx` ×3

That text is not incidental — it is the only feedback that the action is
running, and the user is meant to read it. At 2.29:1 in light theme it is the
hardest-to-read text on the screen at exactly the moment it matters most.

This is the same shape as #635's headline defect one layer up: there the
*spinner* inside a busy button was invisible; here the *label* is nearly so.

## Options (a real fork — needs a decision, not a default)

1. **`aria-busy` + a distinct busy style.** Keep `disabled:opacity-50` for
   genuinely-unavailable buttons; give the busy state its own treatment
[...]

Issue #635, opened 2026-09-10
[P2.34] Two more hand-rolled spinner rings are 2.33:1 in light mode; eleven sites, four ring colours, one existing icon idiom


frankbria:
Split out of #629 (PR #633) rather than widening it. #629 fixed the `ChapterTab`
loading spinner; the same hand-rolled `animate-spin` ring pattern appears at ten
more sites, and two of them fail the same way.

## Measured

Same maths as #629 — `src/__tests__/theme/helpers/contrast.ts` against the
`globals.css` tokens. WCAG 2.1 **1.4.11**, 3:1 floor for non-text UI. A spinner
ring is the whole indicator, so the ring colour *is* the component.

| ring | light `--background` | light `--muted` | dark `--background` | dark `--muted` |
|---|---|---|---|---|
| `border-gray-400` — `BookMetadataForm.tsx:176` | 2.54 ❌ | **2.33** ❌ | 7.80 ✅ | 5.96 ✅ |
| `border-blue-400` — `ClarifyingQuestions.tsx:169` | 2.54 ❌ | **2.33** ❌ | 7.79 ✅ | 5.95 ✅ |
| `border-indigo-500` — 4 full-page loaders | 4.47 ✅ | 4.10 ✅ | 4.43 ✅ | 3.39 ✅ |

`border-blue-400` at 2.33 is the **exact** value #629 fixed, at a second site the
sweep never reached.

The `border-indigo-500` full-page loaders (`app/page.tsx:25`,
`edit-toc/page.tsx:515`, `export/page.tsx:187`, `toc/ReadinessChecker.tsx:5`)
clear the floor everywhere and need no colour change — but see below.

### Not measured here

`border-white` rings inside buttons (`TocReview.tsx:119,139`,
`ClarifyingQuestions.tsx:280`) sit on a button's own surface, not on
`--background`/`--muted`, so they need the button's background resolved first.
Likely fine on a filled primary button; unverified.

## Why the #629 guard did not catch these

`status-dot-contrast.test.ts` reads named consts out of two chapter components.
These rings are inline literals in unrelated files, so nothing sweeps them. This
is the same shape as #620: the fix landed, the sweep did not follow.

## The larger point

Eleven hand-rolled spinners, four different ring colours, three different
markup shapes — while the app already has an icon spinner idiom
(`HugeiconsIcon` + `Loading03Icon`, used by `ChapterEditor`'s save indicator and,
[...]

PR #652, opened 2026-09-13
fix(a11y): four stranded pale accents under AA, and a guard the issue's predicate would have missed (#637)

### LINKED ISSUE #637: [P2.35] The stranded-accent half of #632 is unguarded: 27 unprefixed light `text-<hue>-{300,400}` sites
Split out of #632 (PR #636), which guarded one half of the theme-blind accent bug
and explicitly scoped out the other.

## The other half

#632 banned the **dark surface** with no `dark:` prefix — `bg-<hue>-900/NN`
authored for a dark page. The mirror defect is the **light foreground** with no
`dark:` prefix: `text-<hue>-{300,400}`, a pale accent that only makes sense on a
dark surface. Painted on a light one it is the same mud, reached from the other
direction, and #633 (P2.33) is the same shape again with the surface and glyph
swapped.

The two co-occur: every one of the 17 sites #632 converted was a
`bg-<hue>-900/NN` **paired with** a `text-<hue>-400`, and the #632 guard only
catches the surface. A file can now paint `dark:bg-red-900/20` correctly and
still strand a bare `text-red-400` on the light card, and both guards pass.

## What is left

A sweep of shipped `src/` for `(text|border|ring|fill|stroke)-<non-neutral
hue>-(200|300|400)` with no `dark:` in its variant chain finds **27** sites,
after #636 lands:

| count | file |
|---|---|
| 3 | `app/dashboard/books/[bookId]/summary/page.tsx` |
| 2 | `components/chapters/ChapterTab.tsx` |
| 2 | `components/toc/ErrorDisplay.tsx` |
| 2 | `app/error.tsx` |
| 2 | `app/dashboard/page.tsx` |
| 1 each | `app/global-error.tsx`, `app/layout.tsx`, `app/dashboard/books/[bookId]/page.tsx`, `app/dashboard/books/[bookId]/edit-toc/page.tsx`, `app/dashboard/books/[bookId]/chapters/[chapterId]/page.tsx`, `app/dashboard/books/[bookId]/export/page.tsx`, `components/chapters/VoiceTextInput.tsx`, `components/chapters/questions/QuestionProgress.tsx`, `components/toc/ClarifyingQuestions.tsx`, `components/toc/NotReadyMessage.tsx` |

**Not all 27 are failures, and that is the work.** A bare `border-red-300` is a
*light*-authored border — wrong in dark, not in light, so it is #633's bug rather
than this one. A bare `text-red-400` on a `bg-red-50` card is a live light-mode
failure today. They need separating by the surface each actually sits on before
anything is converted, which is why #632 scoped this out rather than widening
its net.

## Why static, again

Same reason as #632 and #623: several of these are icons and rings, and axe's
`color-contrast` rule does not evaluate non-text elements at all. A green axe run
is not evidence here.

## Acceptance criteria

- [ ] Each of the 27 classified against the surface it renders on: live light-mode
      failure, #633-shaped dark-mode failure, or intentionally theme-fixed
- [ ] Live failures converted to `text-<hue>-700 dark:text-<hue>-400` pairs,
      measured arithmetically in both themes on every surface they land on
- [ ] Guard extends `src/__tests__/theme/dark-translucent-overlay.test.ts` (or
      sits beside it sharing `helpers/sources.ts`), with the same per-value
      ledger + stale-row shape and a fixture-based anti-vacuity test
- [ ] Mutation-checked against an unledgered and a ledgered file
- [ ] No fix that regresses the theme it was not aimed at

frankbria:
Closes #637.

## The issue's predicate could not be used, and that is the main finding

#637 asks for a sweep of `(text|border|ring|fill|stroke)-<hue>-(200|300|400)` **"with no `dark:` in its variant chain"**. That flags the light half of every correct `border-red-300 dark:border-red-700` pair — the idiom #631 established and #632 measured.

On this tree it returns **28 sites, of which 16 are that exact correct pair**. The four real defects would have arrived buried in four times their number of false positives, and "fixing" them would have meant unpicking #631.

The guard tests for being **stranded** instead: no `dark:`-prefixed use of the *same property and same hue* on the line. That finds **12**, no false positives. It also skips comment lines — `ChapterTab.tsx` quotes `border-blue-400` inside a #629 explanatory comment, and matching prose would have put a ledger row on a file whose markup is clean.

## Four live light-mode failures, measured on the surface each renders on

| site | before | after |
|---|---|---|
| `summary` input error — `text-red-400` on white | **2.77:1** | 6.47 |
| `summary` revert link — `text-indigo-400` on white | **2.98:1** | 7.90 |
| `edit-toc` add-subchapter hover — `text-indigo-400` | **2.98:1** | 7.90 |
| `dashboard` header icon (32px) — `text-indigo-400` | **2.98:1** | 7.90 |

The last is a **non-text graphic**, so 1.4.11's 3:1 applies and it missed by **0.02** — the kind of margin no eyeball review catches, and axe reports nothing because its `color-contrast` rule does not evaluate non-text elements at all (the #623/#632 blind spot).

All four take #631's `-700 dark:-400` pair.

## Eight ledgered, each with the measurement that justifies it

- **`error.tsx` (×2), `global-error.tsx`, `layout.tsx`** — theme-fixed dark error boundaries. They render when the app has already failed, so they deliberately do not depend on the theme system; `text-red-400` on their literal `bg-gray-900`/`bg-gray-950` is **6.41** / **7.28**, and a `dark:` pair would be dead code since there is no light surface to swap to.
- **3 × `disabled:` button labels** — indigo-400 on indigo-800 = **3.33**, green-400 on green-800 = **4.09** (×2). Exempt under **WCAG 1.4.3's inactive-component clause**. Whether disabled states are legible enough in practice is #642, deliberately not relitigated here.
- **`VoiceTextInput` success toast** — a light-only card, correct in light (**6.81**) and wrong in dark (**2.78**). That is **#633's shape**, not this one, and needs the whole card rather than its border.

## Mutation M1 found a live hole in the first cut of this guard

Injecting a stranded `text-orange-400` into a row that already carried `dark:text-red-400` **passed**. Pairing was checked per-property, so *any* `dark:text-` on the line counted as the counterpart.

That is not an edge case: a pale accent added to an alert row already full of correct pairs is the commonest way this defect arrives, so the guard was quietly useless exactly where it mattered most. Pairing is now property **and** hue, with M1's shape pinned as a fixture.

Six mutations, all RED, tree verified clean after each:

| | mutation | fails on |
|---|---|---|
| M1 | stranded accent in an unledgered file | `outside the ledger` |
| M2 | ledgered file above its recorded count | `recorded count` |
| M3 | one of the four fixes reverted | `outside the ledger` |
| M4 | the sweep pattern broken | the anti-vacuity fixture |
| M5 | a burned-down row left stale | `no longer contains` |
| M6 | net-zero hue swap inside a ledgered file | `recorded count` |

M6 is the #620 per-value lesson: a per-file total would let a hue swap net out.

## `npm run typecheck` never saw this guard

`tsconfig` excludes `**/__tests__/**` (#625), so the repo's own gate is blind to every test file. Running `tsc` directly on it — the procedure that issue documents — surfaced **two real `TS2345` arity errors** (`it.each` tuples carrying two elements against one-parameter callbacks) that would otherwise have shipped.

## Verification

- New guard: **15/15**
- All theme guards: **303/303, 8 suites** — no regression in the theme this was not aimed at
- Full frontend suite: **144 suites, 2628 passed, 4 skipped**
- Coverage 89.44 / 80.51 / 86.55 / 90.63 (stmts/branch/funcs/lines) — over the 85/85/75/85 floors
- `npm run lint` 0 errors; `npm run typecheck` clean; `tsc` directly on the guard clean
- Sweep drops 12 stranded → 8, exactly the ledgered set

## Known limitations

- The 16 correctly-paired sites are untouched and unguarded — this guard only catches *stranded* accents, not a pair whose two halves are both wrong. That is a different sweep.
- `VoiceTextInput`'s toast is ledgered, not fixed; its dark-mode failure is real and belongs to #633.


frankbria:
## Third-party review

**opencode/GLM (primary): no output, twice.** Two bounded attempts (200s and earlier in the same session), both returned zero bytes and had to be killed — the documented silent-hang mode. Not a clean bill of health; it produced no verdict at all.

**codex (fallback): one [P2] finding, and it was right.**

> For utilities with state variants, this treats any same-property/hue `dark:` class in the run as a valid counterpart even when it applies in a different state. For example, `className="hover:text-indigo-400 dark:focus:text-indigo-700"` or `className="text-red-400 dark:hover:text-red-700"` is reported clean, but the pale unprefixed color still applies without a matching dark replacement in its actual state.

**Reproduced before fixing** (this repo has been burned by a finding whose reasoning was right but whose scenario did not reproduce). Both of codex's exact strings returned `[]` from the shipped predicate — confirmed false greens.

It also lands harder than stated: **one of this PR's own four fixes was a hover-only defect** (`edit-toc` add-subchapter). A state-blind guard would have failed to catch the very shape it was built for.

Fixed in `a1e37ee`: a counterpart must carry the same **non-dark variant set**. `hover:` pairs with `dark:hover:` or `hover:dark:`, resting pairs with resting, and a `focus:` dark class no longer rescues a `hover:` light one. Both scenarios are pinned as fixtures, along with the two matching-state controls.

### Three false greens in this guard, and how each was found

Worth recording together, because only one came from a reviewer:

| | defect | found by |
|---|---|---|
| 1 | pairing per **property** only — a `dark:text-red-400` rescued a stranded `text-orange-400` | **mutation M1** |
| 2 | pairing per **source line** — two elements on one line pooled their utilities | probing a question opencode's prompt raised before it hung |
| 3 | pairing ignored **state** — `dark:hover:` rescued a resting-state colour | **codex** |

All three are the same underlying error — asking "is there *a* dark class nearby?" instead of "is there a dark class that replaces *this* one, where it paints". The guard now asks the second question along all three axes. Defect 2 is the utility-pooling false green #635's guard shipped and its review caught, which is the second time this repo has hit it.

### Verification after the fix

- Guard **15/15**; all theme guards **303/303** (8 suites)
- `tsc` directly on the guard: clean (`npm run typecheck` excludes `**/__tests__/**`, #625)
- **8 mutations, all RED**: unledgered file gains one, ledgered file over count, a fix reverted, pattern broken, stale row, net-zero hue swap, utility pooling — plus the clean baseline. Tree verified clean after each.
- No change to the real tree from any of the three tightenings, i.e. every one was closing a hole rather than reclassifying a site.

frankbria:

Every acceptance criterion mapped to outcome evidence, not "it rendered".

### AC1 — each site classified against the surface it renders on

The issue's list was 27 across 16 files. The sweep found **30 across 20** on the pre-#641 tree (it missed `app/page.tsx`, `ChapterTabs.tsx`, `ErrorNotification.tsx`, and undercounted `ChapterTab.tsx` and `ClarifyingQuestions.tsx`) — the #620 lesson a fourth time: the sweep finds what the hand-written list forgets.

Of those, **16 are the correct `#631` light/dark pair** and were never defects. Post-#641 (which converted one spinner to the icon idiom), the stranded set is **12**, classified:

| n | class | verdict | evidence |
|---|---|---|---|
| 4 | live light-mode failure | **fixed** | 2.77–2.98:1, below AA |
| 4 | theme-fixed dark error boundary | ledgered | 6.41 / 7.28:1 — already compliant |
| 3 | `disabled:` label | ledgered | WCAG 1.4.3 inactive-component exemption |
| 1 | light-only toast | ledgered | 6.81 light / **2.78 dark** → #633's shape |

### AC2 — conversions measured arithmetically in both themes

Derived from `globals.css` tokens and the Tailwind v3 palette, not copied:

| site | before | after (light) | after (dark) |
|---|---|---|---|
| summary input error | 2.77 | **6.47** | 4.5+ |
| summary revert link | 2.98 | **7.90** | 4.5+ |
| edit-toc hover | 2.98 | **7.90** | 4.5+ |
| dashboard icon (32px, 1.4.11 → 3:1) | **2.98** | **7.90** | 4.5+ |

**The colours were confirmed to actually render** — the #634 trap, where `text-primary` was priced as the `--primary` token when it resolves to a brand override. Checked here: `tailwind.config.js` overrides only `primary`/`secondary`/`destructive`/neutral families; `red` and `indigo` are untouched, and `globals.css` has no `@layer utilities` rule for any of these classes. So `text-red-700` → `#b91c1c` and `text-indigo-700` → `#4338ca` straight from the palette, which is what was measured.

### AC3 — guard shape

`src/__tests__/theme/stranded-light-accent.test.ts`, beside `dark-translucent-overlay.test.ts` and sharing `helpers/sources.ts` and `helpers/contrast.ts`. Per-distinct-utility ledger (`stranded-accent-baseline.json`), #624 stale-row assertion, fixture-based anti-vacuity test, and the self-exclusion check (`sources` must not contain the guard itself — the #620 self-match trap).

### AC4 — mutation-checked

Required: "an unledgered and a ledgered file". Delivered **7 mutations + clean baseline, all RED**:

```
baseline   15/15 clean                                    PASS
M1  stranded accent in an UNLEDGERED file                 -> outside the ledger
M2  LEDGERED file above its recorded count                -> recorded count
M3  one of the four fixes reverted                        -> outside the ledger
M4  sweep pattern broken                                  -> anti-vacuity fixture
M5  burned-down row left stale                            -> no longer contains
M6  net-zero hue swap inside a ledgered file              -> recorded count
M7  utility pooling: 2 elements, 1 line, same prop+hue    -> outside the ledger
mutations passed: 8   failed: 0        tree clean
```

Each revert asserted with `git diff --quiet HEAD --`, with a stale-`index.lock` check first — a leftover lock silently no-ops `git checkout` and lets mutations accumulate unnoticed. (One was in fact present earlier in this session and caught by exactly that assertion.)

**M1 and M7 were not confirmations — they found live holes** and are the reason the predicate is now property + hue + state + class-run scoped rather than the one-axis check it started as.

### AC5 — no regression in the theme this was not aimed at

- All 8 theme guards: **303/303**
- Full frontend suite: **144 suites, 2628 passed, 4 skipped**
- Coverage 89.44 / 80.51 / 86.55 / 90.63 vs floors 85 / 85 / 75 / 85
- `npm run lint` 0 errors; `npm run typecheck` clean; `tsc` run directly on the guard clean (the repo gate excludes test files, #625)
- Sweep: **12 stranded → 8**, exactly the ledgered set — no site silently reclassified

### What this demo does not show

No browser screenshots. This change is static analysis plus four class swaps whose colours are read from source; the contrast maths here is the same helper #632 validated against real axe in Chromium (it reported `#210e0e` / 6.69 where the helper computes 6.70). A render would confirm the classes apply, which is instead confirmed above by checking the override paths directly — the failure mode #634 actually hit.

PR #653, opened 2026-09-13
fix(api): merge TOC updates instead of replacing, and derive total_chapters (#496)

### LINKED ISSUE #496: [P3.23] TOC update replaces table_of_contents wholesale, so a partial save resets total_chapters and other metadata
Surfaced by the third-party review on #495 (fix for #492). **Latent — no current caller triggers it.**

## The behavior

`_update_toc_internal` (`backend/app/db/toc_transactions.py:155`) writes the whole document:

```python
updated_toc = {
    **toc_data,
    "generated_at": current_toc.get("generated_at"),
    ...
}
```

Anything the caller omits is not merged from the stored TOC — it is dropped. So a client that sends only `{"toc": {"chapters": [...]}}` (which is a valid body, and is exactly what the E2E helpers send) silently clears `total_chapters`, `estimated_pages`, and `structure_notes`.

`GET /books/{id}/toc` then defaults `total_chapters` to `0` (`backend/app/api/endpoints/books.py:1326`) while returning real chapters — an internally inconsistent response.

This is the same root pattern as #492 (a missing key resolving to a silently destructive default), applied to the metadata fields rather than to `chapters`.

## Why it is not firing today

The only production caller, `bookClient.updateToc` (`frontend/src/lib/api/bookClient.ts:812`), types all four fields as required and resends the full object. #495 closed the chapters-loss path. The next partial-update caller — a mobile client, a script, a refactor — would hit this.

## Suggested fix

Either:
- merge over the stored TOC (`{**current_toc, **toc_data, ...}`), so omitted fields are preserved; **or**
- derive `total_chapters` from `len(chapters)` server-side rather than trusting the client, which also removes a class of client/server disagreement.

Add a test: PUT `{"toc": {"chapters": [...]}}` after a full save must not zero out `total_chapters` on read-back.

## Priority

P3.23 — real correctness bug in a save path, but unreachable from any shipped client, so below the live dashboard-performance work in #493.

frankbria:
Closes #496.

## The bug

`_update_toc_internal` built the new document as `{**toc_data, ...}`, so every key the caller omitted was **dropped rather than carried over**. `{"toc": {"chapters": [...]}}` is a valid body — and exactly what the E2E helpers send — and it silently cleared `total_chapters`, `estimated_pages` and `structure_notes`. `GET /toc` then answered `total_chapters: 0` alongside real chapters.

Same root pattern as #492: a missing key resolving to a destructive default, here on the metadata rather than on `chapters`. Latent when filed — the only shipped caller types all four fields as required — so this is a trap for the next partial-update caller, not a live failure.

## Both remedies taken, because they cover different halves

The issue offered merge **or** derive. Both are here:

1. **Merge over the stored TOC** — omitted fields survive.
2. **Derive `total_chapters` from the chapter list** — not redundant with (1). Every other producer in the codebase already computes it as `len(chapters)`: `ai_service.py:713`, `export_service.py:959`, the flat chapter listing at `chapters.py:473`, and the frontend's own `edit-toc/page.tsx:137`. Accepting a client value only ever created a way for the two to disagree; `chapters.py:482`'s `current_toc.get("total_chapters", len(chapters))` fallback now always agrees with its own default.

**Merging cannot resurrect deleted chapters.** The endpoint still rejects a body without `chapters` (#492/#495), so `chapters` is effectively required; an intentional clear is `{"chapters": []}`, which is present and wins the merge.

## A third defect, found while testing the first two

The handler returned `"toc": toc_data` — the client's own request body. So a partial update answered **200 describing the partial object** while the database held the merged one, and a client trusting that response disagreed with its own next `GET`. That is the same inconsistency the issue complains about, one layer up.

The response now reports what was stored, in the same shape `GET /toc` returns. Called out because it is beyond the issue's literal scope.

## Mutation-checked, one per defect

Each fix reverted independently, so no test is standing in for another:

| mutation | fails |
|---|---|
| un-merge → `{**toc_data, ...}` | only `test_partial_update_preserves_omitted_metadata` |
| un-derive `total_chapters` | only `test_total_chapters_is_derived_not_trusted` |
| restore the `toc_data` echo | only `test_put_response_agrees_with_subsequent_get` |

Every revert asserted with `git diff --quiet HEAD --`; tree verified clean; all three green again afterwards.

## Verification

- New tests: 3, each RED before the fix and GREEN after
- `test_books_toc_coverage.py` 32 passed, `test_toc_transactions.py` 55 passed — 87 with no regression
- Full backend suite: **1259 passed, 9 skipped, 92% coverage**
- `ruff` clean; pre-commit backend lint / tests / coverage all passed

## Known limitations

- `total_chapters` now counts **top-level** chapters. That matches every producer and the stored-TOC semantics, but note `chapters.py:473` reports a *flattened* count (including subchapters) under `"structure": "flat"` — a different response shape, computed from its own list, deliberately untouched here.
- A client that deliberately wanted to blank `structure_notes` must now send `""` explicitly rather than omitting the key. That is the intended trade: omission meaning "erase" is the bug.


frankbria:
## Third-party review

**opencode/GLM (primary): skipped, with reason.** It was attempted twice earlier in this session — on a 49-line diff and a 448-line diff — and returned **zero bytes both times**, having to be killed (the documented silent-hang mode). Rather than burn a third timeout on the same known failure, this went straight to the fallback. Stating it plainly because "skipped" is not "clean": opencode has produced no verdict on this change.

> The changes correctly merge partial TOC updates, derive total_chapters from stored chapters, and align the PUT response with persisted data. I did not find any introduced correctness, security, or reliability issues that warrant inline findings.

Acceptance criteria mapped to outcome evidence.

**AC — "PUT `{"toc": {"chapters": [...]}}` after a full save must not zero out `total_chapters` on read-back"** (the issue's literal ask):

`test_partial_update_preserves_omitted_metadata` does exactly this — full save with `total_chapters: 2`, `estimated_pages: 180`, `structure_notes: "Two-act structure."`, then a partial PUT carrying only `chapters`, then `GET`. Asserts `estimated_pages == 180` and `structure_notes == "Two-act structure."` survive, and chapters update to the new three. **RED before the fix, GREEN after**, with the un-merge mutation confirming it is this change that makes it pass.

**AC — the fix itself**: both remedies the issue offered, for the reason given in the PR body (they cover different halves).

**Outcome evidence per defect, not "the tests pass":**

```
un-merge   {**current_toc, **toc_data} -> {**toc_data}
           => FAILS only test_partial_update_preserves_omitted_metadata
un-derive  updated_toc["total_chapters"] = len(...)  removed
           => FAILS only test_total_chapters_is_derived_not_trusted
un-fix     response "toc" back to toc_data
           => FAILS only test_put_response_agrees_with_subsequent_get
```

Three mutations, three distinct failures, no overlap — so no test is standing in for another, and each fix is independently load-bearing. Every revert asserted with `git diff --quiet HEAD --` (a stale `.git/index.lock` silently no-ops `git checkout`, and one was in fact present twice during this session), tree verified clean, all three green again afterwards.

**No regression:**
- `test_books_toc_coverage.py` 32 passed · `test_toc_transactions.py` 55 passed — **87 together**
- Full backend suite **1259 passed / 9 skipped, 92% coverage**
- `ruff` clean; pre-commit backend lint + tests + coverage passed

**What this demo does not show:** no UI walkthrough. The defect is unreachable from any shipped client — `bookClient.updateToc` types all four fields as required and resends the whole object — so there is no user-facing flow that reproduces it. The API-level tests are the reproduction, which is why the mutation matrix above carries the weight rather than a screenshot.

Issue #561, opened 2026-08-28
[P3.28] Backend lint hook is vacuous twice over: `ruff` is not installed and the hook ends in `|| true`


frankbria:
## What

`.pre-commit-config.yaml:86`:

```yaml
- id: backend-lint
  name: Backend Linting (ruff)
  entry: bash -c 'cd backend && uv run ruff check app/ tests/ || true'
```

Two independent reasons this can never fail, and one that means it never even runs:

1. **`|| true`** — the hook reports Passed regardless of ruff's exit code.
2. **`ruff` is not a backend dependency.** `uv run ruff` errors with `Failed to spawn: ruff / No such file or directory`; it is in neither `pyproject.toml`'s main deps nor the `test` extra. So the command fails immediately, and `|| true` swallows that too.

Net effect: every commit prints `Backend Linting (ruff)...Passed` having linted nothing. Found while landing #559 — `uvx ruff` on the new files reported a real finding (`BLE001`) that the hook had already declared clean.

`black` is in the same position: `CLAUDE.md` names it as the backend formatter, but nothing in pre-commit or CI runs it, and the backend is not black-clean at baseline (`uvx black --check app/` → 37 of 56 files would be reformatted).

## Why P3

No correctness risk — CI's real gates (tests + coverage, both enforced and branch-protected since #118) are unaffected, and this is style only. It sits with the other "a gate reports green while doing nothing" items, #519 and #547.

## Suggested shape

1. Add `ruff` to the backend `test` extra so `uv run ruff` resolves.
2. Drop `|| true`.
3. Expect the first honest run to be red. Decide then whether to fix the findings or land a baseline config — do **not** re-add `|| true`.
4. Decide black's status explicitly: either wire it up and reformat the backend in one mechanical commit, or drop the claim from `CLAUDE.md`. A named-but-unenforced formatter is worse than neither, because it makes every diff a coin flip on whose formatting wins.
5. Leave one mutation-checked tripwire, per the #534 lesson: after fixing, break something on purpose and confirm the hook goes red.

## Related

- #559 — where this surfaced
- #519, #547 — checks that fail or no-op while reading as complete

PR #654, opened 2026-09-13
fix(ci): audit the full dependency surface, extras included (#522)

### LINKED ISSUE #522: [P3.27] The pip audit gate scans production-only, so advisories in test/dev extras never fail CI — GHSA-6w46-j5rx-g56g is open and unbaselined

frankbria:
Closes #522.

## Decision: widen, not document-only

The issue offered two honest options — audit `--all-extras`, or keep production-only and say so. **Widening**, for a reason the issue itself supplies: the neighbouring lockfile-sync step *already* exports `--all-extras`. The two steps scanning different sets was not a considered trade-off, it was an accident. And a silent blind spot is precisely what the baselined design (#342) exists to prevent — an advisory outside the scanned surface isn't deferred with a reason, it's simply unseen, so GitHub's alert list and CI disagree with no record of why.

**Widening cost nothing.** Export goes 478 → 977 lines; the only advisory on the wider surface is click's `PYSEC-2026-2132`, already ledgered. No new baseline entries.

Documented **as well as** widened — the AC says "either/or", but the doc half is cheap and the gate's real surface was only discoverable by reading the workflow.

## The issue's driving instance is stale

`GHSA-6w46-j5rx-g56g` affects pytest `< 9.0.3`. The issue quotes `pytest==8.4.1`; the pin is now **9.1.1**, and both alert rows read `fixed`:

```
$ gh api .../dependabot/alerts  | filter GHSA-6w46-j5rx-g56g
[{"state":"fixed","pkg":"pytest","patched":"9.0.3"}, {"state":"fixed",...}]
```

Checked against the live alert list rather than taken from the issue body. So the second AC — "patched or recorded with a reason" — is satisfied by **already patched**, and nothing needs baselining.

## Two limits now written down instead of discovered

`docs/references/quality-standards.md` gains a table of what each scanner actually covers, plus:

- **npm advisories below `high` never fail this gate.** `audit_gate.py` defaults `--min-severity` to `high` and the workflow does not override it. A moderate is not baselined, not reported, and not a failure — by design, but invisible unless you go looking. (This is the same question #637 raised about the theme guards and it deserved a written answer.)
- **Only `frontend/package-lock.json` is scanned.** A lockfile elsewhere is invisible to this gate *and* to Dependabot version updates, which is why a separate step fails the build when one appears.

## Mutation-checked

| | mutation | result |
|---|---|---|
| M1 | synthesised pytest (test-extra) advisory fed to the real gate | **exit 1**, names `GHSA-6w46-j5rx-g56g` — the widened surface is genuinely read, not merely exported |
| M2 | the true wider surface | **exit 0** — no hidden backlog was being masked by the narrow scan |
| M3 | baseline's `pypi` section emptied | **exit 1** on click — the gate is not vacuous on this input |

M1 is the one that matters: widening the export proves nothing on its own if the gate doesn't act on what it finds.

## Incidental finding

While measuring the two exports: **`ruff` appears in neither** the production export nor `--all-extras`. That independently confirms the premise of **#561** ("`ruff` is not installed and the hook ends in `|| true`") — the backend lint hook is invoking a binary the dependency set does not provide.

## Verification

- `tests.yml` parses as valid YAML; the `Run pip-audit` step exports `--all-extras`
- Three mutations above
- No baseline change required
- Required checks on this PR


frankbria:
## Third-party review

**opencode/GLM (primary): skipped.** Two attempts earlier this session (49-line and 448-line diffs) both returned zero bytes and had to be killed — the documented silent-hang mode. Skipped rather than burning a third timeout on a known failure. Stating it because "skipped" is not "clean": opencode has produced no verdict here.

> The workflow change consistently widens the PyPI audit export to include optional extras, and the accompanying documentation reflects that behavior. I did not identify any discrete introduced defects that would break existing behavior or CI.

### Limit of that review

codex **tried to run the export itself and could not** — its sandbox mounts `/tmp` read-only, so both `uv export` attempts failed (`Read-only file system (os error 30)`). Its clearance therefore rests on reading the diff, not on reproducing the 478 → 977 claim or the advisory count.

That claim is load-bearing here, so the evidence for it is mine (measured locally, shown in the PR body) plus CI, which runs the real `uv export --all-extras` with a writable filesystem and network and will fail this PR if the wider surface carries an unbaselined advisory. **CI is the actual check that widening was free.**

Same pattern as #651 earlier tonight, where codex's sandbox had no DNS and couldn't run `npm audit`. Worth noting as a standing limitation rather than rediscovering it each time: for any change whose correctness depends on running a package-manager command, codex reviews the reasoning and CI proves the fact.

**AC1 — "the audited surface is either widened to the extras or documented as production-only in both the workflow comment and quality-standards.md"**

Both, not either. The step now exports `--all-extras`; the workflow comment explains what changed and why; `quality-standards.md` gains a per-ecosystem surface table plus the two previously-undocumented limits (npm sub-`high` never fails; only `frontend/package-lock.json` is scanned).

Outcome evidence, not "it parses":

```
$ python3 -c "yaml.safe_load(open('.github/workflows/tests.yml'))"
  valid YAML; jobs: [frontend-tests, backend-tests, security-audit, e2e-tests, quality-summary]
  pip-audit export line: uv export --all-extras --no-emit-project --format requirements-txt > /tmp/reqs.txt

$ surface coverage (prod export vs all-extras export)
  pytest          prod=0  all-extras=1
  pytest-asyncio  prod=0  all-extras=1
  locust          prod=0  all-extras=1
  httpx           prod=0  all-extras=1
```

The test extras were genuinely absent before and are genuinely present now — which is the claim, rather than "the flag changed".

**AC2 — "`GHSA-6w46-j5rx-g56g` is either patched or recorded in `security-baseline.json` with a reason"**

**Already patched.** Verified against the live alert list, not the issue body: the advisory affects pytest `< 9.0.3`, the pin is now `9.1.1`, and both alert rows read `state: fixed`. Nothing to baseline.

**The gate acts on what it now sees** — the part widening alone would not prove:

```
M1  synthesised pytest advisory -> real gate   exit 1, names GHSA-6w46-j5rx-g56g
M2  the true wider surface                     exit 0, "1 known, already in the baseline"
M3  baseline pypi section emptied              exit 1, names click PYSEC-2026-2132
```

**What this demo does not show:** no UI. This is a CI-surface change; the reproduction is the mutation matrix above, and the real proof is the `Security Audit` job on this PR running the widened export against the live advisory database.

PR #655, opened 2026-09-13
fix(ci): make the backend lint hook able to fail, and fix the 218 findings it was hiding (#561)

### LINKED ISSUE #561: [P3.28] Backend lint hook is vacuous twice over: `ruff` is not installed and the hook ends in `|| true`

frankbria:
Closes #561.

## The hook was vacuous three ways

```yaml
entry: bash -c 'cd backend && uv run ruff check app/ tests/ || true'
```

1. `|| true` swallowed the exit code.
2. `ruff` was in neither the main deps nor the `test` extra, so `uv run ruff` died with `Failed to spawn: ruff`.
3. That failure was swallowed too.

Every commit printed `Backend Linting (ruff)...Passed` having linted nothing.

**Mutation-confirmed in both directions** — the part that matters, since a hook that merely *looks* fixed is the same bug:

| | |
|---|---|
| `F821` injected into a production file, **new** hook | `Backend Linting (ruff)...Failed` |
| same defect, **old** `\|\| true` form | **exit 0** — would have reported Passed |

## The rule set is pinned, which the issue did not ask for

Measured, and it changes the shape of the fix: **ruff 0.16.7's own defaults report 1522 findings** in this tree, where the documented classic default (`E4,E7,E9,F`) reports **218**. Without an explicit `select`, this gate would change meaning on every ruff release and could turn red on a dependency bump that altered nothing in the repo. `[tool.ruff.lint] select = ["E4","E7","E9","F"]`, plus `target-version` and `line-length`.

## Turning it on found a real latent bug

`ruff --fix` removed `apply_chapter_content_update` and `update_book_summary_atomic` from `app/db/database.py` as unused imports — and **immediately broke the build**, because `books.py` imports both *from that facade*:

```
ImportError: cannot import name 'apply_chapter_content_update' from 'app.db.database'
```

They were part of the module's real public surface but **missing from its declared `__all__`**, so ruff was correct and the declaration was wrong. Fixed by adding them to `__all__` — F401 is now clean *without* removing the imports, and the facade's stated surface matches its actual one. That is the value the issue predicted from an honest first run, and it would not have surfaced any other way.

## 218 → 0

153 safe auto-fixes, then the remainder by hand rather than by ignore:

- **2 × `F841`** discarding the result of a side-effecting `await` — binding dropped, call kept
- **2 × `E722`** bare `except:` swallowing `BaseException`, incl. one around a `float()` parse → `except ValueError`
- **`F811`** duplicate `ObjectId` import (`bson` and a `.base` re-export)
- **2 × `F401`** dead imports inside `try:/except ImportError` availability probes
- **`F841`** an `if/else` whose first branch only bound a variable nothing read — condition inverted rather than left with an empty body
- **`F821`** a genuine bug: `tests/test_system_e2e.py` used `timezone` without importing it
- **`E402`** in `app/main.py` moved to the top (verified no circular import)

`conftest.py`'s 13 `E402`s are exempted **with a stated reason**: it must set the asyncio event-loop policy before pytest and motor are imported. That is a real ordering constraint, not a style lapse.

## Enforced in CI, not just locally

A pre-commit hook is bypassable with `--no-verify` and only fires on `files: ^backend/`. Added a `Backend lint (ruff)` step to the `Security Audit` job so the gate is actually enforced. It starts green — `app/` and `tests/` are at 0 findings — rather than arriving with a backlog.

## Two stale claims corrected

- **`CLAUDE.md` was wrong about its own hooks.** It said the installed `.git/hooks/pre-commit` is the bd flush hook and "runs no quality gates". It is the pre-commit framework's own hook and the gates **do** run — which is why `Backend Linting (ruff)...Passed` has been printing on every commit. Same class of defect as the issue itself: documentation asserting a gate's behaviour without checking it.
- **`black` is not claimed by this repo.** It is named as the backend formatter only in the *operator's global* `~/.claude/CLAUDE.md`, which is out of scope for a repo PR and is a cross-project preference. So there was no project-level false claim to remove. What I did instead is record the distance so the decision has numbers: **`black` would reformat 36 of 56 files; `ruff format` would reformat 42.** No formatter is enforced, and adopting either is its own mechanical PR.

## Verification

- `uv run --extra test ruff check app/ tests/` → **All checks passed** (0 findings)
- Full backend suite **1259 passed / 9 skipped, 92% coverage** — this is what validates the `--unsafe-fixes` used for `E712`, which is a genuine semantic change (`x == True` → `x`)
- `app.main` imports cleanly; both facade exports resolve
- `tests.yml` parses; both this PR's ruff step and #522's `--all-extras` widening survive the rebase
- Mutations above

## Known limitations

- The pinned set is deliberately narrow. Ruff's `UP` (424), `I` (149) and `B` (136) findings are real and unaddressed — a wider set is a follow-on burn-down, not a bigger version of this PR.
- No formatter is enforced, by decision rather than omission.


frankbria:
## Third-party review

**opencode/GLM (primary): skipped.** Two bounded attempts earlier this session (49-line and 448-line diffs) both returned zero bytes and had to be killed — the documented silent-hang mode. Skipped rather than spending a third timeout on a known failure. "Skipped" is not "clean": opencode has no verdict here.

> The changes primarily make the backend ruff lint gate effective, pin its rule set, and clean up the resulting lint findings. I did not identify any discrete introduced bug that would break existing behavior or CI.

Worth noting what it chose to do: for a 66-file mechanical lint sweep it **AST-parsed every file under `backend/app`** looking for syntax damage. That is the right instinct for this diff shape — the risk in a mass auto-fix is not logic, it is a broken file — and it came back clean.

### Limit

codex reviewed the diff; it did not run ruff or the test suite. The load-bearing claims here are "0 findings" and "1259 passed", and the evidence for those is mine plus CI, which now runs `ruff check app/ tests/` as an enforced step in the `Security Audit` job. **CI is the check that the gate is real** — and since this PR is what adds that step, this PR's own run is the first proof it works.

**AC — "the hook can fail"**, which is the whole issue:

```
F821 injected into app/services/chapter_status_service.py
  new hook:  Backend Linting (ruff)...Failed
  old form:  `... || true` -> exit 0        (would have reported Passed)
```

Both directions, on the same injected defect. The second is what proves the original was vacuous rather than merely untested.

**AC — "add ruff to the backend test extra so `uv run ruff` resolves"**

```
before: error: Failed to spawn: `ruff` / No such file or directory
after:  ruff 0.16.7
```

**AC — "drop `|| true`"** — done, and the entry now pins `--extra test` so the resolution is explicit rather than incidental.

**AC — "expect the first honest run to be red; decide then"**

218 findings. Decided: fix them, not baseline them — `app/` and `tests/` are at **0**. The one exemption (`conftest.py` E402 ×13) carries its reason in the config, because the asyncio event-loop policy genuinely must be set before pytest and motor import.

**AC — "decide black's status explicitly"**

Decided, and the honest answer is that the premise was slightly off: `black` is named only in the **operator's global** `~/.claude/CLAUDE.md`, not in this repo, so there was no project claim to drop and editing a cross-project preference file is out of scope for a repo PR. What is actionable is the measurement, now recorded: **`black` would reformat 36 of 56 files, `ruff format` 42.** No formatter is enforced, by decision.

**AC — "leave one mutation-checked tripwire"** — the two mutations above, plus the CI step means the tripwire is enforced rather than local-only.

**Unplanned outcome worth flagging:** turning the gate on found a real latent bug (`__all__` disagreeing with the facade's actual export surface, which broke the build the moment `--fix` acted on it) and a stale claim in `CLAUDE.md` about the repo's own hooks. Neither was in the issue.

PR #658, opened 2026-09-13
feat(ci): type-check the test sources against a ledger (#625)

### LINKED ISSUE #625: [P3.38] npm run typecheck covers no test file in the repo — tsconfig excludes **/__tests__/**, and SWC strips types without checking

frankbria:
Closes #625.

## The gap, confirmed rather than asserted

Per the repo's own `gate-must-cover-the-file` lesson, the input set was checked, not assumed:

```
$ npx tsc --noEmit -p tsconfig.json --listFiles | grep -c __tests__
0
$ npx tsc --noEmit -p tsconfig.test.json --listFiles | grep -c __tests__
(all of them — 3187 files in the project)
```

`tsconfig.json` excludes `**/*.test.ts(x)`, `**/__tests__/**`, `jest.setup.ts`, `**/playwright*.ts` and `**/e2e/**`; jest runs through `next/jest`, whose SWC transpile **strips types without checking them**. Between the two, nothing type-checked a test file.

**Not theoretical.** Earlier today #637's guard shipped two real `TS2345` arity errors that `npm run typecheck` could not see — caught only because that issue's own lesson said to run `tsc` on the file by hand. That is the workaround this PR replaces with a gate.

## The measurement chose the design

The issue said to measure before deleting exclusions, because the number decides everything. It does:

**176 errors across 29 files.** Deleting the exclusions outright would have blocked the next PR touching any of them — so this is the issue's path 3, in the ledgered shape the repo already uses for `security-baseline.json` and `gray-literal-baseline.json`:

- `tsconfig.test.json` — same compiler options, test files included
- `typecheck-test-baseline.json` — per-file counts, each with a reason naming that file's dominant error code (e.g. *"TS2540 ×27 — assigns to a read-only property (mocking a getter or a frozen global)"*)
- `scripts/typecheck-tests.mjs` — fails on any file **not** in the ledger or **above** its count; merely *reports* rows that are now clean or shrunk, so nobody's build breaks because someone fixed something

Concentration, for whoever burns it down: `metrics.test.ts` 45, `ChapterEditor.saveStatus` 18, `ChapterQuestionsPerformance` 17, `ChapterEditor.localStorage` 16, `errorHandler` 14 — five files carry 110 of the 176.

## The guards are deliberately unledgered

All eight `src/__tests__/theme/*` guards are **type-clean today**, so none has a ledger row. That is the point: they are the entire enforcement for #331/#610/#618/#620/#623/#629/#632/#634/#637 and they parse CSS and JSON by hand, so *a type error in a guard is a guard that may not be guarding*. With no row, any error introduced in one fails immediately — which is exactly what AC1 asks for.

## Mutation-checked three ways

| AC | mutation | fails as |
|---|---|---|
| 1 | `TS2322` into `core-authoring-pages-tokens.test.ts` | **no ledger row** |
| 2 | `TS2322` into `ChapterTab.keyboard.test.tsx` | **above recorded count**, `1 -> 2` |
| — | `TS2322` into `toast.test.ts` | **above recorded count**, `6 -> 7` |

**Being precise about AC2**: that file was *already* ledgered at 1 error, so the mutation exercised the count path, not the missing-row path. Both are the gate working, but they are different assertions and it would be wrong to claim the one that did not fire. AC1's target was unledgered and did exercise the missing-row path.

Every revert asserted with `git diff --quiet HEAD --`; the gate returns to PASS afterwards.

## One thing the script does that the issue did not ask for

It distinguishes a **crashed** `tsc` from a clean one. `tsc` exits non-zero both when it finds errors and when it breaks, so output carrying no `file(line,col): error TSxxxx` diagnostics is treated as a broken run and fails loudly. A gate that reads a failed tool as "no errors" is precisely the false-green this issue is about — the same reasoning `audit_gate.py` uses for a truncated audit report.

## Verification

- `npm run typecheck:tests` → **PASS**, 176 known, all ledgered
- Wall clock **6.1s** (base typecheck 1.5s) against the AC's ~30s ceiling
- Wired into **Frontend Tests**, which is branch-protected — not a local-only script
- Base `npm run typecheck` still clean; jest unchanged at **144 suites / 2628 passed**; lint 0 errors
- Three mutations above

## Known limitations

- `**/playwright*.ts` and `**/e2e/**` are now *included* by this config, and three `e2e` files are in the ledger. The issue asked for them to be decided together, and this decides them the same way: checked, with their current errors ledgered.
- The 176 are ledgered, not fixed. Five files carry 110 of them, so a burn-down is tractable but is its own PR.


frankbria:
## Third-party review

**opencode/GLM (primary): skipped** — two bounded attempts earlier in this session both returned zero bytes (documented silent-hang). No verdict from it, which is not the same as a clean one.

**codex (fallback): one [P2], and the honest answer is "does not reproduce, fixed anyway".**

> Because `tsconfig.test.json` extends the main config with `incremental: true`, the first `tsc` run writes `tsconfig.test.tsbuildinfo`; on a subsequent unchanged run, `tsc` can exit non-zero without re-emitting diagnostics, so this wrapper hits the "produced no output" branch and `npm run typecheck:tests` fails even though the ledger is unchanged.

### Reproduced first, per the repo's verify-before-fix rule

```
run 1  PASS — no new type errors in test sources (176 known, all ledgered)
run 2  PASS
run 3  PASS

raw tsc, repeat runs:
  run A exit=1 bytes=49239
  run B exit=1 bytes=49239
```

**It does not reproduce.** `tsc --noEmit` re-emits byte-identical diagnostics on a cached incremental run, so the crashed-tsc branch is never reached. The reasoning is plausible — it is exactly the shape of bug worth raising — but the behaviour is not there.

### Taken anyway, for a different reason than the one given

`incremental` is now `false` for this project, because **the option earns nothing here**: the gate reads tsc's stdout every run and never reuses output, while the inherited `incremental: true` was writing a ~700KB `tsconfig.test.tsbuildinfo` for no benefit. And the interaction codex describes is version- and environment-dependent enough that removing it beats relying on today's behaviour.

So: finding not confirmed, change made, and the changelog and config both say which is which rather than implying the review caught a live bug.

### The gate caught my own mistake doing it

The first attempt put the explanatory note **inside `compilerOptions`**. tsc validates keys there (a `"//"` key is only tolerated at a tsconfig's root), and the gate reported:

```
Type errors in files with no ledger row (1):
  tsconfig.test.json  (1)
```

A gate that fails on its own configuration file within minutes of being written is behaving correctly. Note moved to a root-level key.

### Also cleaned

`tsconfig.measure.tsbuildinfo`, a 704KB artifact left by my own measurement step. Gitignored, so it never threatened the repo, but untidy. All four `.tsbuildinfo` files in `frontend/` are confirmed gitignored and untracked.

### Limit of this review

codex read the diff; it did not run the gate. The load-bearing claims — 176 errors across 29 files, 6.1s wall clock, three passing mutations — are mine, and CI is what confirms them, since this PR adds the step to the branch-protected `Frontend Tests` job.

Issue #661, opened 2026-09-13
[P3.43] Every parallel PR conflicts on docs/CHANGELOG.md — six times in one batch, and one duplicate heading reached main


frankbria:
## The friction

`docs/CHANGELOG.md` groups entries under a date heading, newest-first *within* the day. Every PR therefore inserts at the same line, immediately under `### <today>`.

With more than one PR open at a time, that means **every merge makes every other open PR conflict**, always on the same file, always in the same way.

## Measured, not impressionistic

On 2026-09-12 this happened **six times** across eight PRs (#651, #652, #653, #654, #655, #656, #657, #658, #659, #660). Each occurrence needed the identical manual resolution: keep my bullet, keep everything from `main`, reorder. One of them also produced a **duplicate `### 2026-09-12` heading** that survived into `main` and had to be merged back by hand in #522, because the conflict resolution silently created a second section rather than appending to the first.

The resolution is mechanical enough that I ended up scripting it mid-session — which is the signal that it should not be manual at all.

## Why it is worth fixing rather than absorbing

- It is pure overhead: no PR has ever had a *semantic* conflict here, only a positional one.
- It scales with parallelism, so it gets worse exactly when throughput matters.
- It is silent-failure-prone. The duplicate-heading case shows the manual fix can leave `main` slightly wrong in a way no gate checks.
- A stale rebase is the moment the `#620`-family lessons bite: resolving a conflict badly is how a ledger row or a changelog entry goes missing without anyone noticing.

## Options

1. **Append within the day instead of prepending.** Entries land at the *end* of the current date section. Two PRs then edit different lines and git merges them cleanly. Costs: the newest entry for a given day is last, which reads slightly worse.
2. **One file per entry**, assembled at release (`docs/changelog.d/<issue>.md`, the `towncrier` / `changesets` shape). No two PRs ever touch the same file. Costs: a build step, and the assembled file becomes generated rather than edited.
3. **Leave it, add a guard.** Keep the current shape but add a test asserting each `### <date>` heading appears at most once, so the duplicate-heading failure can't reach `main` again. Cheapest, fixes only the silent half.

Recommend **1**, with **3** as a cheap independent addition regardless of which is chosen — the duplicate-heading guard is three lines and closes a real hole that already occurred once.

## Acceptance criteria

- [ ] Two PRs opened against the same day can both merge without a `docs/CHANGELOG.md` conflict — demonstrated, not asserted
[...]

PR #670, opened 2026-09-13
fix(lint): react-hooks/refs back to error (#584)

frankbria:
Closes the `react-hooks/refs` third of #584. Composes with #669 (`immutability`); together they leave only `set-state-in-effect`.

## I had called this un-investigable. It wasn't.

I recorded on #584 that `BookCreationWizard.tsx:100` was flagged at `form.handleSubmit(onSubmit)` *"for reasons I still cannot explain"*, and that guessing at a fix for a submit path with an in-flight guard is how double-submits ship. Then I treated the whole rule as blocked on that.

**It was blocked on reading the rule, which is in `node_modules`.**

```js
case 'MethodCall':
case 'CallExpression': {
    ...
    if (fnType.fn.readRefEffect) {
        error: 'Cannot access refs during render'
```

On a call evaluated **during render**, if any argument is a function carrying `readRefEffect`, it errors — because the callee might invoke it during render. `handleSubmit` is an opaque library call, and `onSubmit` reads `submitInFlight.current`.

So the fix is to move the **call** out of render:

```tsx
onSubmit={(e) => form.handleSubmit(onSubmit)(e)}
```

`handleSubmit(onSubmit)` returns a submit handler, so invoking it with the event is equivalent. One line, behaviour preserved, and the hypothesis was **confirmed by measurement before the fix was written** — not after.

## Three sites, three different shapes

| site | shape | fix |
|---|---|---|
| `BookCreationWizard:100` | ref-reading fn passed to an opaque call during render | call moved to event time |
| `useSwipeGesture:33` | latest-callbacks ref **assigned** during render | assignment moved into an effect |
| `dashboard/page.tsx:212` | ref read in render **and** in a callback | state for render, ref for callbacks, written together |

The dashboard one is the interesting case: a ref is *correct* at two of its three use sites. `hasLoadedOnce.current` is read inside the fetch callbacks, where it must be synchronous and must not itself cause a render. Only the render read is wrong — a render that depends on a ref will not re-run when the ref changes. So it now carries both flags, set in the same place so they cannot drift, with the reason in a comment.

`useSwipeGesture`'s write was actually safe — `latest.current` is only read inside event handlers, which run after commit — but a render-phase ref write is exactly what this rule exists to catch, and an effect is the documented home for it. The explanatory comment moved with it.

## The fourth mutation caught a faulty mutation, not a passing guard

| | mutation | result |
|---|---|---|
| M1 | `BookCreationWizard` call moved back into render | **errorCount: 1** |
| M2 | `useSwipeGesture` ref write back into render | **errorCount: 1** |
| M3 | dashboard render reads the ref again — *first attempt* | **errorCount: 0** ← the edit never applied |
| M3′ | same, redone without shell interpolation | **errorCount: 2** |

M3's first run reported zero because shell escaping mangled `&&` and the replacement silently matched nothing. **That is the exact near-miss I flagged earlier today** — a mutation that did not run is indistinguishable from a rule that does not fire, and it is the failure mode that would let a vacuous guard ship looking verified. Caught because the Python assertion raised rather than the count being taken at face value.

## Verification

- `react-hooks/refs` tree-wide: **0**; the `"warn"` downgrade deleted
- `npx eslint .`: **0 errors**
- `npm run typecheck` clean
- Dashboard tests **29/29**
- Four mutations above, each revert asserted against a committed baseline

## Composition

This branch still shows 2 `immutability` **warnings**, because that fix is #669 and not yet merged. Merging both leaves `set-state-in-effect` (26 sites) as the only rule still downgraded — and that one genuinely is a React refactor rather than a reading problem.


frankbria:
Rebased onto `main` after #673 merged — the conflict was `docs/CHANGELOG.md` only (the #661 shape: positional, not semantic; both entries kept). No code changed in the rebase.

Re-verified on the rebased branch: `npx eslint .` → **0 errors**, 437 warnings — which is the point of this PR, since `react-hooks/refs` is back at its default `error` here. 7 suites / 71 tests covering the three touched files pass.

PR #673, opened 2026-09-13
fix(chapters): remove TabBar's dead test-environment branch (#672)

### LINKED ISSUE #672: [P3.44] TabBar forks production code on a test environment — a dead branch that writes canScrollUp/Down behind handleScroll's back
## The finding

`src/components/chapters/TabBar.tsx:80-96` branches on whether `scrollIntoView` exists, and the fallback is labelled for tests:

```tsx
if (typeof activeTab.scrollIntoView === 'function') {
  activeTab.scrollIntoView({ behavior: 'smooth', block: 'nearest' });
} else {
  // For test environments, simulate scrolling by updating scroll state
  setCanScrollUp(true);
  setCanScrollDown(true);
}
```

Shipped code taking a different path because it detects a test environment. Surfaced while classifying #584's `set-state-in-effect` sites — it was flagged there, but it is not a state-derivation problem and does not belong in that fix.

## Why it matters

- **The branch that runs in tests is not the branch that runs for users.** Any test exercising scroll behaviour here exercises the fallback, so the real `scrollIntoView` path has no coverage at all — the shape of `gate-must-cover-the-file`, one level in.
- `setCanScrollUp(true)` / `setCanScrollDown(true)` in the fallback are **unconditional**, where the real handler computes them from `scrollTop` (line 47). So in tests the scroll affordances report a state the DOM does not have.
- jsdom not implementing `scrollIntoView` is a **test-setup** problem with a test-setup fix (a stub in `jest.setup.ts`), not a reason to fork production code.

## Suggested shape

1. Stub `Element.prototype.scrollIntoView` in `jest.setup.ts` — one line, applies everywhere.
2. Delete the `else` branch so the component has one path.
3. Confirm the affected tests still pass against the real path, and that `canScrollUp`/`canScrollDown` are still driven only by the scroll handler.

## Acceptance criteria

- [ ] `TabBar` has no test-environment branch
- [ ] `scrollIntoView` is stubbed at the test-setup level
- [ ] A test asserting scroll-affordance state exercises the same code path users get — mutation-checked by breaking the scroll handler and confirming it fails
- [ ] `grep -rn "test environment" frontend/src` finds no other shipped component doing this (I have not checked the rest of the tree)

frankbria:
Closes #672.

## The issue's premise was wrong, and the probe is why

I filed #672 five minutes before working it, claiming the `else` branch existed because `scrollIntoView` is missing under jest, and therefore *"the real `scrollIntoView` path has no coverage at all."*

`src/jest.setup.ts:134` already stubs `Element.prototype.scrollIntoView` — added for Radix Select. So the `else` branch cannot run even under jest.

Rather than argue that, I installed a probe in the branch:

```ts
} else {
  throw new Error('PROBE: the test-env branch DID run');
}
```

and ran every suite that mounts `TabBar`: **36 tests passed, the probe never fired.**

So the real `scrollIntoView` path is the *only* path that runs, and it is already covered. **This PR is dead-code removal, not a coverage fix.** The issue body has been corrected on GitHub to say so.

## What changed

`TabBar` branched on `typeof activeTab.scrollIntoView === 'function'` and, in the `else`, set `canScrollUp`/`canScrollDown` to `true` unconditionally "for test environments" — where `handleScroll` derives both from `scrollTop`/`scrollHeight`. Two writers for one pair of state values, one of them lying.

- The branch is gone. `scrollIntoView` is called directly.
- The `try/catch` stays, comment corrected. It claimed to guard a missing method in tests; what it can actually catch is an engine throwing on the options object.
- `canScrollUp`/`canScrollDown` now have exactly one writer.

## Acceptance criteria

| AC | Result |
|----|--------|
| Dead branch removed | Done — `git diff` shows the `else` gone |
| Real path still exercised | 36 tests across the TabBar suites, probe-verified |
| No other shipped component forks on the test environment | Swept `test environment`, `NODE_ENV === 'test'`, `isTest`, `jsdom` across `src/` — **none** |
| Suite green | 146 suites, 2652 passed, 4 skipped |

Side effect: this removes one `set-state-in-effect` violation (the `setCanScroll*` pair inside the effect), one of the 22 sites classified on #584.

## Verification

- `npx eslint src/components/chapters/TabBar.tsx` — 0 errors
- `npm run typecheck` + `npm run typecheck:tests` — clean
- Full jest run — 146 suites, 2652 passed
- Pre-PR third-party review: `codex review --base main` (opencode attempted first, produced 0 bytes again — the documented hang). Verdict: *"I did not identify any actionable regression introduced by this diff."*

## Known limitations

The `catch` is now genuinely unreachable in every environment this ships to and in jest. It is kept as a one-line `console.debug` because a throwing `scrollIntoView` would otherwise take down the effect; removing it is not worth a separate PR, and keeping it costs nothing.


frankbria:
## Post-PR third-party review

Diff unchanged since the pre-PR pass (`HEAD == REVIEWED_SHA`), so this is the recorded 7c-1 output rather than a re-run.

**Reviewer:** `codex review --base main`. opencode was attempted first and produced 0 bytes — the documented concurrent/large-diff hang, third occurrence this session.

**Verdict:**

> The change removes a dead fallback around `scrollIntoView` while preserving defensive error handling, and the surrounding scroll state is still derived by the existing scroll handler. I did not identify any actionable regression introduced by this diff.

No findings to triage.

frankbria:
## Demo — acceptance criteria with outcome evidence

My first probe ("36 tests passed, the `throw` never fired") was weaker than I presented it. A `throw` in the `else` also stays silent when the **whole effect body** is skipped, and in `TabBar.test.tsx` it is: that suite stubs `ChapterTab` down to a plain `div` that drops `...provided.draggableProps`, so `querySelector('[data-rfd-draggable-id=…]')` returns `null` and the effect returns before either branch. Silence there proves nothing.

Re-probed at the branch point instead, across all six suites that mount `TabBar`:

| Probe site | Count |
|---|---|
| `activeChapterId`/ref early return | 0 |
| `querySelector` → **null** (effect returns) | 22 |
| `querySelector` → **FOUND** (branch point reached) | **5** |

So the branch point *is* reached 5 times — by the suites that render the real `ChapterTab` — and the `else` still never ran, because `jest.setup.ts:134` stubs `scrollIntoView`. Dead confirmed, on evidence that could have come out the other way.

**AC2, outcome evidence.** Instrumented the call itself rather than asserting "it rendered":

```
4 × PROBE672:CALLED ch-1 {"behavior":"smooth","block":"nearest"}
1 × PROBE672:CALLED ch1  {"behavior":"smooth","block":"nearest"}
```

Five real invocations, each on the element whose `data-rfd-draggable-id` matches the active chapter, each with the intended options. The real path is covered.

| AC | Evidence |
|----|----------|
| 1 — dead branch removed | `grep` for `test environment` / `typeof activeTab.scrollIntoView` in `TabBar.tsx`: no match |
| 2 — real path exercised | 5 instrumented calls above, correct element and options |
| 3 — single writer for `canScrollUp`/`canScrollDown` | only `handleScroll` assigns them |
| 4 — no other shipped component forks on the test env | swept `test environment`, `NODE_ENV === 'test'`, `isTest`, `jsdom` across `src/`: 3 hits, all benign — a `__mocks__` file, and two comment lines in `src/lib/navigation.ts` |

On that last one: `src/lib/navigation.ts` is the contrast worth naming. It exists *because* `window.location` cannot be stubbed under jsdom 26 — but it solves that with a **seam** (a module tests can `jest.mock`) whose runtime behaviour is byte-identical, not with a branch that behaves differently when observed. That is the pattern; what `TabBar` had was the anti-pattern.

All probes reverted; `git diff --quiet HEAD` clean after each. Suites: 6 passed, 86 tests.

PR #674, opened 2026-09-13
fix(lint): three more set-state-in-effect sites, one hiding a staleness bug (#584)

frankbria:
Partial progress on #584, which the issue explicitly allows — the three rules only flip back to `error` at zero. `set-state-in-effect` **21 → 18**; `refs` unchanged at 4.

## One of these was not a lint fix

`useUserPreferences` kept a module-level cache and an effect that mirrored it into component state:

```ts
const [preferences, setPreferences] = useState(cachedPreferences);
useEffect(() => {
  if (cachedPreferences) { setPreferences(cachedPreferences); return; }
  …
}, [getUserProfile]);
```

The mirror was not merely an extra render. **An already-mounted consumer never saw an invalidation.** `src/app/dashboard/settings/page.tsx:137` and `src/app/profile/page.tsx:157` both call `invalidateUserPreferencesCache(updated?.preferences ?? null)` after a save — and at that moment `ChapterEditor`, `DraftGenerator`, `ExportOptionsModal` and `DraftGenerationButton` are mounted with an effect that has already run and nothing to re-run it. They kept serving pre-save preferences until they remounted.

The existing suite looked like it covered this:

```ts
invalidateUserPreferencesCache({ …, default_writing_style: 'creative' });
const second = renderHook(() => useUserPreferences());   // ← newly rendered
expect(second.result.current).toMatchObject({ default_writing_style: 'creative' });
```

A **newly rendered** hook reads the fresh cache in its `useState` initializer. That was the one case that was never broken.

### The fix

The cache becomes a real external store — a listener set, `publish()` on fill and on invalidation, and `useSyncExternalStore` with `getServerSnapshot: () => null` so the first paint matches the old `useState(null)`. Two details kept deliberately:

- A consumer mounting **mid-fetch** still gets the result. The old hook chained its own `.then` onto the shared promise; now it subscribes, which is equivalent and one fewer moving part. Pinned by a test.
- `inflight` is cleared only by the generation that set it. `invalidateUserPreferencesCache` already nulls it, so an unguarded `.finally` on a stale fetch could null out a **newer** one.

## Two genuine derivations

`LoadingStateManager` mirrored `visible` and `timeRemaining` through effects. Both are functions of props:

```ts
const visible = isLoading && delayElapsed;
const timeRemaining = estimatedTime === undefined ? undefined : Math.max(estimatedTime - countedDown, 0);
```

The effects now own only the timers. Because `isLoading` alone already forces `visible` false, the reset moved into effect cleanup and a finished load no longer costs a second render pass.

## Mutation checking found this refactor untested — in my own code

Dropping **either** cleanup reset left all 53 existing tests green:

| Mutation | Before | After |
|---|---|---|
| drop `setDelayElapsed(false)` from cleanup | 53 passed | **1 failed**, 54 passed |
| drop `setCountedDown(0)` from cleanup | 53 passed | **1 failed**, 54 passed |
| old effect-mirror `useUserPreferences` vs. the new tests | — | **1 failed**, 5 passed |

Not one existing test loads twice. Without the resets, a second load would skip the 200ms anti-flicker delay entirely, and the countdown would carry over — telling a user that a fresh 5-second operation has 2 seconds left. Two tests now walk a second load, and each mutation fails exactly one of them. The third row is the staleness fix, pinned against the old hook.

Each mutation was reverted with the revert asserted (`git diff --quiet HEAD -- <file>`), on a committed tree.

## Why only three

The remaining 18 are classified per-site in [this comment](https://github.com/frankbria/auto-author/issues/584#issuecomment-5650755879). I looked again at six more while here and they are genuinely not mechanical:

- **`settings/page.tsx:86`** (`setActiveTab` from `?tab=`) and **`ChapterEditor.tsx:126`** (restore the tab from `sessionStorage`) look like `useState` initializers, but both read browser-only state. Moving them into the initializer makes the client's first render differ from the server's — a hydration mismatch traded for a render. The effect is the shape that avoids it.
- **`ChapterEditor.tsx:180`** (`hasBackup`) has **eight** setters — restore, dismiss, auto-save. It is real state seeded from storage, not an external store.
- **`settings/page.tsx:76`** and **`ActiveSessionsList.tsx:61`** are fetch kick-offs; the rule flags a synchronous `setState` inside the callee, and fixing that is a data-layer decision.

## Verification

- `npx eslint . -f json` → `set-state-in-effect: 18`, `refs: 4`, **0 errors** tree-wide
- Full suite: **146 suites, 2656 passed**, 4 skipped
- `npm run typecheck` and `npm run typecheck:tests` clean (176 ledgered, no new)
- Pre-PR third-party review: `codex review` — *"I did not identify any discrete correctness, security, or reliability issues introduced by the diff."* (opencode attempted first, zero output again.) Codex's own targeted `npm test` failed to start on a `--showConfig` parse error inside its sandbox; the same command passes here, and the full suite above is the evidence.

## Known limitations

`invalidateUserPreferencesCache()` with no argument clears the cache but does not trigger a refetch in mounted consumers — they render `null` (their own defaults) until something remounts. That is unchanged from before this PR, and no caller uses the no-argument form outside tests.

Closes nothing — #584 stays open until the count reaches zero.


frankbria:
Rebased onto `main` after #670 and #673 merged — `docs/CHANGELOG.md` only, the #661 shape (positional, not semantic; every entry kept). No code changed in the rebase.

Re-verified on the rebased branch, which now has `react-hooks/refs` back at `error` from #670: `npx eslint .` → **0 errors**. Loading and preferences suites: 61 passed.

Auto-merge enabled (squash).

Issue #613, opened 2026-09-08
[P3.37] Dashboard E2E page object targets data-testids the app has never rendered


frankbria:
Surfaced while mapping the dashboard for #493.

## The mismatch

`frontend/tests/e2e/page-objects/dashboard.page.ts` locates books by two test ids:

```ts
24:  return this.page.locator('[data-testid="books-list"]');
48:  return this.page.locator(`[data-testid="book-item"]:has-text("${title}")`);
93:  const books = this.page.locator('[data-testid="book-item"]');
```

Neither exists in the app. `BookCard` renders:

```
frontend/src/components/BookCard.tsx:84:   data-testid="book-card"
```

and the dashboard renders the grid inline with no list wrapper at all, so there is no `books-list` element to find either.

## Why it matters

These locators cannot match, so every helper built on them either times out or silently reports zero books. A "count the books on the dashboard" helper that always returns 0 does not fail loudly — it makes any assertion of the form "no books yet" or "fewer than N books" pass for the wrong reason. That is the [[guard-tests-can-self-match]] shape: a check that runs, goes green, and measures nothing.

Worth auditing which staging specs actually depend on these helpers before deciding how much is currently vacuous.

## Fix direction

Point the page object at `book-card`, and add a `data-testid="books-list"` wrapper to the dashboard grid if a list-level locator is genuinely wanted — #493 has just added a pager below that grid, so a wrapper is now a more useful anchor than it was.

## Acceptance criteria

- [ ] Page-object locators match elements the app actually renders.
- [ ] Any spec whose assertions were passing on a zero-length locator identified and fixed.
- [ ] The book-count helper proven non-vacuous — it returns 24 on a seeded page, not 0.

PR #676, opened 2026-09-13
feat(ci): ledger the react-hooks backlog so it can only shrink (#675)

### LINKED ISSUE #584: [P3.31] 30 react-hooks v7 violations warned rather than fixed during the ESLint 9 migration

### LINKED ISSUE #675: [P3.35] set-state-in-effect is the one backlog in the repo with no ledger — 18 violations sit in 437 warnings nobody reads
`frontend/eslint.config.mjs` downgrades three `eslint-plugin-react-hooks` 7 rules to `warn` so the ESLint 9 upgrade could land as an upgrade. #669 restored `immutability`, #670 restores `refs`, and #671/#673/#674 have taken `set-state-in-effect` from 26 to 18. The last one will not reach zero soon: the per-site classification on #584 shows the remainder are fetch kick-offs needing a data-layer decision, latches, and initialization-from-storage constrained by hydration.

**Until it reaches zero the rule is `warn`, and `npx eslint .` already prints 437 warnings.** A new violation lands in that pile and nobody sees it. The backlog is not the risk; the backlog growing invisibly is.

## What this is

Every other backlog in this repo is a **ledger that can only shrink**, not a downgraded rule:

| Ledger | Gate |
|---|---|
| `security-baseline.json` | `scripts/audit_gate.py` |
| `gray-literal-baseline.json` | `src/__tests__/theme/gray-literals.test.ts` |
| `dark-overlay-baseline.json` | #632's guard |
| `stranded-accent-baseline.json` | #637's guard |
| `typecheck-test-baseline.json` | `scripts/typecheck-tests.mjs` |
| `unrun-spec-baseline.json` | `scripts/check-spec-coverage.mjs` |

`set-state-in-effect` is the one backlog in the repo with no ledger. It should have the same shape: the 18 known sites recorded per file with a count, anything **not** in the ledger fails, and a file whose count drops is reported as prunable rather than failing anyone's build.

## Acceptance criteria

1. `frontend/react-hooks-baseline.json` ledgers the current violations **per file, per rule, with a count** — per the #624 rule that a per-file total lets a net-zero swap through, and per the #631 rule that a row the tree no longer contains must fail as stale rather than quietly pre-authorising a future violation.
2. A script (`frontend/scripts/react-hooks-gate.mjs`) runs ESLint over the tree, compares against the ledger, and fails on: a violation in a file with no row, a file above its recorded count, or a stale row. Shrinkage is **reported, not failed** — the contract `audit_gate.py` and `typecheck-tests.mjs` already use.
3. It distinguishes a crashed ESLint run from a clean one. `typecheck-tests.mjs` had to learn this the hard way: a run that produces no parseable diagnostics is a broken run, not a pass.
4. Wired into the `Frontend Tests` workflow, so it is enforced rather than available.
5. Mutation-checked in both directions: introducing a new violation in a ledgered file fails; introducing one in an unledgered file fails; deleting a ledgered row while the violation remains fails; fixing a violation is reported, not failed.
6. A vacuity assertion: if the ESLint sweep returns a suspiciously small number of files, the gate fails rather than passing over an empty set (#613's shape).

## What this does not do

It does not fix any of the 18, and it does not close #584 — #584's condition is the rules back at their default `error`, which stays correct. This makes the interval until then safe, which is the part currently missing. When the count reaches zero the ledger empties and the gate can be deleted along with the `warn` lines.

frankbria:
Closes #675. Does **not** close #584 — its condition is the three rules back at their default `error`, which stays right.

## The gap

`eslint.config.mjs` holds three `eslint-plugin-react-hooks` 7 rules at `warn` so the ESLint 9 upgrade could land as an upgrade. #669 and #670 restored two of them; `set-state-in-effect` is down from 26 to **18** across #671, #673 and #674. The rest will not go quickly — the [per-site classification](https://github.com/frankbria/auto-author/issues/584#issuecomment-5650755879) shows fetch kick-offs needing a data-layer decision, a latch, and initialization-from-storage constrained by hydration.

Meanwhile `npx eslint .` prints **434 warnings**. A new violation lands in that pile and nobody sees it. The backlog was never the risk; the backlog growing invisibly was.

## The shape this repo already uses

| Ledger | Gate |
|---|---|
| `security-baseline.json` | `scripts/audit_gate.py` |
| `gray-literal-baseline.json` | `theme/core-authoring-pages-tokens.test.ts` |
| `dark-overlay-baseline.json` | #632's guard |
| `stranded-accent-baseline.json` | #637's guard |
| `typecheck-test-baseline.json` | `scripts/typecheck-tests.mjs` |
| `unrun-spec-baseline.json` | `scripts/check-spec-coverage.mjs` |
| **`react-hooks-baseline.json`** | **`scripts/react-hooks-gate.mjs` ← this PR** |

`set-state-in-effect` was the one backlog here with a downgraded rule instead of a ledger.

**Counts are per file per rule.** A per-file total would let a `refs` violation swap in for a `set-state-in-effect` one and net out — the #624 rule. **A stale row fails**, because it pre-authorises the next violation of that rule in that file — the #631 rule. **A count that drops is reported, never failed** — `audit_gate.py`'s contract; nobody's build should break because someone fixed something.

Current ledger: **18 violations across 13 files**, all `set-state-in-effect`. Runs in 7.7s.

## It caught two real stale rows before CI ever ran it

Neither was planted:

1. The first baseline was generated before #670 merged, so it ledgered four `refs` rows that #670 had already fixed. The gate failed them as stale.
2. Rebasing past #674 left two more — `LoadingStateManager` (2) and `useUserPreferences` (1).

Both are the ordinary consequence of a ledger written at one commit and read at another, which is exactly how a stale row reaches a repo. Each is a separate commit here rather than squashed into the first, so the sequence is on the record.

## Mutations

Seven, covering both directions and the ways the gate itself could go quiet:

| # | Mutation | Result |
|---|---|---|
| 1 | new violation in a **ledgered** file | FAIL — `src/app/page.tsx  set-state-in-effect: 1 -> 2` |
| 2 | new violation in an **unledgered** file | FAIL — `src/components/chapters/TabBar.tsx (1)` |
| 3 | delete a ledger row while the violation remains | FAIL — reported as unledgered |
| 4 | ledger a violation that does not exist | FAIL — stale row |
| 5 | inflate a ledgered count (simulating a fix) | **PASS**, reported under *Shrunk* |
| 6 | point the sweep at a 3-file subtree | FAIL — vacuity assertion, `3 < 200` |
| 7 | make ESLint emit `stylish` instead of `json` | FAIL — "broken run, not a pass" |

6 and 7 are the ones that matter most: a gate whose sweep silently returns nothing passes forever, which is #613's shape, and `typecheck-tests.mjs` had to learn the same lesson about a crashed `tsc`. Every mutation was reverted with the revert asserted (`git diff --quiet HEAD -- <file>`) on a committed tree, and all seven re-run after the review fixes below.

## Two real defects the pre-PR review caught

`codex review` (opencode attempted first, zero output — fourth time this session), both fixed in `0b4b678`:

- **[P1] The baseline ledgered four `refs` rows #670 had fixed.** Finding 2 above — the gate was right and the ledger was stale.
- **[P2] The pair key joined file and rule with a literal NUL byte**, so git classified the script as binary: `git diff` reported `Bin 0 -> 6581 bytes` and the entire gate would have been **unreviewable in this PR**. Replaced with `[file, rule]` tuples. Any delimiter is a guess about what a path cannot contain, and none is needed. Now diffs as 186 lines of text; `NUL bytes: 0` asserted.

## Verification

- `npm run gate:react-hooks` → PASS, 18 known, all ledgered
- `npx eslint .` → **0 errors**
- Full suite: 146 suites, 2652 passed, 4 skipped
- `npm run typecheck`, `typecheck:tests` (176 ledgered, no new), `check:specs` — all clean
- `.github/workflows/tests.yml` parses; the step sits with the other frontend ledger gates

## Known limitations

The gate runs its own full ESLint pass rather than reusing the `Run linting` step's output. That is 7.7s of duplicated work; plumbing a JSON artifact between two steps costs more than it saves at this size. Worth revisiting only if CI time becomes a problem.

`MIN_FILES_LINTED` is a fixed 200 against a tree that lints 398 files. It is a tripwire for a broken sweep, not a coverage measure — if the frontend ever genuinely shrinks below 200 files, this fails loudly and gets adjusted, which is the right failure mode.


frankbria:
## Post-PR third-party review

Diff unchanged since the pre-PR pass on the final commit, so this records that pass rather than re-running it.

**Reviewer:** `codex review`. opencode was attempted first and produced 0 bytes — the documented hang, fourth occurrence this session.

It returned **two findings, both real, both fixed before this PR opened** (`0b4b678`):

- **[P1]** the baseline ledgered four `react-hooks/refs` rows that #670 had already fixed, so the gate would have failed its own first CI run. Rebaselined — and this is the gate's stale-row rule working as designed, against its own ledger.
- **[P2]** the pair key used a literal NUL byte, making git classify `react-hooks-gate.mjs` as binary (`Bin 0 -> 6581 bytes`). The whole script would have been unreviewable in this diff. Replaced with `[file, rule]` tuples.

All seven mutations were re-run after those fixes and still behave correctly.

PR #677, opened 2026-09-13
fix(lint): ExportProgressModal's elapsed reset moves into cleanup (#584)

### LINKED ISSUE #584: [P3.31] 30 react-hooks v7 violations warned rather than fixed during the ESLint 9 migration

frankbria:
Partial progress on #584. `set-state-in-effect` **18 → 17**. The rules stay at `warn` until zero, per the issue's close condition.

## The route I had said did not exist

I wrote on #584 that `ExportProgressModal:72` had *"no fix I could find that does not either keep the synchronous reset or read a ref during render, which would trip the `refs` rule."*

It had one, and #674 found it by using it elsewhere — **reset in the effect's cleanup**:

```ts
useEffect(() => {
  if (status !== 'processing' && status !== 'pending') return;
  const interval = setInterval(() => setElapsedTime((prev) => prev + 1), 1000);
  return () => { clearInterval(interval); setElapsedTime(0); };
}, [status]);
```

`elapsedTime` is only read while `status` is `processing` or `pending`, so clearing it on the way out rather than on the way in is invisible — and it costs one render pass fewer than a synchronous `setState` in the effect body.

## The component had no test file

Not a thin one — none. Nothing covered the counter, the two thresholds it drives (`> 10` shows elapsed time, `> 20` swaps to "this is taking longer than expected"), or what a **second** export sees. Five tests now do.

The pair that matters cuts both ways:

| Mutation | Result |
|---|---|
| drop `setElapsedTime(0)` from cleanup | **2 of 5 fail** |
| restore the original pre-refactor effect (reset at the top) | **5 of 5 pass** |

The second row is the one worth reading. These tests describe the component's behaviour, not this refactor — so the fact that the old code passes them is how the change is known to be behaviour-preserving rather than merely green. Without a test that discriminates, a refactor and a regression look identical.

With the reset dropped, a second export inherits the first one's count and opens straight into "this is taking longer than expected" — for an export one second old.

## The gate from #675 had its first real use

One PR after landing, it refused this one until `ExportProgressModal`'s ledger row was pruned:

```
Stale rows — these are fixed, delete them (1):
  src/components/export/ExportProgressModal.tsx  react-hooks/set-state-in-effect (1)
```

Pruned; 17 across 12 files.

## A change that did *not* ship

In the same comment I called **`ActiveSessionsList:61`** "exact": `loadState` already initialises to `'loading'`, and `loadSessions` has exactly one non-effect caller (the Retry button), so hoisting `setLoadState('loading')` into that handler should have removed the violation.

I made the change and measured. **The warning stayed.** `set-state-in-effect` flags *any* `setState` reachable inside an `async` callee an effect invokes — `await` first or not. The shape of the reset was never what the rule objected to. Reverted; nothing from it is in this diff.

### What the probe found next, and why I stopped

Converting the callee from `async`/`await` to a promise chain, changing nothing else:

```ts
const loadSessions = useCallback(() => {
  return authClient.listSessions()
    .then(({ data, error }) => { …; setLoadState('loaded'); })
    .catch(() => setLoadState('error'));
}, []);
```

**silences the rule.** So every remaining fetch kick-off is mechanically "fixable" — by reshaping readable `async`/`await` into promise chains to suit a linter that cannot analyse `await`. Both forms satisfy the rule's stated intent; only one is legible to it.

That would close #584 and leave the codebase worse, so it is [recorded on the issue](https://github.com/frankbria/auto-author/issues/584#issuecomment-5651717982) as a trap rather than a route. It is also the clearest argument for the ledger #676 just landed: the 17 cannot grow, and each can be fixed when there is a reason beyond the count.

## Verification

- `npm run gate:react-hooks` → PASS, 17 known, all ledgered
- `npx eslint .` → **0 errors**
- `npx jest src/components/export` → 41 passed
- Pre-commit ran in full (lint, typecheck, unit tests, coverage ≥85%) — all passed, no `--no-verify`
- Pre-PR third-party review: `codex review` — *"No discrete regressions or actionable bugs were found… the baseline update matches the removed lint violation."*

## Known limitations

`ExportProgressModal` still has no coverage for its four status states, the cancel affordance, or the error path; this PR adds only the elapsed-time tests, because that is what changed. The rest is worth a test file of its own rather than being smuggled in here.


frankbria:
## Post-PR third-party review

Diff unchanged since the pre-PR pass, so this records that pass rather than re-running it.

**Reviewer:** `codex review` (opencode attempted first, 0 bytes — fifth occurrence this session; the documented hang).

> No discrete regressions or actionable bugs were found in the changed files. The elapsed-time refactor preserves the intended counter behavior and the baseline update matches the removed lint violation.

No findings to triage.

PR #678, opened 2026-09-13
fix(git): union-merge the changelog so parallel PRs stop conflicting (#661)

frankbria:
Addresses the cost #661 describes without making the decision #661 defers. The issue stays open for the convention question.

## The measurement

Since #661 was filed, **every** PR has conflicted on `docs/CHANGELOG.md` and nothing else:

| PR | Conflicted with |
|---|---|
| #673 | #671 |
| #670 | #673 |
| #674 | #670, then #673 |
| #676 | #674 |
| #677 | #676 |
| **this one** | #677 |

Eight resolutions, all positional, all mechanical, none semantic. Two landed on top of *another* conflict in the same file minutes later. #661's own guard (`test_changelog_headings.py`) fired correctly every time, so the duplicate-heading failure mode stayed closed — the cost is entirely in the resolutions.

## What this is, and what it is not

`.gitattributes`:

```
docs/CHANGELOG.md merge=union
```

`union` is one of git's built-in low-level merge drivers: on a conflicting hunk it keeps **both** sides instead of writing markers. For an append-only log where every PR inserts a distinct line at the same offset, that is exactly the resolution a human performs by hand, every time.

**#661 deliberately leaves the convention question to a human** — append-within-day vs. one file per entry assembled at release — because it changes how everyone writes entries. This is not that. Entries are written exactly as they are today. It stops the deferred decision from costing a resolution per PR while it waits.

## Guarded by reproduction, not by grep

`scripts/test_changelog_merge.py` builds a throwaway repo, copies **this repo's real `.gitattributes`** into it, and runs the actual case: two branches each inserting an entry under the same `### <date>`. A test that greps a config file for a line would prove the line is there and nothing about whether git honours it.

The anti-vacuity half builds the same fixture **without** the driver and asserts it still conflicts — a fixture that cannot conflict would make the other assertion pass whatever `.gitattributes` said, which is #613's shape. It also pins what matters after the merge: both entries present, **one** heading, no markers.

Writing that test found a thing worth knowing: the driver must be on the **base** commit, because git reads merge attributes from the tree being rebased onto, not from the commits being replayed. My first fixture emptied `.gitattributes` on the feature branch and the rebase succeeded anyway — the anti-vacuity case was passing vacuously in the opposite direction.

| Mutation | Result |
|---|---|
| delete `docs/CHANGELOG.md merge=union` | FAIL |
| change `union` to a driver that does not exist | FAIL |

The second matters more than it looks: git falls back to a normal merge **silently** when a named driver is not configured, so a typo would restore the conflicts with no signal at all. That is the likelier real-world breakage, and `merge=beads` on the line above is precisely such a custom driver.

No workflow change needed — CI's `Test the audit gate and dependabot config` step already runs `uvx --with pytest --with pyyaml pytest scripts/ -q`, which picks the new file up. **83 passed, 4 skipped.**

## Known limitations

- **The blank line between two merged entries is collapsed.** Both remain one markdown list and render correctly, but the source spacing differs from a hand resolution.
- **Two PRs genuinely editing the same entry would have both halves kept silently** rather than flagged. For an append-only log that does not happen, and `test_changelog_headings.py` still catches a mis-merged `### <date>` section.
- **Untested: whether GitHub's server-side mergeability check honours `.gitattributes`.** All eight conflicts above were paid in local `git rebase`, which does honour it, so the benefit stands either way. The next pair of parallel PRs settles the question empirically.
- **This PR itself conflicted on the changelog**, because `main` does not carry the driver until it lands. That is the last one.


frankbria:
## Post-PR third-party review

**Reviewer:** `codex review` (opencode attempted first, 0 bytes again).

> The diff adds a targeted `.gitattributes` union merge rule for the changelog plus a reproduction test that exercises the intended same-heading rebase scenario and anti-vacuity conflict case. I did not identify any discrete introduced bug that would break existing behavior or tests.

No findings to triage.

PR #680, opened 2026-09-13
fix(git): reject a commit message that closes an issue it says it does not

### LINKED ISSUE #123: [P1.14.4] Test coverage for auth/session validation cluster
Parent: #117 (coverage triage)

## Gap (cohesive auth plumbing, ~101 uncovered stmts)
- `core/better_auth_session.py` — 36% (49 missing) — session cookie validation
- `api/dependencies.py` — 64% (29 missing) — auth dependencies
- `api/middleware/session_middleware.py` — 53% (23 missing)

High security criticality.

## Scope (atomic, one session)
Cover valid/expired/invalid/missing session-cookie paths, dependency resolution, and middleware behavior.

## Acceptance
- [ ] All three modules >=85%
- [ ] Negative auth paths (invalid/expired/missing cookie) covered

### LINKED ISSUE #584: [P3.31] 30 react-hooks v7 violations warned rather than fixed during the ESLint 9 migration

frankbria:
## A disclaimer written to prevent a closure caused one

The merge commit for the ledger-gate PR carried this sentence, added specifically so the issue would stay open:

> Does not `close` `<issue>` — its condition is the rules back at `error`.

It closed that issue, with **17 violations outstanding and all three rules still at `warn`**. GitHub's linked-issue parser matches `close|fixes|resolved` followed by an issue number anywhere in a commit message or PR body, and does not parse negation. The disclaimer *was* the trigger.

Reopened, with a comment explaining the cause. Nothing was lost — the gate is in place and the ledger is accurate — but the state was wrong for about fourteen hours.

*(This PR body deliberately avoids writing the trigger phrase, for the same reason.)*

## The guard

`scripts/check-commit-message.sh`, wired at the `commit-msg` stage. Sixteen fixtures pin it on **both** sides:

- **Rejected (8):** the real sentence verbatim; `doesn't fix`; `will not resolve`; `Does NOT Close`; `does not fixes`; a colon between keyword and number; a body line opening with `#`; and a negation wrapped across a line break.
- **Accepted (7):** deliberate `Closes #584` and `Fixes #12` — a guard that rejects everything is as useless as one that rejects nothing, and only the second half of that is obvious; keyword-free references (`#584's close condition`, `Partial progress on #584`, `Per #584`); and negation with no issue number (`does not close the dialog`).

To reference an issue without closing it, drop the keyword: **`#584's close condition`**, not the negated form.

## Three defects in the guard, and all three were the same class

None was in the pattern. The pattern was right from the first line and never changed. All three were in the **filtering of the input** — which is the thing to suspect in any guard that reads a file line by line.

**1. The wiring assertion was vacuous.** "Is this hook actually wired up" checked for the string `commit-msg` anywhere in `.pre-commit-config.yaml` — which also matches `default_install_hook_types: [pre-commit, commit-msg]` a hundred lines away. Retagging the hook to `stages: [manual]` left it green. Found by mutating the one thing the assertion claimed to check. Now parses the YAML and asserts the hook's own `stages`.

**2. It skipped lines starting with `#`.** Reasoning: git strips comments. Git does — in **editor** mode, where cleanup defaults to `strip`. `git commit -m` uses `--cleanup=whitespace` and keeps every line. The very first message this guard ran on opened its body with an issue number, so the skip discarded that entire line *and the negated keyword inside it*. **The hook reported `Passed` on the one message it exists to catch**, and that commit would have closed the issue a second time. Caught by running the finished guard against the message that motivated it.

**3. `grep` is line-based; commit bodies wrap at ~72 columns.** Found by the pre-PR reviewer:

```
fix: x

This does not
close #123 yet.
```

reads as safe line by line, while GitHub — which has no concept of the wrap — closes #123. Reproduced against the previous version (`exit=0`) before fixing. Newlines now collapse to spaces before matching.

## Mutations

Eight, including reverting each of the three fixes individually:

| Mutation | Result |
|---|---|
| drop the negation requirement | 2 fail — legitimate `Closes #584` now rejected |
| stub the check to always pass | 9 fail |
| revert to line-based matching (undo fix 3) | 2 fail |
| re-add the `#`-line stripping (undo fix 2) | 1 fail |
| retag the hook to `stages: [manual]` (undo fix 1) | 1 fail |
| drop `commit-msg` from `default_install_hook_types` | 1 fail |
| delete the hook block entirely | 1 fail |
| remove the executable bit | 14 fail |

Each reverted with the revert asserted, on a committed tree.

## A fourth instance of a known mistake, recorded rather than quietly fixed

The mutation loop's `git checkout HEAD -- .pre-commit-config.yaml` silently deleted the **uncommitted hook block it was testing**, and the next run failed against a tree missing the feature — which reads exactly like a broken test. The standing lesson says commit before mutating; the gap was applying that to the file under mutation and not to the other files the same loop reverts. `tasks/lessons.md` now says so.

## Known limitations

- **A PR body closes issues the same way, and no local hook can read it.** This covers commit messages only. Stated in the script's header rather than left to be discovered.
- A `#`-prefixed line that genuinely is a comment will now trip the guard. For a guard the asymmetry decides it: a false positive is a reword, a false negative is the failure already paid for.
- The hook is bypassable with `--no-verify`, like every pre-commit hook here. Moving it to CI is not possible in the same form — `actions/checkout` defaults to depth 1, so the branch's commit messages are not reliably available.

## Verification

- `uvx --with pytest --with pyyaml pytest scripts/ -q` → **99 passed, 4 skipped**
- Pre-commit ran in full on every commit here, the new hook included, and passed
- Pre-PR third-party review: `codex review`, which produced finding 3 above


frankbria:
## Post-PR third-party review

**Reviewer:** `codex review` (opencode attempted first, 0 bytes — sixth occurrence this session).

It returned **one finding, and it was right**:

> **[P2] Detect negations split by line wrapping** — when a commit message wraps between the negation and the closing keyword, e.g. `This does not\nclose #123`, this grep runs line-by-line and will not match even though GitHub still sees `close #123` on the second line and closes the issue.

Reproduced against the then-current guard before fixing — `exit=0` on exactly that input — then fixed by collapsing newlines to spaces before matching, with two wrapped fixtures added and a mutation that reverts the flattening (2 tests fail).

Worth noting what the finding says about the guard's history: this was the **third** defect in it, and all three were in the *filtering of the input* — skipping `#` lines, and letting line boundaries break a match — never in the pattern, which was correct from the first line and never changed.

No other findings.

Issue #682, opened 2026-09-13
[P3.36] The destructive button's white label is 3.76:1 — below AA in every state, delete buttons included


frankbria:
Found while implementing #642 and deliberately left out of it: #642 is about the *busy* state, this is the variant's ordinary, enabled state.

## The measurement

`tailwind.config.js` pins the brand red and its label:

```js
destructive: {
  DEFAULT: "rgb(239, 68, 68)",
  foreground: "white",
},
```

White on `rgb(239, 68, 68)` measures **3.76:1**. WCAG 2.1 SC 1.4.3 requires **4.5:1** for normal-sized text. There is no inactive-component exemption here — this is an enabled, clickable button.

| state | ratio | note |
|---|---|---|
| enabled | **3.76** | below AA |
| busy (after #642) | **3.76** | same fill, full opacity |
| disabled-for-unavailable | 1.97 | exempt under 1.4.3 |

Theme-fixed, so it is the same figure in light and dark. Measured with `src/__tests__/theme/helpers/contrast.ts` against this repo's `tailwind.config.js`, the same helper used by #629, #632, #634 and #642.

## Where it shows

Every `<Button variant="destructive">`. The two that matter most are the ones a user reads under pressure:

- `src/components/books/DeleteBookModal.tsx` — "Delete Book" / "Deleting..."
- `src/app/profile/page.tsx` — "Delete account permanently" / "Deleting…"

Both are irreversible actions where the label is the last thing read before committing.

## Why this is not fixed in #642

#642 took these buttons from 1.97:1 to 3.76:1 by removing the opacity flattening, which is a real improvement and the whole of what that issue was about. The remaining shortfall is the red itself, and closing it means one of:

1. **Darken the brand red** — e.g. `rgb(220, 38, 38)` (`red-600`) is 4.83:1 with white, `rgb(185, 28, 28)` (`red-700`) is 6.94:1. Changes the brand everywhere the destructive variant appears.
2. **Keep the red, change the label colour** — a very dark red or near-black on the existing fill. Unusual for a destructive button and probably worse.
3. **Accept it and record why** — 3.76:1 is close, and large text (18.66px bold or 24px) only needs 3:1, so a size change would also clear it.

1 is the straightforward answer and it is a brand decision, which is why this is a ticket rather than a commit.

## Guarded already

`src/__tests__/theme/busy-button-contrast.test.ts` (from #642) pins the current figure:

```ts
[...]

PR #683, opened 2026-09-13
fix(a11y): busy buttons keep their status text legible (#642)

frankbria:
Implements **option 1** from #642, which asked for a decision on a three-way fork.

## The decision, and why it isn't really a fork

`button.tsx` puts `disabled:opacity-50` in the shared base string. CSS `opacity` groups the element, so the fill and the label are **each** blended 50% with the page behind the button rather than composited against each other. The default variant's label drops from **6.29:1 to 2.29:1** in light.

For a button that is merely *unavailable* that is correct — WCAG 1.4.3 exempts text in an inactive component, and the greyed look is the affordance. The problem is that this codebase also uses `disabled` for **busy**, on buttons that render their own status text while working. That text is the only signal the action is running, and it becomes the hardest text on the screen to read at exactly the moment it matters.

The three options were: (1) a distinct busy treatment, (2) raise the disabled opacity everywhere, (3) leave it. Option 2 was already eliminated on the issue — `opacity-70` reaches 3.4, still short of 4.5. That leaves a real choice between 1 and 3, and **option 1 is the only one that reaches AA while changing nothing about the unavailable state**: `busy` is opt-in, so no existing call site moves unless it says so, and a genuinely disabled button looks exactly as it did.

```tsx
<Button disabled={!isConfirmed || isDeleting} busy={isDeleting}>
```

Note the compound condition: `busy` takes the **busy half only**, so a button greyed for being unconfirmed still reads as unavailable.

## The call-site list was wrong three times, and the third correction is the interesting one

| pass | found |
|---|---|
| #642's own list | 5 |
| my enumerated-verb sweep | 11 |
| after the reviewer's finding, broadened | **23** |

The reviewer pointed out that my pattern missed `Verifying…`, `Changing…` and `Please wait…`. The reason it did is worth stating plainly: **I enumerated ten verbs taken from what the sweep had already found**, which is circular. A gerund plus an ellipsis is the actual shape. Broadening to `/(\b\w+ing|Please wait)(\.\.\.|…)/` found twelve more sites than my list, more than doubling the fix.

**And narrowing the pattern back could not fail the guard.** A sweep of the form "everything matching X must also have Y" passes trivially when X matches nothing — shrinking X makes it *greener*. So the one mutation that should have caught this was the one that survived. A floor assertion (`>= 23` busy-labelled Buttons found) now makes it fail. The existing anti-vacuity fixtures did not cover this: they pin the pattern against a literal, not against the tree.

## The reviewer's second finding was a hole in the feature itself

Every class in `BUSY_CLASSES` is `disabled:`-prefixed. `asChild` renders someone else's element, which may be an anchor with no `disabled` attribute — so **none of them applied**, and a "busy" link stayed fully clickable. The prop's own contract said otherwise.

Reproduced before fixing:

```
PROBE aria-disabled: null
PROBE has unconditional pointer-events-none: false
```

Fixed with an unprefixed class set plus `aria-disabled`, since an anchor cannot carry `disabled`. My own test for that path had asserted the link does *not* get `disabled` — the half I was thinking about — and never asserted what it gets instead.

## Guarded in two halves

**Measured, not asserted.** `busy-button-contrast.test.ts` reads `BUSY_CLASSES`, the base `disabled:opacity-N`, and the brand colours out of `button.tsx` and `tailwind.config.js`, and the surfaces out of `globals.css`. One of its tests **reproduces the defect** at the base opacity, so the fix is measured against a failure that still exists rather than against a number in a comment.

**Swept, with a parser that holds.** `disabled={count > 0}` mis-slices under a naive `indexOf('>')`; the element scan balances JSX braces, and that case is a fixture.

## Mutations

Eleven, each reverted with the revert asserted on a committed tree:

| Mutation | Result |
|---|---|
| drop `busy` from one call site | sweep fails |
| `BUSY_CLASSES` back to `opacity-50` | vacuity check fails |
| base opacity to 100 | defect no longer reproducible — 2 fail |
| low-contrast brand fill | AA check fails |
| remove `aria-busy` | 3 behaviour tests fail |
| remove the implied `disabled` | 2 fail |
| pass `disabled` through `asChild` | 1 fails |
| apply busy classes unconditionally | 1 fails |
| revert the `asChild` class set | 1 fails |
| drop `aria-disabled` | 1 fails |
| narrow the label pattern | floor fails |

## What this does not fix

**Filed as #682**: the `destructive` variant's white label is **3.76:1** on the brand red in *every* state, enabled included. `busy` takes the two delete buttons from 1.97 to 3.76 and no further — the rest is the red itself, and darkening it is a brand decision. The guard pins the current figure, so whoever changes the red sees this test fail and finds the issue.

**Raw `<button>` elements.** A further set of busy labels sit on hand-styled `<button>`s (`TocReview`, `ClarifyingQuestions`, `edit-toc`, `summary`) that `buttonVariants` never touched, so `disabled:opacity-50` was never their problem and `busy` does not reach them. Named here rather than left as a silent gap; each carries its own colours and would need its own measurement.

## Verification

- **149 suites, 2675 passed**, 4 skipped
- `npx eslint .` → **0 errors**; `npm run typecheck` clean
- `gate:react-hooks` (17 ledgered), `typecheck:tests` (176 ledgered), `check:specs` — all pass
- Pre-commit ran in full on every commit and passed
- Pre-PR third-party review: `codex review`, which produced both findings above


frankbria:
## Post-PR third-party review

**Reviewer:** `codex review` (opencode attempted first, 0 bytes — seventh occurrence this session).

It returned **two findings, both real, both fixed before this PR opened**:

- **[P2] The busy-label sweep missed existing status labels** — `Verifying…`, `Changing…`, `Please wait…`. Correct, and the cause is worth naming: I enumerated ten verbs taken from what the sweep had already found, which is circular. Broadened to a gerund plus an ellipsis, it found **twelve** more sites than my list — 23 in all, more than double.

  Following that up found something the reviewer did not say and that matters more: **narrowing the pattern could not fail the guard**, because a sweep only fails on what it matches. That mutation was the one that survived. A floor assertion (`>= 23` busy-labelled Buttons) now makes it fail.

- **[P2] `busy` did not make an `asChild` button inert** — every class in `BUSY_CLASSES` is `disabled:`-prefixed, and `asChild` omits `disabled`, so none applied and the link stayed clickable. Reproduced (`pointer-events-none: false`) before fixing, then fixed with an unprefixed class set plus `aria-disabled`. My own test for that path asserted what the link does *not* get and never what it gets instead.

Eleven mutations pass after both fixes.

Issue #684, opened 2026-09-13
[P3.38] Four raw <button> busy labels are 2.29-4.09:1 — the six #683 could not reach, measured


frankbria:
Named as out of scope in #683 (which implemented #642's option 1) rather than left as a silent gap, and measured here so it is a ticket with numbers rather than a suspicion.

## Why #683 does not reach these

#642 was about `<Button>`, whose shared base string carries `disabled:opacity-50`. The fix is a `busy` prop on that component. **Six busy-labelled buttons in the tree are raw `<button>` elements** with hand-written classes — `buttonVariants` never styled them, so they neither inherited the defect nor inherit the fix. Each carries its own colours and needed its own measurement.

## Measured

Same helper as #629/#632/#634/#642 (`src/__tests__/theme/helpers/contrast.ts`), against this repo's `globals.css` and Tailwind palette. WCAG 2.1 SC 1.4.3 floor for normal text is **4.5:1**, and the inactive-component exemption does **not** apply: these buttons are disabled *because they are working*, and the label is the only signal of that.

| site | busy label | disabled treatment | light | dark |
|---|---|---|---|---|
| `edit-toc/page.tsx:581` | `Saving...` | `bg-indigo-600` + `disabled:opacity-50` | **2.29** | 3.37 |
| `summary/page.tsx:327` | `Saving...` | `bg-indigo-600` + `disabled:opacity-50` | **2.29** | 3.37 |
| `toc/NotReadyMessage.tsx:107` | `Analyzing...` | `disabled:bg-green-800` + `disabled:opacity-50` | **2.35** | 3.54 |
| `toc/TocReview.tsx:134` | `Saving...` | `disabled:bg-green-800 disabled:text-green-400` | **4.09** | 4.09 |
| `toc/TocReview.tsx:114` | `Regenerating...` | `disabled:bg-muted disabled:text-muted-foreground` | 4.61 | 5.86 ✅ |
| `summary/page.tsx:229` | `Listening...` | `bg-red-600 text-white`, no opacity | 4.83 | 4.83 ✅ |

**Four of six fail**; two already pass and need nothing.

The first three are the same defect #642 described — CSS `opacity` groups the element, so the fill and the label are each blended 50% with the page behind rather than composited against each other. `TocReview.tsx:134` is different and does not involve opacity at all: `green-400` on `green-800` is simply 4.09:1.

## An existing ledger row points here

[...]

PR #686, opened 2026-09-13
fix(a11y): the four raw <button> busy labels the Button fix could not reach (#684)

frankbria:
Finishes what #683 started. #642's fix landed on the shared `<Button>`; these six busy labels sit on hand-styled raw `<button>` elements that `buttonVariants` never touched, so they neither had the defect nor got the fix.

## Measured first

| site | busy label | disabled treatment | light | dark | |
|---|---|---|---|---|---|
| `edit-toc/page.tsx` | `Saving...` | `bg-indigo-600` + `disabled:opacity-50` | **2.29** | 3.37 | fixed |
| `summary/page.tsx` | `Saving...` | `bg-indigo-600` + `disabled:opacity-50` | **2.29** | 3.37 | fixed |
| `NotReadyMessage.tsx` | `Analyzing...` | `disabled:bg-green-800` + `disabled:opacity-50` | **2.35** | 3.54 | fixed |
| `TocReview.tsx` (accept) | `Saving...` | `disabled:bg-green-800 disabled:text-green-400` | **4.09** | 4.09 | fixed |
| `TocReview.tsx` (regenerate) | `Regenerating...` | `disabled:bg-muted disabled:text-muted-foreground` | 4.61 | 5.86 | already passed |
| `summary/page.tsx` (mic) | `Listening...` | `bg-red-600 text-white`, no opacity | 4.83 | 4.83 | already passed |

## The fixes are smaller than #683's, because the shapes differ

**Three are disabled *only* while busy**, so their "disabled" styling **is** their busy styling and the treatment simply comes off:

- `TocReview` accept: drop `disabled:text-green-400`, keeping `text-white` → **4.09 → 7.13**
- `NotReadyMessage`: drop `disabled:opacity-50` → **2.35 → 7.13**

**Two are disabled for two different reasons** — saving, and an empty TOC or an invalid summary — so the opacity becomes conditional: full while saving (**6.29**), greyed while genuinely unavailable, where WCAG 1.4.3 exempts it and the greyed look is the affordance.

```tsx
className={`... ${isLoading ? 'disabled:opacity-100 cursor-progress' : 'disabled:opacity-50'}`}
```

**`aria-busy` everywhere it belongs.** The microphone toggle is exempt with the reason written into the guard: `aria-pressed` already says what it is, and `aria-busy` — "this element is being updated" — would be the wrong word for a toggle.

## The #637 guard caught the consequence before the suite did

Removing `text-green-400` made `stranded-accent-baseline.json`'s `TocReview` row stale, and the build failed until it was pruned:

```
● has no ledgered utility the file no longer contains (#624)
+   "src/components/toc/TocReview.tsx :: text-green-400",
```

That is the **third** stale-row rule to fire on something real today rather than on a planted mutation — after #676's ledger twice and #685's two forward pointers. The rule earns its keep.

## The sweep now covers raw `<button>` too

Same floor as the `<Button>` half (`>= 6` raw busy buttons found), so narrowing the label pattern fails rather than quietly checking less — the hole the reviewer's finding on #683 exposed.

## A mutation that did not apply looked exactly like one that survived

Four mutations; three failed as intended and the fourth passed. I was one keystroke from writing it up as "the guard does not cover `aria-busy` on `NotReadyMessage`". It does — my replacement string carried fourteen spaces of indentation where the file has ten, so it matched nothing, the file was unchanged, and the suite passed **because there was no mutation to catch**.

A surviving mutation and a no-op mutation produce identical output. Redone with `assert n == 1` on the substitution, it fails correctly:

| Mutation | Result |
|---|---|
| restore `disabled:text-green-400` on TocReview's accept | 1 fails |
| drop `aria-busy` from `NotReadyMessage` (substitution asserted) | 1 fails |
| drop `aria-busy` from `edit-toc` | 1 fails |
| narrow the label pattern | 3 fail (both floors + a fixture) |

`tasks/lessons.md` records it: a mutation check whose mutation is unverified is a test of nothing, and it fails safe in the direction that makes a weak guard look thorough.

## Verification

- **149 suites, 2678 passed**, 4 skipped
- `npx eslint .` → **0 errors**; `npm run typecheck` clean
- Theme guards: 328 passed, including the pruned ledger
- Pre-commit ran in full on every commit and passed
- Pre-PR third-party review: `codex review` — *"I did not find any introduced correctness, accessibility, or test-guard issues that warrant an inline finding."*

## Known limitations

These stay raw `<button>` elements. Converting them to `<Button busy>` would be tidier and would put them under the component's contract, but it changes padding and height metrics I cannot see the result of from here — that wants a human eye, and the contrast defect does not need to wait for it.


frankbria:
## Post-PR third-party review

**Reviewer:** `codex review` (opencode attempted first, 0 bytes — eighth occurrence this session; the documented hang, now without exception).

> The changes correctly add busy-state accessibility markers and preserve contrast for the affected raw buttons. I did not find any introduced correctness, accessibility, or test-guard issues that warrant an inline finding.

No findings to triage. Worth noting that the two findings from #683's review are what shaped this PR: the floor on the sweep, and the habit of asking what a guard does when its pattern matches nothing.

PR #689, opened 2026-09-13
fix(a11y): split the destructive role in two, because one colour cannot do both (#682)

### LINKED ISSUE #682: [P3.36] The destructive button's white label is 3.76:1 — below AA in every state, delete buttons included

frankbria:
Closes #682, and fixes a second failure the issue did not know about.

## The issue's defect, and the one found while measuring it

#682 is about the **button**: white on `rgb(239, 68, 68)` is **3.76:1**, below AA in *every* state including enabled, on two irreversible actions — "Delete Book" and "Delete account permanently".

The same value drives `text-destructive`, and **that role was failing too**, across **44 call sites** nobody had ticketed:

| | light | dark |
|---|---|---|
| `text-destructive` on `--background` | **3.76** | 5.26 |
| `text-destructive` on `--card` | **3.76** | 4.76 |
| `text-destructive` on `--muted` | **3.45** | 4.02 |

## One colour cannot serve both uses

On a dark background they pull opposite ways — white needs a dark red to sit on, red text needs a light one:

| candidate | white on it | on dark `--background` |
|---|---|---|
| red-700 | **6.47** | 3.06 ✗ |
| red-400 | 2.24 ✗ | **7.16** |

So the role becomes two:

- **`destructive.DEFAULT`** → `rgb(var(--destructive-rgb) / <alpha-value>)`, red-700 in light and red-400 in dark. Carries every text shape automatically.
- **`destructive.surface`** → red-700, theme-fixed. Five opaque fills (`button.tsx`, `badge.tsx`, and three `hover:bg-destructive` sites).

Measured on every surface either lands on: light **6.47 / 6.47 / 5.93** on background / card / muted and **5.45** on the `bg-destructive/10` error card; dark **7.16 / 6.48 / 5.47** and **4.76** on `bg-destructive/20`.

`dark:bg-destructive/60` comes off the Button and Badge variants — against a dark-red base it composites to a near-black control that passes contrast at 11.37 while no longer reading as red.

## Two rounds of review, each finding something no guard could

**Round 1 — the override could not reach five of the six class shapes.** The first cut used a `.dark .text-destructive` stylesheet rule. The reviewer pointed out `text-destructive/90` is a *different class*, and so are `data-[variant=destructive]:text-destructive`, `data-[variant=destructive]:focus:text-destructive`, `data-[error=true]:text-destructive` and `*:data-[slot=alert-description]:text-destructive/90`. That is the #632 trap five times over, and each new shape would silently miss. Moving to a token removed the surface entirely.

**Round 2 — the one worth reading.** A theme-aware `var(--destructive)` measures perfectly, and makes Tailwind v3 **stop emitting every opacity variant**. Built the CSS rather than take it on trust:

```
bg-destructive\/10          0
border-destructive\/20      0
text-destructive\/90        0
ring-destructive\/20        0
```

Zero rules each. Error-card tints, invalid-state rings and destructive menu backgrounds would have shipped **colourless**, with the entire theme suite green — because every guard in that directory reads the config and does colour maths, and **none of them can see whether a class produces a rule**.

Fixed with the channel form. `destructive-utilities-emitted.test.ts` now **builds the stylesheet** and asserts each of the 15 destructive classes in the tree appears in it; reverting to the plain `var()` fails 9 tests.

## Two existing ledgers caught the knock-ons before the suite did

- **#635's spinner ledger** named `destructive` as `DeleteBookModal`'s button surface — now `destructive-surface` — and its colour resolver read oklch only, so the new rgb-channel token priced as `null`. Both fixed, with the row's reason rewritten to say why the fill is the theme-fixed shade.
- **#642's recorded shortfall** flips from "there is a gap, and it is this big" to "there is no gap", which is exactly what recording it was for.

## Mutations

Five, each with its substitution count asserted:

| Mutation | Result |
|---|---|
| `DEFAULT` back to a theme-fixed literal | 2 fail |
| `surface` back to red-500 | 2 fail |
| dark token → the light red | 4 fail |
| an opaque `bg-destructive` back in a component | 1 fails |
| light token → the dark red | 4 fail |

plus the one that motivated the emission guard: `DEFAULT` back to a plain `var()` → **9 fail**.

## Verification

- **151 suites, 2702 passed**, 4 skipped
- `npx eslint .` → **0 errors**; `typecheck`, `typecheck:tests`, `gate:react-hooks` all pass
- Pre-commit ran in full on every commit
- Pre-PR review: `codex review`, twice — the second pass: *"consistently split destructive text and surface colors, update the affected call sites, and add targeted guards for contrast and Tailwind utility emission. I did not identify any introduced runtime or test-breaking issues."*

## Known limitations

**This is a visible brand change** — a deeper red in light, a lighter one in dark. Reverting is `tailwind.config.js` plus two lines in `globals.css`; the guards state the ratios they expect, so a revert fails loudly rather than silently.

The emission guard covers the destructive family only. Widening it to every token means a Tailwind build per family and is a bigger job than this bug warrants — but the trap it catches is not specific to `destructive`, and any future token moved to a `var()` has it.


frankbria:
## Post-PR third-party review

**Reviewer:** `codex review`, run twice — the architecture changed between them because the first pass found the change was wrong.

**Round 1 [P2]** — the `.dark .text-destructive` override missed `text-destructive/90`, `data-[variant=destructive]:text-destructive`, `data-[variant=destructive]:focus:text-destructive`, `data-[error=true]:text-destructive` and `*:data-[slot=alert-description]:text-destructive/90`. Each is a different class name. Replaced the override with a theme-aware token, which has no such surface.

**Round 2 [P2]** — and this is the one worth reading: a plain `var(--destructive)` measures perfectly and stops Tailwind v3 emitting **every** opacity variant. Verified by building the CSS: `bg-destructive/10`, `border-destructive/20`, `text-destructive/90` and `ring-destructive/20` each emitted **zero rules**. Fixed with the channel form, and a new guard builds the stylesheet and asserts every destructive class appears in it.

Both findings were beyond what any existing guard could see, because they all read the config and do colour maths. That gap is now recorded in `tasks/lessons.md`: *a guard that checks an input to a generator should ask what it takes for the generator to produce nothing from a valid-looking input, and check the output instead.*

**Final pass:** *"consistently split destructive text and surface colors, update the affected call sites, and add targeted guards for contrast and Tailwind utility emission. I did not identify any introduced runtime or test-breaking issues."*

PR #691, opened 2026-09-14
fix(lint): derive ClarifyingQuestions' "nothing to load" instead of storing it (#584)

frankbria:
A per-site pass over all 17 remaining `set-state-in-effect` sites against the current tree, rather than trusting the classification I wrote earlier. Two looked mechanically fixable. I probed each with `npx eslint` on the single file instead of reasoning about it — **one was, and one was a trap.**

## Fixed: `ClarifyingQuestions.tsx:54`

The effect ended:

```ts
if (bookId && questions.length > 0) {
  loadExistingResponses();
} else {
  setIsLoadingResponses(false);
}
```

That `else` announces *"there is nothing to load"* — a fact already knowable from `bookId` and `questions` during render — by writing it into the **fetch's** state. The state now means one thing, *the fetch has not finished*, and the skeleton reads both conditions where it is rendered:

```ts
const showLoadingResponses = isLoadingResponses && Boolean(bookId) && questions.length > 0;
```

**17 → 16**, and the ledger row for the file is pruned.

## Reverted: `summary/page.tsx:45`, and this is the useful half

Initialising `isLoading` to `true` and dropping the synchronous `setIsLoading(true)` **does** clear the warning — verified by probe, suite green.

Then I read the other `setIsLoading` call sites:

```
29:  const [isLoading, setIsLoading] = useState(true);
56:      .finally(() => setIsLoading(false));   ← the fetch
194:    setIsLoading(true);                      ← the save
203:      setIsLoading(false);
```

```tsx
{isLoading ? 'Saving...' : 'Continue to TOC Generation'}
```

**One flag doing two jobs, and the second is the submit button's label.** The page would have opened with its button reading **"Saving..."** before anything was saved. No test covers it, so nothing would have failed — it was caught by reading, before writing the test.

The real fix is to split it into `isLoadingSummary` and `isSaving`. That is a behaviour-affecting refactor of a page with **no test file**: worth doing, and not as a quiet rider on a lint fix. Recorded on #584 so the next pass starts from the bounded change rather than rediscovering the trap, and in `tasks/lessons.md` as the general form — when a change hinges on a variable's *initial* value, the question is how many distinct things that variable means, not whether the value is right at mount.

## Mutations, and why one of them passing is the point

| Mutation | Result |
|---|---|
| hard-wire `showLoadingResponses` to `false` | 1 fails |
| drop the `questions.length > 0` condition | 1 fails |
| **restore the deleted `else` branch** | **3 pass** |

The third is not a gap. The `else` was **redundant, not load-bearing** — the derivation makes it a no-op — so the old code passing the new tests is exactly how that is known. It is the same property used in #677: a test that only fails on the new implementation describes the refactor, not the behaviour.

The three tests pin what the `else` was providing (no skeleton with no questions, none with no `bookId`) plus the counterweight (the skeleton *does* show while a real fetch is in flight and clears when it resolves), so the derivation cannot swallow the state it derives from.

## The other 15

Re-verified against the current tree, with the specific blocker for each:

| # | shape | blocker |
|---|---|---|
| 6 | fetch kick-off (`fetchBooks()`, `loadPreferences()`, `void loadSessions()`, …) | the rule flags any `setState` reachable inside an `async` callee; the only route that satisfies it is reshaping `async`/`await` into `.then()` chains, which suits the linter and not the reader |
| 4 | init from `localStorage` / `sessionStorage` / URL | a `useState` initializer reads browser-only state, so server and first client render differ — a hydration mismatch traded for a render pass |
| 2 | prop mirrored into user-editable state | the React-docs fix adjusts state during render, which trips `set-state-in-render`, at `error` here |
| 1 | latch (`setIsAuthReady(true)`) | set once, never reset; the derived form re-shows a full-page loader over rendered content on a session refetch |
| 1 | `setError` from a URL param | 14 call sites; splitting into `urlError` + `formError` is a UX decision inside a password-reset flow |
| 1 | `setFormat` from async preferences | initialisation from data that arrives late, behind a `touched` ref |

## Verification

- `npm run gate:react-hooks` → PASS, **16 known**, all ledgered
- **151 suites, 2705 passed**, 4 skipped; `npx eslint .` → 0 errors; typecheck clean
- Pre-commit ran in full and passed
- Pre-PR review: `codex review` — *"cleanly derives the loading UI from existing state and inputs, updates the baseline accordingly, and adds targeted tests for the affected behavior"*


frankbria:
CI caught a type error my local verification could not: the new test file's mocked promise used `as ReturnType<typeof bookClient.getQuestionResponses>` on a `Promise<unknown>`, which widened the resolver's parameter and lost the assignment. Fixed by typing the promise properly, with the fixture now matching the real response shape.

Worth naming why it got through: I ran `npx jest`, `npx eslint` and `npm run typecheck` and reported green. **Neither of the two commands that could have caught it was among them** — `tsconfig.json` excludes every test-shaped path, which is the entire reason `typecheck:tests` exists (#625), and jest runs through SWC which strips types without checking them. Adding a test file makes `typecheck:tests` load-bearing, and that is the case where it is easiest to skip because the change "is only a test". Recorded in `tasks/lessons.md`.

`npm run typecheck:tests` → PASS, 176 known and all ledgered; the 3 tests pass.

PR #694, opened 2026-09-14
fix(export): a Settings change to the default format was ignored after one touch (#693)

### LINKED ISSUE #693: [P3.39] A Settings change to the default export format is ignored once the format has been touched
Found while examining `ExportOptionsModal`'s `set-state-in-effect` site for #584. It is not a lint problem — fixing the lint would not fix this, and fixing this would not fix the lint — so it is filed separately rather than folded in.

## Two facts, both readable in the source

**1. `optionsTouchedRef` is never reset.**

```tsx
const optionsTouchedRef = useRef(false);
useEffect(() => {
  if (!userPreferences || optionsTouchedRef.current) return;
  if (userPreferences.default_export_format) setFormat(userPreferences.default_export_format);
  …
}, [userPreferences]);
```

The only writes are `optionsTouchedRef.current = true` at three call sites (`ExportOptionsModal.tsx:219`, `:315`, `:354`) — the format radio group, and two other controls. Nothing ever sets it back to `false`.

**2. The component never unmounts.** `src/app/dashboard/books/[bookId]/page.tsx:606` renders it unconditionally:

```tsx
{/* Export Modals */}
<ExportOptionsModal
```

It is not behind `{showExportModal && …}`; visibility is controlled by the `isOpen` prop through `<Dialog open={isOpen}>`. So the ref survives every open/close cycle for the life of the page.

## The consequence

The comment says *"Pre-select the user's stored export defaults (#64) **until they interact**"*. "They interact" is meant per-open. As written it is **per-mount**.

So: open the export modal, change the format once, close it. Then go to Settings, change your default export format, come back, and open the modal again. The effect returns early on `optionsTouchedRef.current`, and **your new default is silently ignored** until a full page reload.

That path is live rather than theoretical since #674: `useUserPreferences` became a real external store, so `invalidateUserPreferencesCache(updated)` — which `settings/page.tsx:137` and `profile/page.tsx:157` both call on save — now notifies mounted consumers. Before #674 the consumer would not have seen the change at all, which is presumably why this was never noticed.

**Stated honestly:** facts 1 and 2 are verified by reading. The user-visible consequence follows from them and from #674's notification behaviour; I have not reproduced it against a running app.

## The decision this needs

Resetting the ref on close is one line, but it changes what reopening the modal shows, and that is a product call rather than an obvious bug fix:

1. **Reset on close.** Each open pre-selects the saved defaults; a per-session ad-hoc choice does not persist. Matches what the comment says, and makes a Settings change take effect immediately.
2. **Keep the stickiness, drop the preference check.** Reopening shows the last choice — which is what happens today, since `format` state also persists — but the effect stops pretending it will re-apply preferences.
3. **Reset only when preferences actually change.** Track the preference object identity; a Settings change re-applies, an ordinary reopen keeps the last choice. More code, and arguably the least surprising behaviour.

I would take **1**: it is the smallest change, it matches the documented intent, and "I changed the setting and nothing happened" is a worse failure than "my ad-hoc choice was not remembered".

## Acceptance criteria

- [ ] A decision recorded on the three options
- [ ] Changing the default export format in Settings takes effect in an already-open page, demonstrated by a test
- [ ] Whatever reopening does is pinned by a test, so the behaviour is deliberate rather than incidental
- [ ] Mutation-checked

## Related

- #584 — the `set-state-in-effect` site this was found beside; **unaffected**, and the reason it stays ledgered (initialisation from data that arrives late) is unchanged.
- #674 — made `useUserPreferences` notify mounted consumers, which is what turns this from inert into live.
- #64 — the feature whose intent the comment states.

frankbria:
Closes #693.

## The defect

`ExportOptionsModal` pre-selects the user's stored export defaults "until they interact":

```tsx
const optionsTouchedRef = useRef(false);
useEffect(() => {
  if (!userPreferences || optionsTouchedRef.current) return;
  // …apply default_export_format, default_page_size, include_empty_chapters
}, [userPreferences]);
```

Two facts make that guard permanent:

1. **The ref is never reset.** Its only writes are `= true`, at three control handlers.
2. **The component never unmounts.** `books/[bookId]/page.tsx:606` renders it unconditionally — not behind `{showExportModal && …}` — with visibility carried by `<Dialog open={isOpen}>`.

So one touch of the format control disabled the stored defaults for the life of the page. "Until they interact" meant per-**mount**, not per-open. Change your default export format in Settings, come back, open the modal: **your new default is silently ignored until a reload.**

**#674 is what made it reachable.** Before it, `useUserPreferences` mirrored a module cache into state, so a Settings save was only ever seen by the *next* component to mount. #674 made it a real external store whose saves notify mounted consumers — which is when this guard started swallowing them. The bug was inert until the fix that exposed it.

It was found beside `ExportOptionsModal`'s `set-state-in-effect` site for #584, but **it is not a lint problem**: fixing either would not fix the other, which is why it was filed on its own.

## The fix — option 3 from the issue, not the option 1 I recommended when filing

The ref now guards the **first** preference arrival only:

```tsx
const hasAppliedPreferencesRef = useRef(false);
useEffect(() => {
  if (!userPreferences) return;
  const isFirstArrival = !hasAppliedPreferencesRef.current;
  hasAppliedPreferencesRef.current = true;
  if (isFirstArrival && optionsTouchedRef.current) return;
  // …apply
}, [userPreferences]);
```

- **First arrival:** preferences load asynchronously, so a format picked in the moment before they land must not be overwritten. That is what `optionsTouchedRef` is genuinely for, and why it cannot simply be deleted.
- **Any later arrival:** can only come from the user deliberately saving new preferences, and that wins over a stale ad-hoc choice.
- **Reopening the modal** keeps the last choice, because nothing resets it.

I recommended option 1 (reset on close) when filing. On implementing it, that would re-pick the saved default on *every* open, which is worse for anyone exporting the same book repeatedly in a non-default format. Option 3 fixes the Settings case without that cost.

## Tests — the page had none for this

Four, in `__tests__/ExportOptionsModal.preferences.test.tsx`:

| test | pins |
|---|---|
| applies the stored default when preferences arrive | the ordinary case |
| does not overwrite a choice made before preferences land | the first-arrival guard, and why the ref still exists |
| **applies a later change even after the format has been touched** | **the defect** — `main`'s component fails this |
| leaves the choice alone when the preferences object is unchanged | a re-render for any other reason is not a preference change |

## Mutations

| Mutation | Result |
|---|---|
| `main`'s pre-fix component | **fails** the later-change test |
| drop the first-arrival guard | 1 fails |
| restore the old guard-every-arrival behaviour | 1 fails |
| never apply preferences at all | 2 fail |

**One mutation found an unreachable guard in my own first cut.** It also compared the preferences object's identity (`appliedRef.current === userPreferences`), and dropping that comparison passed all four tests — because the effect's `[userPreferences]` dependency already guarantees the effect only runs on an identity change. Removed rather than justified, which is #672's rule applied to my own code.

## Review

Pre-PR `codex review` returned one finding, marked P1: the hoisted `jest.mock` factory closed over `let preferences`, which it said "can fail to transform before any tests run".

**Checked rather than taken on trust — it does not reproduce here.** This repo transforms with SWC (`next/dist/build/swc/jest-transformer.js`), which does not enforce the rule; the file transforms and passes, alone and in the full suite. `babel-plugin-jest-hoist` *does* enforce it, so the fix is still worth having for portability, and it matches the `mockPush` / `mockParams` naming in #692's test. Renamed to `mockPreferences` in `43bb9df`, with a comment saying exactly that.

## A note on the first commit

`7630383` was committed with `--no-verify` because the session was interrupted for a reboot between the last gate run and the commit. After restarting I ran the skipped hooks explicitly over the range (`pre-commit run --from-ref origin/main --to-ref HEAD`) — all passed — and ran the commit-msg guard against its message. The second commit went through hooks normally.

## Verification

- `npm run lint` → **0 errors**
- `npm run typecheck` → clean
- `npm run typecheck:tests` → PASS, 176 known, all ledgered
- `npm run gate:react-hooks` → PASS, 15 known — this does **not** change the #584 count, as expected
- `npm run check:specs` → PASS
- `npx jest` → **2713 passed**, 4 skipped

## Known limitations

Not reproduced against a running app with a real Settings save: the behaviour is pinned through `useUserPreferences`'s interface, which #674 already tests for the notify-on-invalidate half. `ExportOptionsModal`'s own `set-state-in-effect` site stays ledgered, for the reason recorded on #584.


frankbria:
## Post-PR third-party review

Recorded from the pre-PR pass; the only change since is the fix for its finding.

**Reviewer:** `codex review`. One finding, **[P1]** — the hoisted `jest.mock` factory closed over `let preferences`, which "can fail to transform before any tests run".

**Verified before acting, and the failure does not reproduce here.** jest runs through SWC (`next/dist/build/swc/jest-transformer.js`), which does not enforce the `mock`-prefix rule; the file transformed and passed on its own and in the full 2713-test suite. `babel-plugin-jest-hoist` *does* enforce it, so the rename is a real portability fix rather than a no-op — applied in `43bb9df`, with the comment stating why, and matching the `mockPush`/`mockParams` naming in #692.

Worth recording as a case where the finding's *consequence* was wrong for this repo while its *remedy* was still right.

PR #696, opened 2026-09-14
fix(theme): unclassed borders drew in currentColor, not --border (#695)

### LINKED ISSUE #695: [P2.27] Unclassed borders render in currentColor: the base layer reads v4-only --color-* variables that v3 never defines
## What is wrong

`globals.css`'s base layer reads three variables that don't exist at runtime:

```css
@layer base {
  * { border-color: var(--color-border); }
  body { background-color: var(--color-background); color: var(--color-foreground); }
}
```

The `--color-*` names are defined only inside the `@theme inline { ... }` block, and only Tailwind v4 understands that block. The app runs Tailwind **v3.4.19**, which passes it through as an unknown at-rule. So all three declarations are invalid at computed-value time. Because they come *after* preflight's `border-color`, they still win the cascade and fall back to the property's initial value.

For `border-color` the initial value is `currentColor`. **Any element with a `border` and no border-colour class draws its border in the text colour**: near-black in light mode, near-white in dark. `--border`, the subtle colour this rule was written to apply, never shows up. `<body>` is unaffected only because `layout.tsx` also sets `bg-background text-foreground` on it.

## Measured

Computed style in Chromium, on a stylesheet built from `main` with the installed Tailwind v3, for an unclassed `border` element:

| | light | dark |
|---|---|---|
| `main` (renders) | `oklch(0.145 0 0)` = currentColor | `oklch(0.985 0 0)` = currentColor |
| intended (`--border`) | `oklch(0.922 0 0)` | `oklch(1 0 0 / 0.1)` |

Against the light background that is **19.80:1** rendered, vs the **1.26:1** hairline the token specifies.

## Reach

A sweep of shipped `.tsx`/`.ts` class strings found **38** with a border width and no border colour. They include shadcn's `Card`, `Dialog`, `AlertDialog`, `Alert`, `DropdownMenu` and `Select`, plus the chapter AI panels, the export options and the 2FA recovery codes. So this is most bordered surfaces in the app, not a few outliers.

## Fix

Point the base layer at the raw tokens: `var(--border)`, `var(--background)`, `var(--foreground)`. That's also the correct shape for #513. Under v4 a theme alias is computed on `:root`, so `var(--color-border)` would inherit the light value into `.dark`.

Add a guard that fails when any `var(--x)` outside `@theme` has no declaration outside `@theme`. That covers the whole class of defect, not just these three lines.

## Known consequence

Borders become the intended hairline. One hand-rolled `<input>` in `DraftGenerator.tsx` uses bare `border`, so its boundary drops to 1.26:1. That matches the shared `Input` component's `border-input` today, so it's a design-system-wide WCAG 1.4.11 question, not something this fix introduces into one place.

frankbria:
Closes #695

## Summary

`globals.css`'s base layer read `--color-border`, `--color-background` and `--color-foreground`. Those names exist only inside the `@theme inline` block, and Tailwind v3 ignores that block. So each declaration was invalid at computed-value time, still beat preflight in the cascade, and `border-color` fell back to **`currentColor`**. Every `Card`, `Dialog`, `Alert`, menu and panel with a bare `border` drew a hard text-coloured edge instead of the hairline token.

- The base layer now reads `var(--border)`, `var(--background)`, `var(--foreground)`. That's also the right shape for #513, because under v4 an alias computed on `:root` would carry the light value into `.dark`.
- New guard `theme-vars-defined.test.ts`: any `var(--x)` outside `@theme` with no declaration outside `@theme` fails the build. It covers the defect class, not just these three lines.

## Evidence

Computed `border-top-color` in Chromium for the real `ui/card.tsx` class string, on stylesheets built by the installed Tailwind 3.4.19 from `main` and from this branch:

| | light | dark |
|---|---|---|
| `main` | `oklch(0.145 0 0)` (currentColor) | `oklch(0.985 0 0)` (currentColor) |
| this PR | `oklch(0.922 0 0)` (`--border`) | `oklch(1 0 0 / 0.1)` (`--border`) |

Screenshots of the same probe: `main` shows a black edge in light and a white edge in dark; this branch shows the hairline in both.

Reach: a sweep of shipped class strings found **38** with a border width and no border colour.

## Guard, mutation-checked

| Mutation | Result |
|---|---|
| restore `var(--color-border)` | 1 failed |
| restore `var(--color-foreground)` | 1 failed |
| disable the `@theme` strip (+ restore the alias) | 3 failed (vacuity assertions) |
| clean | 6 passed |

## Gates

- jest: 2713 passed, 4 skipped (full suite, on `main`'s dependency set)
- lint 0 errors · typecheck · typecheck:tests · react-hooks gate (15, unchanged) · changelog heading guard
- Pre-PR third-party review: `codex review --base main` found no issues (opencode hung this session, so codex was used as the fallback)

## Known limitations

- The hand-rolled `<input>` in `DraftGenerator.tsx` uses bare `border`, so its boundary drops to the same hairline the shared `Input` component already uses. Whether input boundaries meet WCAG 1.4.11 is a design-system-wide question; this PR doesn't introduce it in a new place.
- The page looks different: borders are lighter everywhere. That is the design the tokens specify, and the design the stylesheet was always trying to apply.


frankbria:

Reviewed at `afc8cec`, the code change. The only later commit, `1f94c36`, adds a changelog entry.

> The changes correctly replace Tailwind v4-only theme aliases with raw CSS tokens in base styles and add a targeted regression test. I did not identify any introduced correctness, reliability, security, or maintainability issues that warrant blocking the patch.

No findings.

Issue #697, opened 2026-09-14
[P2.28] Opacity variants of var() theme colours emit no CSS: secondary hover, muted tints, input fill and focus-ring colour are dead


frankbria:
## What is wrong

`tailwind.config.js` configures ten theme colours as a bare `var()`:
`border`, `input`, `ring`, `background`, `foreground`, `secondary`, `muted`, `accent`, `popover`, `card` (plus their `-foreground` entries).

Tailwind v3 can't apply an opacity modifier to a colour it can't parse, so **`<token>/<n>` classes for these tokens emit no CSS rule at all**. This is the same mechanism #682 found for `destructive` and fixed there with the channel form. That fix couldn't be reused here, because these tokens are oklch.

## Measured

Every `<prefix>-<token>/<n>` class in shipped source, built with the installed Tailwind 3.4.19, gave **14 distinct classes and 14 with zero rules**. The same absence shows in the full app stylesheet. Controls in the same build emit: `bg-destructive/10` (channel form) and `bg-primary/90` (rgb).

| Class | Files | Where it bites |
|---|---|---|
| `hover:bg-secondary/80` | 8 | `Button variant="secondary"` and `Badge` have **no hover state** |
| `ring-ring/50` | 6 | shared `Input`/`Textarea`/`Select` focus ring falls back to Tailwind's default blue instead of the token |
| `bg-muted/50`, `/30`, `/20`, `/40`, `/80` | 8, 6, 2, 1, 1 | row hovers and AI-panel tints render transparent |
| `bg-input/30`, `/50` | 4, 2 | dark-mode input fill missing |
| `bg-accent/50`, `bg-card/50`, `bg-secondary/50`, `/90`, `border-foreground/50` | 1–2 each | |

## Fix

Define these colours as `color-mix(in oklch, var(--x) calc(<alpha-value> * 100%), transparent)`. A probe build in v3 shows every alpha variant emitting (`bg-muted/50` becomes `calc(0.5 * 100%)`), while opaque classes keep the `--tw-*-opacity` form.

Widen `destructive-utilities-emitted.test.ts` from the destructive family to **every** opacity-modified colour class in shipped source, so the next `var()` token can't regress silently. That guard's comment said widening was "a bigger job than the bug warrants"; this issue is the evidence that it does.

Tailwind v4 (#513) handles these natively, so the config change is interim. The guard isn't.

Issue #634, opened 2026-09-10
[P2.33] ChapterTab's errored tab paints a light-only bg-red-50: its label is 1.05:1 in dark mode


frankbria:
Found while demoing #629 (PR #633) — the dark-theme screenshot of `ChapterTab`
showed the errored row's label as a blank strip.

## The measurement

`ChapterTab.tsx:94` applies the error state as a light-only surface, with no
`dark:` counterpart:

```tsx
chapter.error && "border-red-200 bg-red-50",
```

The tab's own text colour keeps coming from the active/inactive branch above it
(`text-foreground` when active, `text-muted-foreground` when not), which flips
with the theme while `bg-red-50` does not. Measured from the `globals.css`
tokens with `src/__tests__/theme/helpers/contrast.ts`:

| foreground on `bg-red-50` | light | dark |
|---|---|---|
| `--foreground` (active tab) | 18.10 ✅ | **1.05** ❌ |
| `--muted-foreground` (inactive tab) | 4.59 ✅ | **2.36** ❌ |

1.05:1 is not "low contrast", it is invisible: near-white text on a near-white
card. WCAG 2.1 **1.4.3** wants 4.5:1 for this text. An errored chapter is
precisely the tab a user needs to be able to read.

The companion `border-red-200` is 1.45:1 on light `--background` — it also does
no work as the error affordance in the theme where it *is* applied, though the
red `AlertCircleIcon` on the same row carries that signal.

This is the most severe of the open P2.3x contrast cluster (#629 was 2.33:1 on a
dot; #632 is 1.76:1 on a badge) — those are hard to see, this one cannot be read
at all.

## Not covered by the existing guards

- `axe-core` cannot see it: jsdom carries no stylesheet, and the repo's axe runs
  do not render this state.
- The #623/#629 guard (`status-dot-contrast.test.ts`) sweeps **non-text**
  indicator colours against `--background`/`--muted`. This is body text against a
  palette-literal card surface — a different pairing the sweep does not model.
- #632 is about dark **translucent** overlays (`bg-<hue>-900/NN`). This is the
  mirror failure: an opaque *light* surface that was never given a dark variant.

## Fix sketch

Give the error surface a dark counterpart the way #623 did for the dots — e.g.
`border-red-200 bg-red-50 dark:border-red-900 dark:bg-red-950` — and verify both
foregrounds clear 4.5:1 on it in both themes.

## Acceptance criteria

[...]

PR #699, opened 2026-09-14
feat(frontend): Tailwind CSS v4 and tailwind-merge 3, painting what v3 painted (#513)

### LINKED ISSUE #513: [P3.24] Migrate to Tailwind CSS v4 together with tailwind-merge v3

frankbria:
Closes #513

## Summary

Tailwind CSS 3.4.19 → 4.3.3 and tailwind-merge 2 → 3, **with the app painting what it painted on v3**. `@tailwindcss/upgrade` did the mechanical renames. It also made about a dozen silent mistakes, each found by measuring rendered output rather than reading the diff, and each fixed by hand here.

**The gate was the rendered result, not the CSS text.** A Chromium computed-style diff of all **975** class strings in shipped source: v3 built from `main`, v4 built from this branch, both themes, 64 properties.

| Group | Count | Why it's not a change |
|---|---|---|
| `rounded-full` | 400 | v4 writes `calc(infinity * 1px)`, v3 writes 9999px; both draw a pill |
| `outline-hidden` | 56 | `outline-style: none` vs a transparent 2px outline; neither shows. Forced-colours still gets v3's outline |
| transforms | 30 | v4 uses the separate `translate`/`rotate`/`scale` properties; geometry checked, identical |
| prose colours | 8 | typography plugin's grays come from v4's oklch palette, about one channel unit off v3's hex |
| `shadow-2xs` (inputs, outline buttons), `rounded-xs` (dialog close) | 8 + 16 | shadcn classes that didn't exist in v3 now render as designed, the same call as #697 |
| button width | 4 | probe artifact: the page didn't define next/font's `--font-nunito-sans`; with it defined, geometry is identical |
| **total** | **522** | **0 unexplained** |

## What the codemod got wrong, and the fix

| Defect | Effect if shipped | Fix |
|---|---|---|
| a later `@theme inline` redefined `--color-primary` | brand indigo repainted near-black | single `@theme`, brand pinned `rgb(79 70 229)` |
| destructive aliased through `:root` | stuck at the light red in dark | `@theme inline` resolves at the element |
| v4's palette is oklch, a different set of colours | ~100 utilities shift; every published ratio invalid | v3's 242 palette colours pinned as hex |
| `:root`/`.dark` moved into `@layer utilities`, #610's override left unlayered | an unlayered rule beats every utility in v4 | `main`'s layout restored |
| `variant="outline"` → `"outline-solid"` (2 props) | wrong button variant (typecheck caught it) | reverted; word-level sweep of every rename found no other |
| `bg-opacity-50`, `ring-opacity-50` left as-is | mobile nav overlay **solid black** | slash form, plus a tripwire guard |
| `editor.css` never opened | **production build failed** on `@apply prose` | `@reference` the theme; `outline-none` → `outline-hidden` |
| rewrote guard-test fixtures and the English word "ring" | guards silently lost coverage | test edits reverted, reviewed by hand |
| typography plugin and `tailwindcss-animate` dropped | 0 `.prose` rules | `@plugin`, `tw-animate-css` |
| reversed 9 stacked child-targeting variants (`*:data-[slot=select-value]:flex` → `data-[slot=select-value]:*:flex`) to keep v3's meaning | kept 9 shadcn rules dead: grey destructive menu icon, unclamped select value | shadcn's original order; v4 applies them as designed (measured on real markup) |

The mechanical `/50` conversion of `ring-opacity-50` would itself have halved a dark ring v3 painted opaque: in v3, `dark:ring-blue-700` reset the opacity. The computed diff caught it.

## tailwind-merge 3

Locked at **3.6.0**, because 3.7.0 was published the day before and has had no exposure. Diffing v2 against v3 over 3945 static class strings in shipped source turned up only the two opacity sites above, plus v3 correctly parsing `h-(--x)`, which v2 couldn't.

## Guards

The theme suite read `tailwind.config.js` (deleted) and `tailwindcss/colors` (oklch under v4, a palette the app doesn't paint).

- New `helpers/palette.ts` reads the pinned palette and `@theme` colours from `globals.css`, the file that actually decides them.
- `opacity-utilities-emitted.test.ts` builds the real `globals.css` with `@tailwindcss/cli`, and rejects v3 `-opacity-N` utilities.
- Mutation-checked, all caught:

| Mutation | Result |
|---|---|
| drop `--color-muted` from `@theme` | 5 failed |
| brand primary → indigo-400 | 18 failed |
| destructive fill back to red-500 | 2 failed |
| pinned green-700 lightened | 6 failed |
| palette pin reformatted (vacuity) | 8 suites fail to load |
| `bg-black/50` back to `bg-opacity-50` | 1 failed |

## Also

- `postcss.config.mjs` → `@tailwindcss/postcss`; `autoprefixer` removed (v4 prefixes itself).
- Dependabot: the build group names `@tailwindcss/*`, and the `tailwindcss`-major ignore is removed (its own rule: remove when #513 closes). The config test now matches Dependabot's glob patterns with `fnmatch`. The old `==` check **rejected the correct config**, `@tailwindcss/*`, outright.
- Comments and the theme-guard live rule that said "defined in `tailwind.config.js`" now point at `@theme`, so the next reader doesn't price a declaration that no longer exists (#634).

## Gates

- jest **2782 passed**, 4 skipped · lint 0 errors · typecheck · typecheck:tests · react-hooks gate (15) · `scripts/` pytest 104 passed · npm audit 0 advisories · **`next build` passes**, and the shipped CSS was checked: `.text-red-600{color:#dc2626}`, `.bg-primary{background-color:#4f46e5}`, `--ring` at #697's values
- Pre-PR third-party review (`codex review --base main`): **2 findings, both real, both fixed** (select value, destructive menu icon). One detail was off: the review says they "previously applied", but on v3 they were already dead. A sweep of every reordered variant list then found 7 more of the same kind. Details in the review comment.

## Known limitations

- **Prose grays** differ from v3 by about one channel unit (e.g. gray-900 `rgb(17,24,39)` → `oklch(0.21 0.034 264.7)`). Pinning them means overriding about 36 `--tw-prose-*` variables for a difference nobody can see; not done.
- The computed diff covers static states on the element carrying the class. Hover and focus were checked by hand on the classes #697 changed, and child-targeting variants by comparing emitted selectors, not exhaustively.
- Several `wip(513)` commits were made `--no-verify` while the branch was red by design. From the guard port onward every commit ran the hooks, and squash-merge collapses the history.


frankbria:
## Third-party review (codex, `codex review --base main`), pre-PR at `259f7b5`

> The Tailwind v4 migration includes incorrectly reordered arbitrary variants that no longer target the intended child elements. This causes visible regressions in select value layout and destructive dropdown icon styling.
> 
> Full review comments:
> 
> - [P2] Restore child-slot variants for select values — frontend/src/components/ui/select.tsx:41-41
>   When a select value contains longer text or inline content, these new `data-[slot=select-value]:*:` variants compile as if the trigger itself had `data-slot="select-value"`, so they never match the actual child `<SelectValue data-slot="select-value">`. This drops the line-clamp/flex/gap styling that previously applied to the selected value and can let select labels overflow or misalign next to the chevron.
> 
> - [P3] Restore destructive menu icon targeting — frontend/src/components/ui/dropdown-menu.tsx:78-78
>   For destructive dropdown items with an icon, `[svg]:*:data-[variant=destructive]:text-destructive!` compiles to a selector that expects the class to be on an SVG, not on the menu item. Because the generic SVG rule later paints icons `text-muted-foreground`, destructive item icons no longer get the red destructive color.
> 
> 
> 

### Triage

Both findings are **real**, and both are fixed in `8009148`. One part of the premise was checked and is wrong: the review says the classes "previously applied". The v3 build of `main` emits `.*\:data-\[slot\=select-value\]\:flex[data-slot="select-value"] > *`, which requires the *trigger itself* to carry the value's slot. It never does, so the rule was dead on `main`. v3 emitted no rule at all for the `[svg]` form. The codemod reversed the variant order to preserve v3's meaning, and v3's meaning was a no-op. Restoring shadcn's original order makes v4 apply them as designed.

Measured in Chromium on real component markup:

| | `main` (v3) | this PR (v4) |
|---|---|---|
| selected value | `display: block`, no clamp | `display: flex`, `line-clamp: 1` |
| destructive menu icon | `oklch(0.54 0 0)` (muted grey) | `rgb(185, 28, 28)` (destructive) |
| alert description | `rgb(185, 28, 28)` | destructive at 90% |
| select item's last span | `display: block` | `display: flex` |

A sweep of **every** class whose variant list the codemod reordered found nine: the two reported, plus `alert.tsx`'s description colour, the other three `select-value` classes, and three `*:[span]:last:` classes. All nine were dead on v3, and all are restored. The computed-style diff couldn't see these because it measures the element carrying the class, not its children; that blind spot is recorded in `tasks/lessons.md`.

frankbria:
## Third-party review (codex, `codex review --base main`), post-PR at `8009148`

Re-run because the code changed after the pre-PR pass (the nine restored variants, `8009148`).

> I did not identify any discrete, actionable regressions introduced by the Tailwind v4 migration or related guard/config updates.

No findings.

PR #703, opened 2026-09-14
fix(lint): read ChapterEditor's saved tab instead of copying it in an effect (#584)

frankbria:
Part of #584 (15 → 14 `set-state-in-effect` sites). It doesn't close it: the close condition is the ledger emptying.

## Summary

`ChapterEditor` restored the per-chapter tab choice with an effect that called `setView` from `sessionStorage`. #584 filed this shape ("init from storage") as blocked: a `useState` initializer that reads storage renders differently on the client and mismatches the server HTML.

`useSyncExternalStore` removes that blocker. Its server snapshot is the writing view, so React hydrates against the server HTML, then re-renders with the client's saved view. #674 already used this pattern for `useUserPreferences`.

A choice made in the editor is stored tagged with its chapter key. That keeps both behaviours the effect provided:
- reusing the editor for another chapter (TabContent renders it without a key) falls back to **that** chapter's saved view;
- a choice sticks even when `sessionStorage.setItem` throws (privacy modes).

## Evidence

**Hydration, tested rather than assumed.** `ChapterEditor.hydration.test.tsx` renders to a server string with no saved view, hydrates on a client that has `questions` saved, and asserts no recoverable errors, no mismatch console errors, and the questions view shown afterwards. The same probe run against the old effect version gives the same clean result, so behaviour is unchanged.

| Mutation | Result |
|---|---|
| naive `useState` initializer reading storage | hydration test **fails** (hydration error) |
| ignore the in-editor choice | 1 failed (storage-refuses test) |
| drop the chapter tag | 1 failed (reuse-after-click test) |
| never read storage | 2 failed |

The existing toggle tests seed storage and never click, so they couldn't reach the chapter tag or the write-failure path. Two new tests cover those.

## Gates

- jest full suite green · lint 0 errors · typecheck · typecheck:tests · react-hooks gate **14** (ledger row 3 → 2)
- Pre-PR third-party review: `codex review --base main` found no regressions

## Known limitations

- The other "init from storage/URL" sites on #584 aren't this shape. `summary/page.tsx:93` races the fetch that also writes `summary`, `ChapterEditor`'s `hasBackup` has eight setters, and `settings:86` strips the URL param it reads.


frankbria:

Reviewed at `5091d61`, the PR head.

> The change preserves the per-chapter tab behavior while avoiding hydration mismatches, and the added tests cover the relevant restore, reuse, and storage-failure scenarios. I did not identify any actionable regressions in the diff.

No findings.

PR #704, opened 2026-09-14
fix(lint): VoiceTextInput resets its mode during render, not in an effect (#584)

frankbria:
Part of #584 (14 → 13 `set-state-in-effect` sites). The close condition is the ledger emptying, so #584 stays open.

## Summary

`VoiceTextInput` copied its `mode` prop into state in an effect, so a prop change first painted the stale mode, then corrected it. It now uses React's documented pattern for state that resets when a prop changes: keep the previous prop in state and adjust during render.

```tsx
const [modeProp, setModeProp] = useState<InputMode>(mode);
if (modeProp !== mode) {
  setModeProp(mode);
  setCurrentMode(mode);
}
```

## The blocker on #584 was wrong

#584's table listed "prop mirrored into user-editable state" as blocked, because the render-time adjustment "trips `set-state-in-render`, at `error` here". Probed against the installed rules:

| Probe | Result |
|---|---|
| guarded (`if (prev !== prop) { setPrev(prop); setX(prop) }`) | **clean** |
| unguarded `setX(prop)` in render (positive control) | `react-hooks/set-state-in-render` **error** |

## Tests first

Two tests were written and passed against the **old** effect before it was replaced:
- a parent's `mode` change is followed;
- a toggle holds across an unrelated re-render, and a text → voice → text prop round trip resets it.

| Mutation | Result |
|---|---|
| compare against the shown mode instead of the previous prop | 2 failed |
| remember the prop but skip the reset | 2 failed |
| key the reset on the prop's **value** (a tagged choice) | 1 failed, the round-trip test |

## Gates

- jest full suite green · lint 0 errors · typecheck · typecheck:tests · react-hooks gate **13** (VoiceTextInput's row removed)

## Known limitations

- `ChapterEditor`'s other site filed as a "prop mirror" isn't one: it pushes `initialContent` into the TipTap editor and reads `editor.getHTML()`, which is effect work.


frankbria:

Reviewed at `a46562f`, the PR head.

> The diff replaces the mode-prop mirroring effect with a guarded render-time reset and updates the hook baseline/tests accordingly. I did not find an introduced correctness, reliability, security, or maintainability issue that warrants an inline finding.

No findings.

PR #705, opened 2026-09-14
fix(export): book stats load per open without a synchronous loading flag (#584)

frankbria:
Part of #584 (13 → 12 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

`ExportOptionsModal` loaded book stats with an effect that called `setLoadingStats(true)` synchronously and cleared it when **any** response settled. That was a lint warning, and it hid a race:

1. Open the export modal; stats start loading.
2. Close it before they arrive, then reopen; a second request starts.
3. The **first** open's response lands and is treated as the current one: its figures appear and Export enables while the current request is still running. If it lands after the second response, it overwrites the current figures.

## Fix

- Opens are counted **during render**, using the lint-clean pattern from #704.
- Each open's request is keyed `bookId:openCount`.
- `loadingStats` is **derived**: the modal is open, and no response has settled for this key.
- The effect sets state only after awaiting, and ignores a response for an open the user dismissed.

Unchanged: each reopen hides the figures and disables Export until *that* open settles, and a failure re-enables Export and toasts.

## Tests

| Test | Old effect | This PR |
|---|---|---|
| hides stats and disables Export until the request settles | pass | pass |
| re-enables Export on failure | pass | pass |
| refetches on every open and waits for that open's response | pass | pass |
| ignores a dismissed open's response that lands **first** | **fail** | pass |
| ignores a dismissed open's response that lands **last** | **fail** | pass |

Two corrections from checking the tests themselves:
- The first race test originally let the stale response land with a bare `setTimeout`, and **passed against the racy code**, because the render was never flushed. It now resolves inside `act`.
- A mutation removing the `ignore` flag **survived** that test. Keying alone already hides a stale response that arrives first; the flag matters only when it arrives last. The fifth test covers that ordering.

| Mutation | Result |
|---|---|
| apply a dismissed open's response (drop `ignore`) | 1 failed |
| key the request on the book only | 1 failed |
| never settle on failure | 1 failed |
| don't count opens | 1 failed |

The three `trackOperation` mocks returned a new function per render, unlike the real hook's `useCallback([])`, which would make any effect depending on it re-run forever. They now return one function, so the effect keeps its full dependency list with no lint suppression.

## Gates

- jest green (the 50 export tests plus `PreferenceDefaults`) · lint 0 errors · typecheck · typecheck:tests · react-hooks gate **12**


frankbria:

Reviewed at `36c4841`, the PR head.

> The changes correctly key and ignore export stats requests across modal open/close races, and the added tests cover both stale-response orderings. I did not find any introduced issues that would break existing behavior.

No findings.

PR #706, opened 2026-09-14
fix(lint): derive ActiveSessionsList's loading state from the current attempt (#584)

frankbria:
Part of #584 (12 → 11 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

`ActiveSessionsList`'s loader called `setLoadState('loading')` before every request. On mount that was a no-op, since the initial state is already `'loading'`; it only did anything on **Retry**, going error → loading. It was also the site's lint warning.

Now:
- Retry starts a new **attempt** (setState in an event handler).
- The effect runs per attempt and records how it ended: `{ attempt, state: 'loaded' | 'error' }`.
- `loadState` is **derived**: an attempt with no recorded outcome is loading.

This is the keyed shape from #705, minus its stale-response guard, deliberately. Retry only renders once the current attempt has failed, so two attempts are never in flight, and a guard would be code no test could reach.

## Tests

Written first, and passing against the old code:
- Retry shows "Loading sessions…" and hides the error while the request is in flight, then lists the sessions;
- a retry that also fails shows the error again.

| Mutation | Result |
|---|---|
| ignore which attempt an outcome belongs to | 1 failed |
| Retry doesn't start an attempt | 3 failed |
| errors never recorded | 3 failed |

## Gates

- `SecuritySettings.test.tsx` 20 passed · full suite via pre-commit · lint 0 errors · react-hooks gate **11** (row removed)


frankbria:

Reviewed at `9f0e72b`, the PR head.

> The changes correctly derive retry loading state from the active attempt and add targeted tests for retry behavior. I did not identify any introduced correctness issues in the modified code.

No findings.

PR #707, opened 2026-09-14
fix(lint): derive the settings page's load state from the current attempt (#584)

frankbria:
Part of #584 (11 → 10 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

The same shape as #706, applied to the settings page. `loadPreferences` called `setLoadState('loading')` synchronously before each request. That was a no-op on mount and mattered only on Retry. Now:
- Retry starts a new load **attempt**;
- the effect records `{ attempt, state: 'loaded' | 'error' }`;
- `loadState` is **derived**: an attempt with no outcome is loading.

Unchanged: the `active` guard, the once-per-visit theme-sync ref, and saving staying disabled until a load succeeds.

## Tests

Written first and passing against the old code: while a retry is in flight, the page shows the disabled "Loading..." button and no error alert, then enables Save once the retry succeeds.

| Mutation | Result |
|---|---|
| ignore the attempt key | 1 failed |
| Retry is a no-op | 2 failed |
| load errors never recorded | 2 failed |

## Gates

- `SettingsPageSave` + `SettingsPageBilling` 20 passed · full suite via pre-commit · typecheck · react-hooks gate **10** (row 2 → 1)

## Known limitations

- The file's other warning, the `?checkout=` / `?tab=` effect, stays ledgered: it strips the param it reads and fires a toast.
- `main` already carries an unused `eslint-disable` directive in this file (line 112); left out of this diff.


frankbria:

Reviewed at `0b8d134`, the PR head.

> The changes correctly derive the loading state from the current load attempt and preserve the existing stale-response guard behavior. I did not find any introduced correctness, security, or maintainability issues in the diff.

No findings.

PR #708, opened 2026-09-14
fix(lint): derive reset-password's URL error message instead of copying it into state (#584)

frankbria:
Part of #584 (10 → 9 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

The reset-password page copied the `?error=` code's message into the form's `error` state with an effect, after the first render. With an error in the URL the form never renders; the page shows only the "Invalid Reset Link" card. So that message is the only one the card can show, and it's now a pure function of the URL (`describeUrlError`). Side effect: no more one-render flash of the generic fallback text before the specific message.

#584's table filed this as "a UX decision" about splitting 14 `setError` call sites. It isn't one: none of those call sites can run while the URL carries an error.

## A vacuous test

`shows error message for expired token in URL` asserted `/invalid reset link/i`. The page **title** is "Invalid Reset Link" in every invalid state, so the test passed whatever the description said.

- It now asserts the expired message itself.
- New: an unrecognised code gets the generic description, and blocks the form even with a token present.
- New: with no URL error, the fallback description shows.

All pass against the old effect.

| Mutation | Result |
|---|---|
| expired code described as "invalid" | 1 failed (**the old test passed this**) |
| unknown code gets no description | 1 failed |
| card ignores the URL | 3 failed |

## Gates

- `ResetPassword.test.tsx` 17 passed · full suite via pre-commit · lint clean for the file · react-hooks gate **9** (row removed)


frankbria:

Reviewed the code change at `413529b`; the only later commit, `5cb69ed`, is the changelog entry.

> The changes remove derived state for URL reset errors and add tests covering the intended invalid-link messages. I did not identify any introduced correctness, security, or maintainability issues in the diff.

No findings.

PR #709, opened 2026-09-14
fix(lint): apply stored export defaults during render, not in an effect (#584)

frankbria:
Part of #584 (9 → 8 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

`ExportOptionsModal` applied a newly arrived preferences object in an effect, so the modal first painted the old options, then corrected them. It now compares against the previous preferences object **during render** and applies the change there, React's pattern for state that follows a prop, as in #704.

#693's rule is unchanged:
- the **first** arrival defers to a choice the user already made;
- every **later** arrival wins, since it can only come from a deliberate Settings save.

"Touched" is now state instead of a ref, because render may not read a ref (`react-hooks/refs`, at error). The three option handlers set it exactly where they set the ref.

## Evidence

No test changes: #693's four tests and `PreferenceDefaults` pass as they are, and they catch every mutation of the new logic.

| Mutation | Result |
|---|---|
| guard every arrival once touched (the pre-#693 bug) | 1 failed |
| never defer to the user's choice | 1 failed |
| format handler doesn't mark touched | 1 failed |
| first arrival never recorded | 1 failed |

## Gates

- export suites + `PreferenceDefaults` 57 passed · full suite via pre-commit · typecheck · react-hooks gate **8** (row removed)


frankbria:

Reviewed at `d333db6`, the PR head.

> The changes move preference application out of an effect without introducing a clear functional regression. I did not find any discrete, actionable bugs in the diff.

No findings.

PR #710, opened 2026-09-14
fix(chapters): read the backup banner from storage, not a hand-synced flag (#584)

frankbria:
Part of #584 (8 → 7 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

`ChapterEditor`'s "local backup available" banner was a `hasBackup` flag, set by a mount effect and kept in step with `localStorage` by six handlers. **Two cases disagreed with storage, and both were visible to users:**

1. **The banner carried across chapters.** The mount effect never reset on a chapter change, and TabContent reuses the editor without a key. Chapter A's banner showed on chapter B, where Restore found nothing.
2. **The banner offered a backup that was never written.** A failed manual save set the flag even when the backup write threw (quota, privacy mode).

The banner is now **read from storage** through `useSyncExternalStore`, for the chapter on screen:
- handlers that write or remove the backup bump a counter so the read re-runs (storage fires no event in the tab that wrote it);
- validation parses a whole chapter and this editor re-renders on every keystroke, so it reruns only when the stored string changes;
- server snapshot `false`, so hydration matches.

## Tests

| Test | Old code | This PR |
|---|---|---|
| reused editor drops the previous chapter's banner | **fail** | pass |
| failed manual save that couldn't write the backup offers no restore | **fail** | pass |
| the 12 existing backup tests | pass | pass |

| Mutation | Result |
|---|---|
| skip validation | 1 failed |
| writes never re-read | 1 failed (Dismiss: the only handler with no other state update) |
| cache reuses an entry for a changed string | 1 failed |
| key pinned to the first chapter | 1 failed |

## Gates

- full suite green (2798) · typecheck · **typecheck:tests** · react-hooks gate **7** (row 2 → 1)
- The test-typecheck gate caught two new errors in the first draft: bare `{ content }` mocks, the same shape as the file's 16 ledgered ones. The new tests use a typed helper instead.

## Known limitations

- The file's remaining ledgered site pushes `initialContent` into the TipTap editor; that's effect work, and it stays.


frankbria:

Reviewed at `e86b2ff`, the PR head.

> The changes correctly derive the backup banner from validated localStorage state and re-render after local writes. I did not find a discrete introduced bug that would break existing behavior or tests.

No findings.

PR #711, opened 2026-09-14
fix(questions): drop a previous chapter's late questions response (#584)

frankbria:
Part of #584 (7 → 6 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

`QuestionContainer` (the interview-questions panel) is reused across chapters, and its initial load had **no stale-response guard**:

1. Open chapter 1; its questions request is slow.
2. Switch to chapter 2; its request is fast and its questions render.
3. Chapter 1's response lands last and **replaces chapter 2's questions on screen**.

Answers are saved against the question ids shown, so this could attach answers to the wrong chapter's questions. The progress fetch that follows had the same race. Both reproduce on the old component.

## Fix

- The chapter's initial load runs **inside its effect**, ignores a response (and its progress fetch) once the chapter has changed, and records which chapter has settled.
- `loading` is **derived**: the retry path's flag, or no settled load for the current chapter.
- Shared `applyQuestions` / `reportQuestionsError` helpers serve the effect and the handler-driven reload (retry, refresh after generating or saving), which keeps its own flag and its `await`ing callers.
- A chapter change clears the previous error **during render**, as each load used to when it began.

## Tests

| Test | Old component | This PR |
|---|---|---|
| a previous chapter's questions landing last don't replace the current ones | **fail** | pass |
| a previous chapter's progress landing after the switch isn't shown | **fail** | pass |
| a chapter change clears the previous chapter's error banner | pass (behaviour kept) | pass |
| the 24 existing container tests | pass | pass |

| Mutation | Result |
|---|---|
| apply a stale response | 1 failed |
| loading not derived from the chapter load | 2 failed |
| keep the previous chapter's error | 1 failed |
| apply stale progress | 1 failed |

Two tests needed a second pass. The progress and error mutations **survived** the first cut, so the progress test was added. The first error test was vacuous: it checked the generator's error while chapter 2 loaded, where the skeleton hides it whatever the state. The stale error is only visible while chapter 1's questions are still displayed, so the test now raises the banner through a failed refresh, then switches.

## Gates

- container tests 27 passed · related chapter/question suites 471 passed · typecheck:tests · react-hooks gate **6** (row removed)
- The file's `exhaustive-deps` warning count is unchanged: `main` already omitted the loader from the deps, and listing the unmemoised helpers would re-run the effect on every render


frankbria:

Reviewed at `3b98d55`, the PR head.

> The changes add a stale-response guard for the initial chapter question/progress load and tests cover the targeted race conditions. I did not identify any new discrete bugs introduced by the diff.

No findings.

PR #712, opened 2026-09-14
fix(books): a book the user left could render, and save, under the next one (#584)

frankbria:
Part of #584 (6 → 5 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

The app router keeps the book page mounted across a `[bookId]` change, and its loader had **no stale-response guard**:

1. Open book A; its request is slow.
2. Navigate to book B; its request is fast and B renders.
3. A's response lands last, and **A's details render under B's URL**. The metadata form on that page would then save A's values to book B.

## Fix

- The load runs **inside its effect** and drops a response once the book, attempt or session has moved on.
- **Try Again** starts a new attempt instead of calling the loader directly.
- `isLoading` is **derived**: loading until a load has settled for this book, attempt and session. A session change still reloads with the skeleton, as before.

## Tests (new file; the page had none)

| Test | Old page | This PR |
|---|---|---|
| the previous book doesn't render when its response lands last | **fail** (shows "Book A" under book B) | pass |
| Try Again shows the skeleton while it runs, then the book | pass | pass |

| Mutation | Result |
|---|---|
| apply a left book's response | 1 failed |
| loading ignores which load settled | 1 failed |
| Try Again is a no-op | 1 failed |

**Harness trap, documented in the test file:** `@/lib/auth-client` is mocked globally in `jest.setup.ts`, and a file-level `jest.mock` of it has no effect. The global `useSession` also returns a **new session object, with a fresh `expiresAt`, on every call**. This page lists `session` in its loader's deps, so under the default mock it refetched **73 times in 300 ms** and every test timed out. The tests configure the global mock with one stable object, which is what better-auth's own `data` is.

## Gates

- book page suites 41 passed · full suite via pre-commit · typecheck:tests · react-hooks gate **5** (row removed) · the page's stale "unnecessary dependency: `session`" warning is gone too, since the effect now uses it


frankbria:

Reviewed at `d3e6759`, the PR head.

> The changes add a stale-response guard for book loading and retry behavior without introducing any clear regressions in the inspected diff. I found no discrete, actionable issues that would warrant blocking the patch.

No findings.

Issue #713, opened 2026-09-14
[P3.45] jest.mock factories for @/lib/auth-client never take effect: the jest.setup.ts block is dead, and per-test auth mocks may be hollow


frankbria:
## What is wrong

`jest.mock('@/lib/auth-client', factory)` does not take effect in the frontend Jest setup. That covers both the 114-line global factory in `src/jest.setup.ts` and a test file's own.

Components receive the **real** `src/lib/auth-client.ts`, which builds its client from `src/__mocks__/better-auth-react.ts` through `moduleNameMapper`.

## Evidence (2026-09-14, found while writing the book page tests for #584)

- A test file with its own `jest.mock('@/lib/auth-client', () => ({ useSession: () => ({ data: { user: { id: 'u1' } } }) }))` that imports `useSession` gets `test-user-id`, the `better-auth/react` mock's data. The module's keys are the real module's exports in SWC's alphabetical order.
- Editing the `jest.setup.ts` factory's `useSession` had **no effect** on the book page; editing `src/__mocks__/better-auth-react.ts` did.
- Removing the whole `jest.setup.ts` block leaves the full suite at an **identical** 2803 passed / 4 skipped.
- `jest.setup.ts` does not import anything that loads `auth-client` before its `jest.mock`, so "the real module is cached first" is not the explanation. The mechanism is unknown.

## Why it matters

Tests that mock `authClient` methods through such a factory, for example `SettingsPageSave.test.tsx`, `SecuritySettings.test.tsx` and `ResetPassword.test.tsx`, may be exercising the `better-auth/react` mock's methods rather than their own. They pass today, but an assertion on a factory-defined `jest.fn` could be vacuous or pass for the wrong reason. The same could apply to other `@/…` modules if the cause is general.

## To do

1. Find the mechanism: SWC `jest.mock` hoisting with next/jest, the module-registry order for `setupFilesAfterEnv`, or something about this module specifically.
2. Sweep test files that pass a factory to `jest.mock('@/lib/auth-client', …)` and check, per file, whether its assertions touch the factory's functions or the `better-auth/react` mock's.
3. Delete the dead `jest.setup.ts` block, or make it effective, once the cause is known.

PR #714, opened 2026-09-14
test: useSession mock returns one stable session, as better-auth does

frankbria:
## Summary

The `better-auth/react` test mock (`src/__mocks__/better-auth-react.ts`) built a **new session object, with a fresh `expiresAt`, on every `useSession()` call**. better-auth keeps `data` referentially stable between renders. Components list `session` in effect and callback deps, so under the mock the book page refetched in a loop (73 times in 300 ms). #712 worked around it per test; this fixes the mock.

- `useSession` now returns one result object.
- The book page tests drop their workaround and become the tripwire.

## This is the mock components actually get

`@/lib/auth-client` builds its client from this module through `moduleNameMapper`. **`jest.mock('@/lib/auth-client', factory)` doesn't take effect in this setup**, whether it's the 114-line global one in `jest.setup.ts` or a test file's own:
- editing the `jest.setup.ts` factory's `useSession` changed nothing (tried first);
- removing that block entirely leaves the suite at an identical 2803 passed;
- a test file's own factory still sees this mock's data.

The block is **left in place**: the mechanism isn't understood, and tests that believe they mock `authClient` methods through such a factory may not be. Investigation filed as #713.

## Evidence

| Check | Result |
|---|---|
| full suite | 2803 passed, 4 skipped (unchanged) |
| book page tests without the workaround | 2 passed |
| mutation: mock returns a fresh object per call | book page tests **2 failed** (loop, timeout) |


frankbria:

Reviewed the change at `01f0d0b`; the only later commit is the changelog entry.

> The change makes the better-auth React mock return a referentially stable session object and updates the affected tests to rely on that default. I did not find any discrete regression introduced by this diff.

No findings.

PR #715, opened 2026-09-14
test: make jest.mock factories for @/lib/auth-client reach the code under test (#713)

### LINKED ISSUE #713: [P3.45] jest.mock factories for @/lib/auth-client never take effect: the jest.setup.ts block is dead, and per-test auth mocks may be hollow

frankbria:
Closes #713

## Summary

**Every `jest.mock('@/lib/auth-client', factory)` silently missed the code under test.** That included the 114-line global factory in `jest.setup.ts` and **14 test files'** own factories.

### Mechanism

A root manual mock, `frontend/__mocks__/@/lib/auth-client.ts`, was left over from the Clerk → better-auth migration (#39).

- SWC compiles `import … from '@/lib/auth-client'` to `require('../lib/auth-client')`, following tsconfig paths.
- `jest.mock('@/lib/auth-client', …)` keeps the **alias**.
- Jest folded the manual mock into only the **aliased** specifier's module id.

So factories registered under one id, while components loaded the real `auth-client.ts`, built on the `better-auth/react` mock, under the other.

### How it was found

| Experiment | Result |
|---|---|
| factory + `require('@/lib/auth-client')` inside the test | factory applies |
| factory + top-level `import` of the same module (`.ts` and `.tsx`) | **real module** |
| the same experiment with `@/lib/api/bookClient` | factory applies |
| SWC output of the test file | `import` rewritten to `require("../lib/auth-client")` |
| every `__mocks__` directory in the repo | `frontend/__mocks__/@/lib/auth-client.ts` |

## Change

- **Delete the root manual mock.** The 14 per-file factories now reach their components.
- **Delete the `jest.setup.ts` factory.** It was always dead, and once live it broke 8 tests that had only ever run against the real module (it lacks `twoFactor.generateBackupCodes`, among others).
- **Tripwire:** `authClientMockReachesImports.test.ts` mocks the module and imports it through the same SWC rewrite.

| Check | Result |
|---|---|
| full suite, before and after | 2803 passed, 4 skipped (unchanged) |
| only the root mock deleted (setup factory now live) | 8 failed across ForgotPassword, ResetPassword, SecuritySettings |
| mutation: restore the root mock | tripwire **fails** |

## Known limitations

- #714's changelog entry calls the mechanism unknown and says the block was "left in place"; this PR's entry supersedes both.


frankbria:

Reviewed at `b522255`, the PR head.

> The changes are limited to test mocking cleanup and a regression test. I did not identify any discrete introduced bug that would break existing behavior or tests.

No findings.

PR #716, opened 2026-09-14
fix(lint): the dashboard loads books from one keyed effect (#584)

frankbria:
Part of #584 (5 → 4 `set-state-in-effect` sites). #584 stays open until the ledger empties.

## Summary

`fetchBooks` called `setIsLoading(true)` synchronously, and was called from an effect **and** four handlers: create, delete, and two retry buttons. Now:

- **One keyed effect.** The load runs inside the effect, keyed on the session, page and a **reload attempt**. The handlers bump the attempt.
- **Supersession.** The effect's cleanup drops a superseded response, replacing the request-id ref with the same guarantee: a page click, a delete refetch and a session change can still overlap, and the newest wins.
- **Derived loading.** `isLoading` is derived from which load settled. That also removes a one-frame "not loading" flicker when a failed page change rolls the page back.

Every handler call was already fire-and-forget: the delete handler's `await fetchBooks()` had nothing after it, and the loader never throws.

## Evidence

The 63 dashboard tests (pagination, deletes, skeletons, accessibility) pass unchanged, and they catch every mutation:

| Mutation | Result |
|---|---|
| never drop a superseded response | 2 failed |
| loading ignores which load settled | 3 failed |
| reload is a no-op | 6 failed |

## Review triage

The pre-PR `codex review` raised one P2 (full comment below): after a failed page change, the rollback leaves the settled key on the failed page, so the dashboard "remains loading and immediately refetches" the rolled-back page. **Measured, it isn't a regression.** Same scenario (page 1 loads, page 2 fails with 429, roll back), on both versions:

| | `getUserBooks` calls (skip) | Next disabled after settling |
|---|---|---|
| `main` | 3 (0, 24, 0) | no |
| this PR | 3 (0, 24, 0) | no |

`main` refetched the rolled-back page too: the rollback's `setPage` changed `fetchBooks`' identity, which re-ran the effect. The existing "keeps the last good page on screen" test already queues that third response. Skipping the refetch would be a behaviour change on top of `main`, not a fix to this diff, so it isn't made here.

## Gates

- dashboard suites 63 passed · full suite via pre-commit · react-hooks gate **4** (row removed)


frankbria:
## Third-party review (codex, `codex review --base main`), at `c665c65`

> The patch introduces a loading-state mismatch after failed page navigation, causing redundant refetches and disabled UI despite already having a valid rolled-back page displayed.
> 
> Review comment:
> 
> - [P2] Settle the rolled-back page key after failures — frontend/src/app/dashboard/page.tsx:162-162
>   When a page fetch fails after navigation, the catch block rolls `page` back to `loadedPageRef.current`, but this `finally` still records the failed request's original `page:reloadAttempt` as settled. Because `isLoading` is derived from the current rolled-back `loadKey`, the dashboard remains loading and immediately refetches a page whose content is already on screen; for transient 429s this adds another request and can keep the pager disabled or show a second failure unnecessarily.
> 
> 

### Triage: not a regression

Measured on both versions with the same scenario (page 1 loads, page 2 fails 429, roll back): `main` and this PR both make 3 requests (skip 0, 24, 0) and both end with Next enabled. `main` refetched the rolled-back page because the rollback changed `fetchBooks`' identity and re-ran the effect. The existing "keeps the last good page on screen" test already queues that third response. No change made; see the PR body.

Issue #718, opened 2026-09-14
[P3.46] Summary page: local draft, server copy and typing race for the same field, and nothing decides which wins


frankbria:
## What

`frontend/src/app/dashboard/books/[bookId]/summary/page.tsx` has three writers of `summary` that race, and nothing decides which one should win:

1. **Local draft restore** (an effect, on `bookId` change): `if (local && !summary) setSummary(local)`.
2. **Server fetch** (an effect, on `bookId` change): `setSummary(data.summary || '')`, whenever the response lands.
3. **The user typing or dictating.**

## Consequences on `main` today

- **Typing during the fetch is overwritten.** Anything entered before the server response lands is replaced by the server copy.
- **A local draft is discarded whenever the fetch succeeds**, even with an empty server summary (`data.summary || ''` sets `''`). The draft survives only if the fetch *fails*.
- **On a `[bookId]` change without a remount**, the previous book's `summary` is still set when the restore effect runs, so `!summary` is false and the new book's draft is never restored.

The debounced autosave also writes `summary` back to localStorage on every change, so the "draft" is usually just the last value shown, including a value the server fetch just wrote.

## Decision needed

Which copy wins when both exist, and whether typing during a load is protected. For example: the server copy wins unless the local draft is newer (this needs a timestamp on the draft), and user edits always win over a load that completes after them.

## Context

This was the last `react-hooks/set-state-in-effect` site on #584. The effect is kept as-is behind a documented `eslint-disable-next-line` so #584's rule could return to `error` without deciding this in a lint change.

Issue #221, opened 2026-07-04
[P0.2.2] Plan upgrade / Stripe checkout flow


frankbria:
Parent epic: #174. Decision record: `docs/adr/2026-07-04-beta-entitlement-model.md`. Deferred from the free-invite beta; out of scope until paid launch is scheduled.

Backend endpoint to create a Stripe Checkout session for upgrading from `free` to a paid plan, and the return/cancel handling. Success is reconciled via the #174-decomposed webhook (P0.2.1), not trusting the client redirect.

AC: authenticated user can start checkout; on completion their `plan` reflects the purchased tier (via webhook).

PR #719, opened 2026-09-14
fix(lint): react-hooks/set-state-in-effect back at error (#584)

### LINKED ISSUE #584: [P3.31] 30 react-hooks v7 violations warned rather than fixed during the ESLint 9 migration

frankbria:
Closes #584

## Summary

`react-hooks/set-state-in-effect` is back at its default **`error`**. Across #703–#712 and #716 the backlog went from 17 to 4, and four of those fixes exposed real races (a dismissed modal enabling Export, a backup banner on the wrong chapter, a previous chapter's questions, a previous book's metadata).

The last four sites are effects doing work render cannot. Each now has an `eslint-disable-next-line` with the reason on the line above:

| Site | Why an effect is correct |
|---|---|
| `app/page.tsx` auth latch | better-auth 1.7.2 sets `isPending` back to `true` on every refetch while `data` is null; a derived value flashes the full-page loader for signed-out visitors, and `isRefetching` can't tell the first load from a refetch |
| `ChapterEditor.tsx` → TipTap | pushes content into an imperative editor, and the saved marker must move with it; whether new content is needed is only knowable by reading the editor |
| `settings/page.tsx` landing tab | read once from a URL the same effect rewrites to strip `?checkout=`; deriving it would break #221's tested clean URL |
| `summary/page.tsx` draft restore | races the server fetch and typing for the same field; **which should win is a product decision, filed as #718**, not one to make in a lint change |

The settings effect's three `setActiveTab` calls become one. That's equivalent, since a checkout result already overrode `?tab=`. Its unused `exhaustive-deps` directive is removed too.

## Gate removal

- the `"warn"` override in `eslint.config.mjs` is removed;
- the #675 ledger (`react-hooks-baseline.json`), `scripts/react-hooks-gate.mjs`, the `gate:react-hooks` npm script and its CI step are deleted: their purpose was holding the backlog while the rule was at `warn`;
- `quality-standards.md` describes the new state, and the bar for a new disable: name the external system or constraint, or derive the value instead.

## Evidence

| Check | Result |
|---|---|
| `npm run lint` | 0 errors |
| **mutation: remove one disable** | lint **exit 1**, `1 error` |
| CI runs `npm run lint` | yes, `tests.yml` Frontend Tests |
| full suite | 2804 passed, 4 skipped |
| settings tests (landing tab consolidation) | 20 passed |
| typecheck · typecheck:tests · `scripts/` pytest (104) | pass |

- Pre-PR third-party review: `codex review --base main` found no issues and judged the settings consolidation equivalent


frankbria:

Reviewed at `e833bcf`, the PR head.

> The changes primarily remove the temporary react-hooks backlog gate now that the relevant rule is enforced by ESLint, and the functional settings-tab consolidation appears equivalent to the prior behavior. Lint runs successfully with no errors, and I did not identify a discrete introduced bug that should block the patch.

No findings.

PR #720, opened 2026-09-14
fix(summary): decide which copy of a book summary wins (#718)

### LINKED ISSUE #718: [P3.46] Summary page: local draft, server copy and typing race for the same field, and nothing decides which wins

frankbria:
Closes #718

## Decision

The summary page had three writers of one field (localStorage draft, server load, the user) and no rule. The rule is now:

1. **The server copy is the source of truth** and replaces the field when it loads, *except over anything the user typed while it loaded*.
2. **The local draft is the fallback** when the load fails. It is shown, not pushed back to the server, until the user edits it.
3. **A book change starts from that book's state**, never the previous book's text.

A successful load already won on the old page, so that behaviour doesn't change.

## Defects this fixes (each reproduced by a test on the old page)

| Defect on `main` | Test |
|---|---|
| typing during a slow load was overwritten by the server copy | fails on `main` |
| the local draft was **never restored**: autosave ran on mount and wrote the empty field over it first | fails on `main` |
| after a `[bookId]` change the previous book's text stayed in the field and was written into the next book's draft key | fails on `main` |
| a previous book's slow load could land over the next book | fails on `main` |

## Found by four rounds of pre-PR review

Each `codex review` round found a real P2 in my own fix. Each was reproduced by a failing test before fixing:

1. the loaded flag reset in an effect cleanup, one render late, so switching from a loaded book let autosave erase the next book's draft → keyed by book;
2. a restored fallback draft read as unsaved and was POSTed a second later, which could overwrite a newer server copy after a transient failure → the draft is the saved baseline;
3. A → B → A before B settled found A's first load still recorded → keyed by **visit**;
4. waiting for the load also held back edits made during it, so a refresh mid-load lost them → autosave skips only the untouched field.

The fifth round found no issues.

## Evidence

- 9 precedence tests; 22 summary-page tests pass
- Mutations, all caught: load overwrites typing · no draft fallback · autosave not gated · no reset on book change · superseded load applied · unkeyed loaded flag · no fallback baseline · visits never advance · edits held until load · no load gate
- The existing "keeps auto-saving work under the gate" test now lets the load settle before advancing its debounce timer. It still types before the load settles and still proves that text is saved.
- Full suite via pre-commit · lint 0 errors · removes the summary page's `set-state-in-effect` disable from #584 (3 remain, documented in `quality-standards.md`)


frankbria:
## Third-party review (codex, `codex review --base main`): five rounds

### Round 1

> The precedence logic fixes several races, but a book switch after a successful load can still erase the destination book's local draft before a failed load can use it. This breaks the documented fallback behavior for that scenario.
> 
> Review comment:
> 
> - [P2] Reset load state before autosaving a new book — frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:63-66
>   When navigating from a loaded book to another book, this render clears `summary` but leaves `summaryLoaded` true until the previous effect's passive cleanup runs. The autosave effect for the new `bookId` can therefore run once with `summary === ''` and overwrite `book-summary-${bookId}` in localStorage before the new fetch settles; if that fetch fails, an existing draft for the new book has already been erased and the fallback cannot restore it.
> 
> 

**Triage:** real; reproduced by a failing test, then fixed (see the PR body).

### Round 2

> The precedence changes mostly address the intended races, but the failed-load fallback can now be automatically persisted back to the server without user action, risking data loss in a realistic transient failure scenario.
> 
> Review comment:
> 
> - [P2] Avoid auto-saving fallback drafts before user edits — frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:108-110
>   When `getBookSummary` fails but a stale local draft exists, the catch path sets `summary` from localStorage while `lastSaved.current` remains `''`; once `setLoadedFor(bookId)` runs, this autosave effect treats that fallback as dirty and POSTs it after 1s even if the user did not edit. In a transient GET failure where the server has a newer summary but the subsequent save succeeds, this can overwrite the server source of truth with an old local draft. Consider baselining `lastSaved.current` to the fallback value or only remote-saving fallback content after a user edit.
> 
> 

**Triage:** real; reproduced by a failing test, then fixed (see the PR body).

### Round 3

> The patch fixes several precedence races, but it leaves a new stale loaded-state path that can erase a returning book's local draft during quick navigation. That is a correctness issue in the core behavior being changed.
> 
> Review comment:
> 
> - [P2] Invalidate loaded state when returning to a book — frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:37-37
>   When a user navigates from book A to book B and quickly back to book A before book B's load settles, `loadedFor` can still be book A, so `summaryLoaded` becomes true immediately for the new book A render. The autosave effect then runs with the render-cleared `summary` and can overwrite `book-summary-A` in localStorage with `''` before the fresh load fails, defeating the draft fallback this patch is trying to protect. The loaded marker needs to be invalidated per navigation/request, not just compared to the current `bookId`.
> 
> 

**Triage:** real; reproduced by a failing test, then fixed (see the PR body).

### Round 4

> The patch fixes several precedence races, but it introduces a data-loss path for edits made while the initial summary request is still pending. That regression should be addressed before considering the patch correct.
> 
> Review comment:
> 
> - [P2] Do not block local draft saves while loading — frontend/src/app/dashboard/books/[bookId]/summary/page.tsx:112-115
>   When `getBookSummary` is slow or hangs, any text the user types before the load settles never reaches `localStorage` because this effect returns on `!summaryLoaded`; if the user refreshes or navigates away during that pending load, the new draft is lost. The new gate is useful for avoiding the initial empty overwrite, but user edits made while loading still need to be persisted locally before the fetch finishes.
> 
> 

**Triage:** real; reproduced by a failing test, then fixed (see the PR body).

### Round 5, at `156b805`

> The changes implement the intended precedence rules for server summaries, user edits, and local drafts with appropriate race guards and coverage for the key navigation/load failure scenarios. I did not find a discrete introduced defect that would block correctness.

No findings.

